PR #1513, opened 2015-01-22
rosie.ws_client: handle KeyboardInterrupt

### LINKED ISSUE #1508: rosie: Gracefully handle authentication failure
`rosie` commands prompt the user for a password when it tries to connect to password-protected servers. If the user is unable to provide a password (e.g. if they don't have an account on one of the servers) the `rosie` command will suddenly exit.

We use two Rose servers - 'au', our local server; and 'u', the shared Met Office system. Not all users at our site have an account on the 'u' server. If we configure the 'u' server centrally then users without an account are not able to use Rosie at all.

The cause appears to be [RosieWSClientAuthManager._prompt()](https://github.com/metomi/rose/blob/master/lib/python/rosie/ws_client_auth.py#L183) raising a KeyboardInterrupt if a value isn't provided.

Expected behaviour is that `rosie` ignores or prints a warning for servers it cannot authenticate to, and provides search results for the remaining servers.

matthewrmshin:
Fix #1508.

### DIFF
--- a/lib/python/rosie/ws_client.py
+++ b/lib/python/rosie/ws_client.py
@@ -155,10 +155,15 @@ class RosieWSClient(object):
                     request_detail["can_retry"] = False
                     requests_kwargs = request_detail["requests_kwargs"]
                     auth_manager = request_detail["auth_manager"]
-                    requests_kwargs["auth"] = auth_manager.get_auth(
-                        is_retry=True)
-                    results[url] = pool.apply_async(
-                        requests.get, [url], requests_kwargs)
+                    try:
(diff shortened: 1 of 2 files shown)


matthewrmshin:
@benfitzpatrick please test and review.

arjclark:
@benfitzpatrick - have taken this. There was confusion about my access to the external.

arjclark:
Looks ok to me. Manually tested and received the error line as:
```
[WARN] path/to/repo/repname/hello: {}: cancelled by user
```

PR #1515, opened 2015-01-27
rose-suite-info metadata: Remove incorrect opts entry.

arjclark:
Removes the accidentally introduced "opts" entry from rose suite-info metadata.


arjclark:
@benfitzpatrick - please review.

Issue #1491, opened 2014-12-19
rose_prune - missing features?


hjoliver:
I'm just putting these up for discussion; it could potentially wait for cylc/cylc#994, cylc/cylc#1159
 
- [x] deletion of log tarballs at some offset
- [x] keeping all output under the datac directories is an admirable goal, to make housekeeping simple, but sometimes it's not convenient.  E.g. for a suite sub-system that continually reprocesses the previous n months data, this data would likely (a) need different housekeeping from the rest of the suite output; and (b) be kept in a flat directory with cycle point filenames. So, it would be good if rose_prune could also house-keep
   * non-datac directory locations
   * cycle point labelled filenames
- [x] NOT globbing.  This comment suggests it can be done already: https://github.com/metomi/rose/issues/898#issuecomment-30506941  - can it? The code seems to simply run ```ls``` and ```rm``` via the shell, on the glob pattern, but I need ```shopt -s extglob``` to get NOT globs in the shell.

hjoliver:
Can how to do NOT globbing with rose prune be documented then?

hjoliver:
#1333 is relevant to my second point (making ```rose prune``` target locations more flexible).

PR #1535, opened 2015-02-09
rose_prune: prune command uses `bash -O extglob`

### LINKED ISSUE #1420: Built-in apps should be documented in the Rose Reference Guide?
They seem to appear only under "Running Tasks" in the User Guide, at the moment.

matthewrmshin:
This ensures that `bash -O extglob` is used to run all pruning commands.

Resolve bullet point 3 of #1491. Close #1420.


matthewrmshin:

PR #1543, opened 2015-02-12
Rosie now requires Subversion 1.8

### LINKED ISSUE #1204: rosie: hangs if subversion requires more than just a password
Once `rosie` has processed the subversion request for a password, subsequent messages from subversion are not displayed and `rosie` just hangs.

For example; On a system where subversion is not able to store passwords encrypted, after prompting for a password, it then asks the user if it is ok to store the password unencrypted. 
```
-----------------------------------------------------------------------
ATTENTION!  Your password for authentication realm:

   <svn://host:3690> c9249a3e-3897-12e2-a225-7b5bed6b8ed6

can only be stored to disk unencrypted!  You are advised to configure
your system so that Subversion can store passwords encrypted, if
possible.  See the documentation for details.

You can avoid future appearances of this warning by setting the value
of the 'store-plaintext-passwords' option to either 'yes' or 'no' in
'/home/user/.subversion/servers'.
-----------------------------------------------------------------------
Store password unencrypted (yes/no)?
```
This message is not being displayed to the user and `rosie` hangs.

matthewrmshin:
Close #1204.

### DIFF
--- a/doc/rose-install.html
+++ b/doc/rose-install.html
@@ -169,9 +169,9 @@
       <dd>
         <p><dfn>used by:</dfn> Rosie, as the version control system.</p>
 
-        <p><dfn>tested with version:</dfn> 1.6 and 1.8.5.</p>
+        <p><dfn>tested with version:</dfn> 1.8.10.</p>
 
-        <p><dfn>minimum version required:</dfn> 1.6.</p>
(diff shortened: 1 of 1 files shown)


matthewrmshin:

PR #1549, opened 2015-03-05
rose config: remove --has from docs

### LINKED ISSUE #1546: rose config --has: no such option
The help implies you can use `--has` but this doesn't work (`--quiet` works fine).
```
> rose config --has rosie-id
Usage: Print the value of OPTION in SECTION.

rose config: error: no such option: --has
```

matthewrmshin:
Option is no longer supported by the logic.

Close #1546.


matthewrmshin:
@arjclark please review.

arjclark:
Looks good to me.

PR #1562, opened 2015-03-17
rosa svn-pre-commit: fix copy with bad owner

### LINKED ISSUE #1556: rosie.svn_pre_commit: handle case insensitive user ID validation
User ID in may validate under LDAP case-insensitive. The may result in bad case for `owner` in `rose-suite.info` being allowed when the suite is created. Subsequent commits are then rejected because `author` does not compare with `owner`.

Validation should be fixed to ensure that the case of the `owner` is correct.

matthewrmshin:
On suite copy, a bad user ID in `owner` or `access-list` was not being
rejected by the Rosie pre-commit hook. This change fixes the problem.

Fix #1556.

### DIFF
--- a/lib/python/rosie/svn_pre_commit.py
+++ b/lib/python/rosie/svn_pre_commit.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-#-------------------------------------------------------------------------------
+#------------------------------------------------------------------------------
 # (C) British Crown Copyright 2012-5 Met Office.
 #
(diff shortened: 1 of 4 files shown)


matthewrmshin:

matthewrmshin:
@benfitzpatrick please hold. My test for this is in another computer.

matthewrmshin:
@benfitzpatrick tests are now pushed.

PR #1565, opened 2015-03-23
rose.suite_run: use orig cylc version on restart

matthewrmshin:
`rose suite-run --restart` should now export `CYLC_VERSION` to the value used to originally start
the suite.

(N.B. The new `rose suite-restart` command should do this already.)


matthewrmshin:
@arjclark please test.

arjclark:
Looks good to me. Manually tested as working correctly now. Test battery passes in my environment.

PR #1574, opened 2015-04-07
Improve output from rose-ana and fix a bug

scwhitehouse:
A user has said it would be nice if, where numbers differ, the values were printed. This change alters the "Within" comparison so that if two values are compared, it prints the numerical values if they differ as well as a percentage difference. If two lists of values are compared only the percentage difference is printed (as before), but now together with the position in the list of that value. So you can now tell that, for example, the fourth value in the list was the first different one and it was different by 6%, say.

This also fixes a bug where a test was meant to be on wallclock time, but it was accidentally testing CPU time instead.

### DIFF
--- a/lib/python/rose/apps/comparisons/within.py
+++ b/lib/python/rose/apps/comparisons/within.py
@@ -23,6 +23,8 @@ import re
 from rose.apps.rose_ana import DataLengthError
 
 OUTPUT_STRING = "%s %s%% %s %s: File %s c.f. %s (%s values)"
+OUTPUT_STRING_WITH_POSN = "%s %s%% %s %s: File %s c.f. %s (value %s of %s)"
+OUTPUT_STRING_WITH_VALS = "%s %s%% %s %s: File %s (%s) c.f. %s (%s)"
 PASS = "<="
 FAIL = ">"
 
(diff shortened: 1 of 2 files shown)


scwhitehouse:
@matthewrmshin, would you mind reviewing this when you get chance? It's in response to Nick's request from a few weeks ago. Thanks.

lib/python/rose/apps/comparisons/within.py
```diff
@@ -75,9 +81,16 @@ def __init__(self, task, val1, val2):
           self.percentage = "XX"
 
     def __repr__(self):
-        return OUTPUT_STRING % ( self.extract, self.percentage, FAIL,
-                                 self.tolerance, self.resultfile,
-                                 self.kgo1file,  self.numvals )
+        if self.numvals == 1:
+            return OUTPUT_STRING_WITH_VALS % ( self.extract, self.percentage,
```
matthewrmshin: Extra space after open bracket breaks PEP8. Ditto for space before close bracket.

lib/python/rose/apps/comparisons/within.py
```diff
@@ -23,6 +23,8 @@
 from rose.apps.rose_ana import DataLengthError
 
 OUTPUT_STRING = "%s %s%% %s %s: File %s c.f. %s (%s values)"
+OUTPUT_STRING_WITH_POSN = "%s %s%% %s %s: File %s c.f. %s (value %s of %s)"
+OUTPUT_STRING_WITH_VALS = "%s %s%% %s %s: File %s (%s) c.f. %s (%s)"
```
matthewrmshin: Perhaps it is worth using named arguments % dict instead of positional arguments % tuple?

scwhitehouse:
That should implement named rather than positional arguments, and also improve the PEP8 compliance.

Issue #775, opened 2013-07-02
rose_ana: test battery


matthewrmshin:
Add tests for the built-in `rose_ana` application.

scwhitehouse:
### Things to test

Including ones added by #1577

 - [X] OutputGrepper - Exact comparison fail - single numeric value
 - [X] OutputGrepper - Exact comparison success - single numeric value
 - [X] OutputGrepper - Exact comparison fail - single text value
 - [X] OutputGrepper - Exact comparison success - single text value
 - [X] OutputGrepper - Exact comparison fail - list of numeric values
 - [ ] OutputGrepper - Exact comparison success - list of numeric values
 - [X] OutputGrepper - Exact comparison fail - list of text values
 - [ ] OutputGrepper - Exact comparison success - list of text values
 - [X] OutputGrepper - Within comparison fail - percentage tolerance, single numeric value
 - [X] OutputGrepper - Within comparison success - percentage tolerance, single numeric value
 - [X] OutputGrepper - Within comparison fail - absolute tolerance, single numeric value
 - [X] OutputGrepper - Within comparison success - absolute tolerance, single numeric value
 - [X] OutputGrepper - Within comparison fail - percentage tolerance, list of numeric values
 - [X] OutputGrepper - Within comparison success - percentage tolerance, list of numeric values
 - [ ] OutputGrepper - Within comparison fail - absolute tolerance, list of numeric values
 - [ ] OutputGrepper - Within comparison success - absolute tolerance, list of numeric values
 - [ ] OutputGrepper - Preset keys test - fail
 - [ ] OutputGrepper - Preset keys test - success
 - [ ] OutputGrepper - External program test - fail, single command for both files
 - [ ] OutputGrepper - External program test - success, single command for both files
 - [ ] OutputGrepper - External program test - fail, separate commands for each file
 - [ ] OutputGrepper - External program test - success, separate commands for each file

matthewrmshin:
@stevewardle Is this issue still relevant with your rewrite?

stevewardle:
It's sort of relevant, thanks - although I was planning to go through the testing as part of the rewrite anyway (especially since the naming of everything has changed!)

matthewrmshin:
#1996 has made the original list a bit redundant. Test battery for *rose_ana* should now be done business as usual.

PR #1577, opened 2015-04-08
Add basic rose-ana test battery

scwhitehouse:
This is a first attempt at dealing with #775 (long overdue, sorry). This should give some basic tests of at least some of the rose-ana functionality. I've tried to keep away from UM-specific tests, but there's some scope for improvement. I do think any testing is better than no testing, so I'll submit this now.

These tests are updated for #1574, so they won't pass on our current site-default Rose.

### DIFF
--- a/t/rose-ana/00-run-basic.t
+++ b/t/rose-ana/00-run-basic.t
@@ -1,0 +1,67 @@
+#!/bin/bash
+#-------------------------------------------------------------------------------
+# (C) British Crown Copyright 2012-5 Met Office.
+#
+# This file is part of Rose, a framework for meteorological suites.
+#
+# Rose is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


t/rose-ana/00-run-basic.t
```diff
@@ -0,0 +1,67 @@
...
+#
+# Rose is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Rose. If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test fcm_make built-in application, basic usages.
```
matthewrmshin: Need to update comment.

Issue #1162, opened 2014-03-06
Rose Bush "Display Options" indicator and reset.


hjoliver:
It would be nice to have an indicator to show that result filtering via "Display Options" is active, as it's not obvious if Display Options is collapsed, and a reset button to turn off all active filtering.

(note I have not checked if this has been done since last release, or if it's a duplicate issue).

matthewrmshin:
Relate to #1110.

arjclark:
Closed by #1579

PR #1579, opened 2015-04-08
rose bush: jobs: hide options only if default

### LINKED ISSUE #1110: rose bush: jobs: add argument to prevent collapse of display options
The `Display Options` tab is collapsed on start up, which annoys some users.

matthewrmshin:
Only hide the *Display Options* form if all options are set as defaults.

Close #1110 and #1162.


matthewrmshin:
@arjclark please review.

arjclark:
Looks good to me. Tests as working.

PR #1581, opened 2015-04-08
rosie go: rephrase "View->Search" all revisions

### LINKED ISSUE #1210: rosie go - "view all revisions" only works with Search
I'm not sure if this is a bug or not, but it seems counterintuitive - setting "view all revisions" has no immediate effect on the view, and clicking "refresh" doesn't help either. It only works in combination with Search.

matthewrmshin:
"View all revisions" is confusing as the functionality only works with
search and query. Rephrase to "Search all revisions".

Close #1210.


arjclark:
@matthewrmshin - while the change looks ok, the underlying search with all revisions is broken, by the looks of it from your earlier interface change to use 0/1 (Error dialog):

 search: {'all_revs': '', 's': 'foo'}: 500

Can you please fix this as part of your change.

matthewrmshin:
The `all_revs` logic should now be working.

arjclark:
Looks good to me. Tested manually as working in my environment. No breakages to the test battery.

PR #1586, opened 2015-04-09
Allow ignoring tasks in rose_ana

### LINKED ISSUE #1585: Allow ignoring of tests in rose-ana
It would be nice to be able to switch-off specific tests in an optional configuration (e.g. where runs have different lengths on different platforms). Currently doing [!Name of Test] in an optional configuration gives an error about NoneTypes; this should be fixed - it's probably fairly simple.

scwhitehouse:
An attempt to fix #1585 by checking that a non-None value is return from self.config.get_value().

@matthewrmshin, there might be a better way of doing this, would you mind taking a look and reviewing when you get a minute?

### DIFF
--- a/lib/python/rose/apps/rose_ana.py
+++ b/lib/python/rose/apps/rose_ana.py
@@ -281,6 +281,12 @@ class Analyse(object):
             newtask = AnalysisTask()
             newtask.name = task
             value = self.config.get_value([task, "resultfile"])
+
+            # If the task is ignored, this will be None, so continue
+            # on to the next task
+            if value is None:
+                continue
+
             if "{}" in value:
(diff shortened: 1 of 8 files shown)


matthewrmshin:
Change looks OK. Want to add a quick test?

scwhitehouse:
Good idea, test added.

matthewrmshin:
I am getting `not ok 16 - 00-run-basic-ignore-notpresent-1` for some reason.

scwhitehouse:
That seems to intermittently fail; I've deleted that test as it wasn't testing anything which hasn't already been tested. I've also added a sleep command in there to stop those annoying NFS errors when it tries to clean the suite but thinks the files are still in use.

PR #1597, opened 2015-04-21
rosie.ws_client_auth: configurable password stores

matthewrmshin:
Specific password stores can now be configured for each prefix per
site/user.


matthewrmshin:
@benfitzpatrick please review. (This is in case we get into trouble with #1592, we need simply ask the user to add a configuration.)

PR #1599, opened 2015-04-21
t/rose-ana/00: fix hang up

matthewrmshin:
For some reason, the event hooks are causing the suite to hang. Removing
them allow the test to robustly pass without hanging.

@scwhitehouse please review.


scwhitehouse:
This looks fine (I'd be interested to know what was wrong with the event hooks, though)

PR #1611, opened 2015-05-01
rose.env.env_export: report 1st time and on change

matthewrmshin:
E.g. This fixes `rose task-run` reporting `export PATH=...` twice.

@kaday please review.

### DIFF
--- a/lib/python/rose/env.py
+++ b/lib/python/rose/env.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-#-----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


kaday:
Its passes test battery and looks good.

@matthewrmshin what do I need to do to test this in my environment?

matthewrmshin:
It already has a self test, so you should need to do any more. If you are keen to check that `rose task-run` only reports `export PATH=...` once, run a single task suite on master and then with this branch. On master, you should see `export PATH=...` twice in the task's `01/job.out`. With this branch, you should see `export PATH=...` once.

kaday:
I am happy that is works in my environment.

PR #1622, opened 2015-05-20
t: fix logic to get temp dir in job host

matthewrmshin:
Using `bash -l` may not work if temp dir gets removed at end of session.


matthewrmshin:
@arjclark could you review this, while @benfitzpatrick is away?

arjclark:
Will do.

arjclark:
Looks ok to me. Test seems to be working fine now.

PR #1630, opened 2015-06-08
rose doc: expand/collapse button fix

kaday:
Only added the missing tags.


kaday:
@arjclark please review

arjclark:
Looks good to me. Tested as working in my environment.

PR #1636, opened 2015-06-18
t: task-env --path=D1 then task-run --path=D2

matthewrmshin:
Ensure that D2 is prepend to PATH.

This change proves that the description of #1608 is invalid.


matthewrmshin:

PR #1644, opened 2015-06-30
rose suite-clean: clean empty top directory

### LINKED ISSUE #1637: rose suite-clean: remove top directory if empty
`rose suite-clean` should remove top level directories if they are empty.

matthewrmshin:
Close #1637.

### DIFF
--- a/lib/python/rose/suite_clean.py
+++ b/lib/python/rose/suite_clean.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-#-----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
 # (C) British Crown Copyright 2012-5 Met Office.
 #
 # This file is part of Rose, a framework for meteorological suites.
@@ -16,11 +16,12 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Rose. If not, see <http://www.gnu.org/licenses/>.
-#-----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
 """Implement the "rose suite-clean" command."""
 
 from glob import glob
 import os
+from pipes import quote
 from rose.config import ConfigLoader, ConfigNode, ConfigSyntaxError
 from rose.env import env_var_process
 from rose.fs_util import FileSystemEvent
@@ -28,7 +29,6 @@ from rose.host_select import HostSelector
 from rose.opt_parse import RoseOptionParser
 from rose.popen import RosePopenError
 from rose.reporter import Reporter
-from rose.resource import ResourceLocator
 from rose.suite_engine_proc import SuiteEngineProcessor, SuiteStillRunningError
 import sys
 import traceback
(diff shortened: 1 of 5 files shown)


kaday:
Works in my environment and passes the test battery.

lib/python/rose/suite_clean.py
```diff
@@ -102,11 +102,37 @@ def _clean(self, suite_name, only_items=None):
...
+                for root in sorted(roots):
+                    cwd = os.getcwd()
+                    if root:
+                        try:
+                            os.chdir(env_var_process(root))
+                        except OSError:
+                            continue
+                    # Reverse sort to ensure that e.g. "share/cycle/" is
+                    # cleaned before "share/"
+                    for name in sorted(self.CLEANABLE_PATHS, reverse=True):
```
arjclark: This doesn't seem particularly friendly/nice/pythonic as a general practice as it makes alterations to CLEANABLE_PATHS problematic at this point in the code. It'd be much better if you altered CLEANABLE_PATHS to be in the desired order of precedence with a note in a comment regarding the importance of maintaining said order.

lib/python/rose/suite_clean.py
```diff
@@ -115,12 +141,30 @@ def _clean(self, suite_name, only_items=None):
...
+                }
+                # Clean empty directories
+                # Change directory to root level to avoid cleaning them as well
+                # For cylc suites, e.g. it can clean up to an empty "cylc-run/"
+                # directory.
+                for root in roots:
+                    names = []
+                    # Reverse sort to ensure that e.g. "share/cycle/" is
+                    # cleaned before "share/"
+                    for name in sorted(self.CLEANABLE_PATHS, reverse=True):
```
arjclark: As above.

lib/python/rose/suite_clean.py
matthewrmshin: Unlike the `List.sort` method, `sorted` does not *alter* the original item, so I am not sure what is the issue here. The constant is used in the logic above in the forward order so items are cleaned in that order by default.

arjclark: I'm aware the sorted operation does not alter the original item. My concern is anyone subsequently working on the code needs to be aware of this requirement at the point of editing the variable in order to not break anything. Any future additions/changes to `CLEANABLE_PATHS` could affect the sorting or have further consequences for the order in which things must be done not necessarily capture-able by a sort operation so it seems appropriate to either define the list in the order it should be cleaned out, and not sort, or put the warning (or some variation there-of) in line with the `CLEANABLE_PATHS` variable.

matthewrmshin: > Any future additions/changes to `CLEANABLE_PATHS` could affect the sorting or have further consequences for the order in which things must be done not necessarily capture-able by a sort operation

The logic here already has a comment, and the `sorted` logic should future-proof any unordered addition to `CLEANABLE_PATHS`. What kind of issues do you have in mind that are not capture-able by a sort?

arjclark: I'm struggling to come up with a specific example with my (apparently) currently limited brain space, rather a worry based on previous experience with this sort of thing going wrong (you may remember the test battery problem where the test worked for you but not for me because our usernames appeared at different points in a sorted list of results :grimacing:). As test battery's working and I can't come up with a specific example other than this has a potential to be a pain in some unforeseen combination of entries that may never exist I'll merge this in.

arjclark:
Tested as working with no unexpected failures.

Issue #1646, opened 2015-07-01
rose bush: take advantage of new tables in cylc suite run time database file


matthewrmshin:
cylc/cylc#1493 and cylc/cylc#1503 introduce 3 new tables to the cylc suite run time database:
* `broadcasts` (see also cylc/cylc#1519)
* `task_jobs`
* `task_job_logs`

Rose Bush should take advantage of them:

* Display current and historical broadcasts in a table. (New page?)
* `task_jobs` should be much easier to query than `task_events`.
  * We should be able to write simpler SQL to get a job listing.
  * We should now be able to easily query timing information (#1308).
* `task_job_logs` will replace the `log_files` table in `log/rose-job-logs.db`.
  This table should be consulted first if it exists.
  We should only need to consult the `log_files` table in `log/rose-job-logs.db`
  to see if the file is in a *tar* archive or is removed.

We should probably have a listing of tasks, with their jobs grouped together.
* Job exit status does not always correspond to the task state.
* Some tasks may have no jobs! (See cylc/cylc#1494 for example.)

matthewrmshin:
The only thing left to do is to take advantage of `task_job_logs`, but this has a dependency on moving job log housekeeping functionality to cylc.

matthewrmshin:
Now that I have thought about it, there is no reason why we have to wait for moving job log housekeeping functionality to cylc. We can make assumption by doing a quick file listing looking for `log/job/CYCLE/` directories and `log/job-CYCLE.tar.gz` files.

PR #1649, opened 2015-07-03
Take advantage of cylc event handler enhancements

matthewrmshin:
rose suite-hook: don't pull remote log if `job.out` already in place.
rose bush:
* base template for cycles, jobs and broadcasts lists.
* cycles list:
  * use `task_jobs` table if available to simplify query.
* jobs list:
  * use `task_jobs` table if available to simplify query.
  * flatten each job entry - the old display leaves a lot of white space.
* broadcasts list: states and events.

Address part of #1646. Dependent on cylc/cylc#1503.


matthewrmshin:
Added basic viewer for broadcast states and events to Rose Bush.

matthewrmshin:
@kaday please review.
@benfitzpatrick or @arjclark please review 2.

kaday:
`cylc files`, `rose files`, `cycles list`, `jobs list: job, job-activity.log, job.err, job.out, job.stats` in Rose bush no longer work :
```
  File "/net/home/h02/kerryday/rose.git/lib/html/rose-bush/view.html", line 53, in template
    {% endblock %}
TemplateSyntaxError: Encountered unknown tag 'endblock'.
```

kaday:
I am not a fan of the adding `1` next to the task in the jobs list when it is the only instance of the task run. When a task has been run more than once then I the submission/attempt number should be next to the task.

kaday:
The `Broadcasts List` table page should display `events` by default and not `states` as the time that the command was issued is very important information.

The hoover over for the time would be more human readable if it was not to so many figures, for example `2015-07-20T11:55:36.125865Z` seems too much detail.

matthewrmshin:
> The Broadcasts List table page should display events by default and not states the time that the command was issued is very important information.

The broadcast states list is a quick and cheap summary of what is going to happen to the tasks in the future. The broadcast events list is the detail of how it gets into that state, and can become quite expensive to display for long running suites. Is it not better to display a summary before going into the detail?

> Further if user has cleared the broadcasts using the -C option then the states page does not exist and you cannot access the events page.

I'll fix this.

> The hover over for the time would be more human readable if it was not to so many figures, for example `2015-07-20T11:55:36.125865Z` seems too much detail.

This is what is recorded in `cylc-suite.db`. I am not entirely sure if it is a good thing to lose the information in case it is useful. (I cannot remember why we store the time stamp in microseconds in the first place.)

matthewrmshin:
Broadcast states list page will now always display a link to the broadcast events list page.

lib/html/rose-bush/broadcast-events.html
```diff
@@ -0,0 +1,50 @@
...
+      <th>point</th>
+      <th>namespace</th>
+      <th>key</th>
+      <th>value</th>
+    </tr>
+  </thead>
+  <tbody>
+    {% for row in broadcast_events %}
+      <tr>
+        <td><abbr class="livestamp" title="{{row[0]}}">{{row[0]}}</td>
```
kaday: missing </abbr>

lib/html/rose-bush/broadcast-events.html
```diff
@@ -0,0 +1,50 @@
...
+          <td><code>{{item}}</code></td>
+        {% endfor %}{# item in row[4:] #}
+      </tr>
+    {% endfor %}{# row in broadcast_events #}
+  </tbody>
+</table>
+</div></div>
+{% endif %}{# broadcast_events #}
+
+</div>
```
kaday: missing </div></div> from matching lines 5 and 7

lib/html/rose-bush/broadcast-events.html
matthewrmshin: Should now be fixed.

lib/html/rose-bush/broadcast-events.html
matthewrmshin: Should also be fixed.

lib/python/rose/bush.py
```diff
@@ -82,6 +82,60 @@ def index(self):
...
+                self.suite_engine_proc.get_suite_state_summary(user, suite))
+        data.update(self._get_suite_logs_info(user, suite))
+        data["broadcast_states"] = (
+            self.suite_engine_proc.get_suite_broadcast_states(user, suite))
+        if form == "json":
+            return simplejson.dumps(data)
+        try:
+            template = self.template_env.get_template("broadcast-states.html")
+            return template.render(**data)
+        except Exception as e:
```
kaday: pylint does not like the use of "e"

lib/python/rose/suite_engine_procs/cylc.py
```diff
@@ -237,6 +278,192 @@ def get_suite_job_events(self, user_name, suite_name, cycles, tasks,
...
+                self.JOB_ORDERS.get(order, self.JOB_ORDERS["time_desc"]))
+        limit_args = []
+        if limit:
+            stmt += " LIMIT ? OFFSET ?"
+            limit_args = [limit, offset]
+        for row in self._db_exec(
+                self.SUITE_DB, user_name, suite_name, stmt,
+                where_args + limit_args):
+            (
+                time,
```
kaday: time is not used

matthewrmshin: Should be good this time.

lib/python/rose/bush.py
matthewrmshin: Replaced `e` with `exc`.

kaday:
I am happy with this now.

Second review needed, over to @benfitzpatrick or @arjclark please review.

arjclark:
Haven't forgotten about this. Will aim to have reviewed today.

arjclark:
Looks like there might be issues with the "events" page with large numbers of entries.

With my local suite "broadcast" the script on the `broadcast -> events` page hangs the browser unless I stop the script.

Perhaps something to split over multiple pages a.la. jobs page?

Otherwise fine.

matthewrmshin:
> Looks like there might be issues with the "events" page with large numbers of entries.

That's true. The current logic is light on the server but heavy on the client. I agree that a paging service is useful. I'll do it as a follow-on to this pull to avoid further delays.

arjclark:
Ok. Will merge in.

Bear in mind that that page will hang now for any of our operational suites so the follow up will need prioritising if you're wanting people to use it.

PR #1657, opened 2015-07-22
rose suite-run: install: handle colon in file name

matthewrmshin:
Improve `rsync` commands in logic.
* The `rose rug-brief-tour` command was far too complicated. All it does is a local copy.
* For `rose suite-run`,
  use `rsync` command's `--exclude=PATTERN` options to exclude items from source,
  instead of listing items in the source directory.

Added a file with colon in its name to the install test.

### DIFF
--- a/bin/rose-rug-brief-tour
+++ b/bin/rose-rug-brief-tour
@@ -21,21 +21,21 @@
 #     rose-rug-brief-tour
 #
 # SYNOPSIS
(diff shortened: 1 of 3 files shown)


arjclark:
Looks ok to me. No unexpected problems from the test battery.

Over to you @kaday for review 2.

kaday:
No unexpected failures in the test battery. Looks ok.

PR #1663, opened 2015-07-27
doc: rose task-env: remove cycle point notes

matthewrmshin:
The notes are no longer required now that we fully support ISO-8601 date
time cycle points.

@kaday please review.


kaday:
Andy eye-balled it, I have fully checked it. It looks okay to me.

PR #1666, opened 2015-08-06
Use hostname for authentication prompts

ScottWales:
When a user first starts `rosie go` it asks to authenticate to the various servers. It requests this using the site prefix, e.g. `e`, which isn't necessarily meaningful to new users.

Instead of using the site prefix prompt users with the full URL, e.g. `https://example.com/rosie`. This way users can easily identify what they're being asked to connect to, and can go to that URL if necessary to learn more about it.

### DIFF
--- a/lib/python/rosie/ws_client_auth.py
+++ b/lib/python/rosie/ws_client_auth.py
@@ -246,8 +246,8 @@ class RosieWSClientAuthManager(object):
         "gnomekeyring": GnomekeyringStore,
         #KeyringStore,
     }
-    PROMPT_USERNAME = "Username for %(prefix)r: "
-    PROMPT_PASSWORD = "Password for %(username)s at %(prefix)r: "
+    PROMPT_USERNAME = "Username for %(host)r: "
+    PROMPT_PASSWORD = "Password for %(username)s at %(host)r: "
     STR_CANCELLED = "cancelled by user"
 
     def __init__(self, prefix, popen=None, prompt_func=None):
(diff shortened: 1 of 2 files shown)


matthewrmshin:
@ScottWales Please ensure that you have read https://github.com/metomi/rose/blob/master/CONTRIBUTING.md and when you are happy, add your name to the list of code contributors as part of this change.

lib/python/rosie/ws_client_auth.py
```diff
@@ -379,7 +379,7 @@ def _prompt(self, is_retry=False):
             if self.username:
                 username = ""
 
-            prompt = self.PROMPT_USERNAME % {"prefix": self.prefix}
+            prompt = self.PROMPT_USERNAME % {"host": self.root}
```
matthewrmshin: The variable is "self.root" here (rightly or wrongly), so we may as well use "root" as the key here.

lib/python/rosie/ws_client_auth.py
```diff
@@ -246,8 +246,8 @@ class RosieWSClientAuthManager(object):
         "gnomekeyring": GnomekeyringStore,
         #KeyringStore,
     }
-    PROMPT_USERNAME = "Username for %(prefix)r: "
-    PROMPT_PASSWORD = "Password for %(username)s at %(prefix)r: "
+    PROMPT_USERNAME = "Username for %(host)r: "
```
matthewrmshin: It would be better to retain the prefix:

"Username for %(prefix)r - %(root)r: "

ScottWales:
@matthewrmshin No worries, have added your suggestions

matthewrmshin:

arjclark:
Reviewing now.

arjclark:
Looks good to me. No problems from the test battery.

PR #1696, opened 2015-09-14
1531: rose bunch: run multiple commands in one app

arjclark:
@matthewrmshin - not ready for merge yet (still several things to do and test). Can you take a preliminary look and see what you think?


matthewrmshin:
It will be good if we can some examples of the user interface, which is likely to trigger the most debate.

We need a better name for this built-in application. How about *rose_bundle*?

matthewrmshin:
You probably should return a non-zero code if any of your commands returns a non-zero code.

There are various style issues that is better discussed offline. Some examples:
* Can use `rose.reporter` to report information and errors.
* `RosePopener`is not being used in the normal way.
* `shlex.split` list is normally better than comma-separated list for shell command related strings. Many commands have `,`in option values.
* Is a named key preferable to an index?

arjclark:
> It will be good if we can some examples of the user interface, which is likely to trigger the most debate.

No doubt, at this point though I'm checking it does things correctly from a functional point of view (interface is ultimately just window dressing) before opening it up to everyone for comment.

> We need a better name for this built-in application. How about rose_bundle?

Agreed, ultimately a placeholder name right now. I don't hate *rose_bundle* either.

> Can use `rose.reporter` to report information and errors

Will do.

> `RosePopener` is not being used in the normal way.

I'll discuss this offline

> `shlex.split` list is normally better than comma-separated list for shell command related strings. Many commands have `,` in option values.

Will swap that in.

> Is a named key preferable to an index?

I've gone with a default of just using an index but I'd be happy enough for the user to be able to supply a set of names.

arjclark:
@matthewrmshin - I've made a number of initial refinements based on the above feedback and pushed them up here.

Can you have a look and then we can discuss what needs changing? 

An example suite I've been using to test the code with is in my home directory under:

 `/OtherSuites/examples/multitest`

matthewrmshin:
I was thinking about this while going home yesterday. I think we may have approached this from the wrong angle. Because we are running external commands, it may actually be easier to manage our own pool of processes. This will allow you to use all the facilities (e.g. popen, reporter) handed down to you from the `app_runner` argument. You can also have a lot more control of whether to die at the earliest failure or to run as many commands as possible. Perhaps something like this?

```python
# Pseudo code
max_procs = whatever
commands = {}  # populate with key: command
procs = {}
while procs or commands:
    while procs:
        for key, proc in procs.items():
            if proc.poll():
                procs.pop(key)
                # ... handle results
    while len(procs) < max_procs and commands:
        key, command = commands.popitem()
        proc[key] = app_runner.popen.run_bg(command, ...)
    sleep(for_a_while)
```

arjclark:
Sounds like a good idea. It'll also allow configuration on the user's part of when they want the task itself to fail.

Will have a play.

arjclark:
@matthewrmshin - I've now got everything I want implemented done in here.

No documentation at present, but current app conf settings are as follows:

```
[bunch-settings] # Section for settings
command-format= some command %(arg1)s %(arg2)s # compulsory
limit=X # Number of commands maximum to run at once (default=5)
names = name1 name2 ... # Optional: allows naming of particular invocations in reporting rather than index based names
abort-on-fail = True # Optional: allows stopping running any more invocations on failure (default=False)
incremental = True # Optional: allows use of incremental mode to only re-run previously failed tasks (default=False). In incremental mode, if the app config has changed since the last run then all command invocations will be re-run. 

[bunch-args] # section for defining lists of args to command
arg1 = foo bar # space separated args to iterate over - number of elements in each arglist must be the same and all args must contain the same number of items
arg2 = baz qux
...
```
Example usage in my home space under /OtherSuites/examples/multitest

@benfitzpatrick @kaday - please take a look at the interface and see what you think (not asking for a code review just whether you think the setting up of running the app makes sense).

kaday:
Can we call it `rose_bunch` rather than `rose_bundle` ?

kaday:
I think the setting up of running the app makes sense.

arjclark:
@dpmatthews - any thoughts on the interface as described above (https://github.com/metomi/rose/pull/1696#issuecomment-144731314)?

@matthewrmshin - If I've had no objections by Wednesday I'll assume everyone's happy with the interface and proceed to document and write tests for formal review.

arjclark:
Discussing with @dpmatthews today will make the following changes/additions:

 * `limit` - default setting will be to run "all" configurations rather than magic number 5.
 * need to check how rose bush handles output files
 * `abort-on-fail` - will change to `fail-handle` with options `continue` and `abort`. This would allow possible future extension to allow `kill` functionality if usecases appear
 * `instances` setting - add a setting that will generate #instances. Caters for where the only difference would be a number. Any arglists will be validated against this to check their length is equivalent. Will reserve an arg keyword to go with it. I intend to start this number from 0 (so 0...9 rather than 1...10).
 * Won't worry about separate working directories at this point. That can be initially handled by command wrapper scripts where needed (all they need do is create a subdirectory to work in). If use cases appear we can implement as needed.

arjclark:

arjclark:
N.B. I'll write a tutorial as a followup pull request.

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,436 @@
+import os
```
matthewrmshin: Where is the header?

arjclark: Added.

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,114 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+    <kbd>[bunch-settings]</kbd> and <kbd>[bunch-args]</kbd> sections in
+    <var>rose-app.conf</var>.</p>
+
+    <p>Each variant of the command is run in the same working directory with
+    its output directed to separate <var>.out</var> and <var>.err</var> files
+    of the form <var>bunch.&lt;name&gt;.out</var>. Should you need separate working
+    directories your should configure your command to create the appropriate
+    subdirectory for working in.</p>
+
+    <p>Note that, when running under batch submission systems, you will need
```
matthewrmshin: What about "... under load balancing systems such as PBS or Slurm, you ..."?

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,114 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+    to set resource requests to reflect the resources required by running
+    multiple commands at once e.g. if one command would require 1GB memory and
+    you have configured your app to run up to 4 commands at once then you will
+    need to request 4GB of memory.</p>
+
+    <p>The following <kbd>[bunch-settings]</kbd> settings are accepted:</p>
```
matthewrmshin: I think the `-settings` bit is redundant, given that all entries in a configuration file are settings. (Also, without the `-settings` bit, the `[bunch]` section will be automatically sorted to live before `[bunch-args]`.)

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,114 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+      <dt><kbd>command-format=FORMAT</kbd></dt>
+
+      <dd>Compulsory. A Pythonic <code>printf</code> style format string to
+      construct the commands to run. Insert placeholders <var>%(ARGNAME)s</var>
+      for substitution of the arguments specified under <kbd>[bunch-args]</kbd>
+      to the invoked command. The placeholder <var>%(instances)s</var> is
+      reserved for inserting an automatically generated index for the command
+      invocation when using the <kbd>instances</kbd> setting.</dd>
+
+      <dt><kbd>instances=N</kbd></dt>
```
matthewrmshin: What about `command-instances`?

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,114 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+
+      <dd>Optional. Allows the user to specify an integer value for the number
+      of instances of a command they want to run. This generates the values
+      used by the <var>%(instances)s</var> value in <kbd>command-format</kbd>
+      Useful for cases where the only difference between invocations would be
+      an index number e.g. ensemble members. Note indexes start at 
+      <var>0</var>.</dd>
+
+      <dt><kbd>limit=N</kbd></dt>
```
matthewrmshin: Would `pool-size` be clearer?

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,114 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+
+      <dd>Optional. Allows the user to limit the number of concurrently running
+      commands. If not specified then all command variations will be run at
+      the same time.</dd>
+
+      <dt><kbd>fail-handle=continue|abort</kbd></dt>
```
matthewrmshin: This looks like a fail mode, rather than a fail handle. (A handle may be confused as a piece of logic to handle the failure.)

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,114 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+
+      <dd>Optional (default=continue). Specify what action you want the job to
+      take on failure of a command it is trying to run. If set to
+      <var>continue</var> it will run all command variants . If set to
+      <var>abort</var> then on failure of on of the command variants it will
+      stop trying to run any new variants N.B. the job will wait for any
+      already running commands to finish before exiting.</dd>
+
+      <dt><kbd>incremental=True|False</kbd></dt>
```
matthewrmshin: Should this just be handled by the `--new` option of the command line?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+    ConfigValueError,
+    CompulsoryConfigValueError)
+from rose.popen import RosePopener
+import rose.job_runner
+from rose.reporter import Event
+
+
+class AbortEvent(Event):
+
+    """An event raised when a task failure will stop further task running"""
```
matthewrmshin: This looks like an event for ERR instead of OUT?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+
+class FailEvent(Event):
+
+    """An event used to report failure of a task in the pool."""
+
+    KIND = Event.KIND_ERR
+
+    def __str__(self):
+        name, rc = self.args
+        return "Command %s failed with return code %s" % (name, rc)
```
matthewrmshin: Is this not just a repeat of `rose.popen.RosePopenError`?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+
+class InconsistentArglength(Exception):
+
+    """An event raised when arg lengths are not consistent"""
+
+    ERROR_MSG = "Inconsistent arg lengths"
+
+    def __str__(self):
+
+        return  self.ERROR_MSG
```
matthewrmshin: Can this be a `ConfigValueError`?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+class LaunchEvent(Event):
+
+    """An event raised when adding a command to the task pool."""
+
+    LEVEL = Event.V
+    KIND = Event.KIND_OUT
+
+    def __str__(self):
+        name, cmd = self.args
+        return "Adding command %s to pool: %s" % (name, cmd)
```
matthewrmshin: You may want to use `rose.popen.RosePopenEvent` instead?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+class SucceededEvent(Event):
+
+    """An event used to report success of a task in the pool."""
+
+    LEVEL = Event.V
+    KIND = Event.KIND_OUT
+
+    def __str__(self):
+        name, rc = self.args
+        return "Command %s completed successfully # rc = %s" % (name, rc)
```
matthewrmshin: I think a simpler `[ OK ] command ...` will be sufficient, `rose_ana` style. (If I remember correctly, `rose.reporter` allows you to specify an alternate `prefix` when reporting an event.)

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+
+class SkippingEvent(Event):
+
+    """Event raised when a command is skipped in incremental mode"""
+    LEVEL = Event.V
+    KIND = Event.KIND_OUT
+
+    def __str__(self):
+        name = self.args
+        return "Skipping %s: previously ran and succeeded" % (name)
```
matthewrmshin: Again, I think the best way to report is a simple `[SKIP] command ...` string.

matthewrmshin: Or `[DONE] command ...`?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+    FAIL_HANDLE_TYPES = [TYPE_CONTINUE_ON_FAIL, TYPE_ABORT_ON_FAIL]
+    FAIL_HANDLE = TYPE_CONTINUE_ON_FAIL
+    INCREMENTAL = False
+
+    def run(self, app_runner, conf_tree, opts, args, uuid, work_files):
+        """ Run multiple instaces of a command using sets of specified args"""
+
+        # Allow naming of individual calls
+        self.invocation_names = conf_tree.node.get_value(
+                                            [self.SETTINGS_SECTION, "names"])
```
matthewrmshin: May not pass PEP8.

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+        if self.invocation_names:
+            self.invocation_names = shlex.split(self.invocation_names)
+            if len(set(self.invocation_names)) != len(self.invocation_names):
+                raise ConfigValueError([self.SETTINGS_SECTION, "names"],
+                                       self.invocation_names,
+                                       "names must be unique")
+            if "job" in self.invocation_names:
+                raise ConfigValueError([self.SETTINGS_SECTION, "names"],
+                                       self.invocation_names,
+                                       "may not name an invocation 'job'")
```
matthewrmshin: Why not?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+
+        fail_handle = conf_tree.node.get_value([self.SETTINGS_SECTION,
+                                          "fail-handle"])
```
matthewrmshin: Can set a default here so you don't have to do the next `if`.

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+        if fail_handle:
+            if fail_handle in self.FAIL_HANDLE_TYPES:
+                self.FAIL_HANDLE=fail_handle
```
matthewrmshin: Setting a constant?

lib/python/rose/apps/rose_bunch.py
arjclark: Leftover code from previous handling of job output code - not needed anymore.

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+                                    [self.ARGS_SECTION, "instances"]),
+                                   "reserved keyword")
+
+        # Set up instances if needed
+        instances = conf_tree.node.get_value([self.SETTINGS_SECTION, "instances"])
+
+        if instances:
+            try:
+                instances = int(instances)
+                instances = range(instances)
```
matthewrmshin: Can be a single statement?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+                        self.dao.add_command(key)
+
+                app_runner.handle_event(LaunchEvent(key, cmd))
+                procs[key] = app_runner.popen.run_bg(
+                                                cmd,
+                                                shell=True,
+                                                stdout=open(cmd_stdout, 'w'),
+                                                stderr=open(cmd_stderr, 'w'))
+            sleep(self.SLEEP_DURATION)
+
```
matthewrmshin: Report commands not run on abort?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+        return self.OUTPUT_TEMPLATE % (self.name, "err")
+
+
+class RoseBunchDAO(object):
+
+    """Database object for rose_bunch"""
+
+    T_COMMANDS = "commands"
+    T_CONFIG = "config"
```
matthewrmshin: @benfitzpatrick will probably argue that we ought to say `TABLE` something or something `TABLE`.

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+    def create_tables(self):
+        """Create tables as appropriate"""
+        existing = []
+
+        for row in self.conn.execute(
+            "SELECT name FROM sqlite_master WHERE type=='table' ORDER BY name"):
+            existing.append(row[0])
+
+        if existing:
+            self.NEW_RUN = False
```
matthewrmshin: Assigning a constant?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+        return has_succeeded
+
+    def flatten_config(self, config):
+        """Flatten down a config node to set of keys and values for recording"""
+        flat = {}
+        keys_and_nodes = list(config.node.walk())
+        for keys, node in keys_and_nodes:
+            if not isinstance(node.value, dict):
+                if node.state not in [rose.config.ConfigNode.STATE_USER_IGNORED,
+                                      rose.config.ConfigNode.STATE_SYST_IGNORED]:
```
matthewrmshin: Use `node.is_ignored()`?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+                    flat["_".join(keys)] = node.value
```
matthewrmshin: Why not `[x]y` or `x=y` (configuration metadata style) for keys?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+
+        for key, value in res.items():
+            args.append((key, value))
+
+        i_stmt = "INSERT OR REPLACE INTO " + self.T_CONFIG + " VALUES (?, ?)"
+        self.conn.executemany(i_stmt, args)
+        self.conn.commit()
+        return
+
+    def same_last_config(self, current):
```
matthewrmshin: `cmp_prev_config`? (The word `previous` is normally clearer in meaning than `last`.)

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,458 @@
...
+        for index, name in enumerate(self.invocation_names):
+            invocation = RoseBunchCmd(name, self.command_format, index)
+            for key, vals in sorted(multi_args.items()):
+                invocation.argsdict[key] = shlex.split(vals.value)[index]
+            if instances:
+                invocation.argsdict["instances"] = instances[index]
+            commands[name] = invocation
+
+        procs = {}
+        log_format = os.path.join(os.environ['ROSE_TASK_LOG_DIR'], "%s")
```
matthewrmshin: I guess this is the only point where this application will work with `rose task-run` and not `rose app-run`? Perhaps it is worth having a logic where if this environment variable is not set, then the logs should go to the current working directory?

t/rose-task-run/31-app-bunch/app/bunch_incremental/rose-app.conf
```diff
@@ -0,0 +1,11 @@
...
+meta=rose_bunch
+
+[bunch-args]
+arg1=true false true
+
+[bunch-settings]
+command-format=%(arg1)s
+limit=1
+incremental=True
+fail-handle=continue
```
matthewrmshin: No tests for the `names` option?

t/rose-task-run/31-app-bunch/app/bunch/rose-app.conf
```diff
@@ -0,0 +1,11 @@
+mode=rose_bunch
+meta=rose_bunch
+
+[bunch-args]
+arg1=1 2 3 4
+
+[bunch-settings]
+command-format=echo arg1: %(arg1)s
+limit=1
+instances = 4
```
matthewrmshin: Do we want to have a bigger population? If only to test what happens when we have more than 1 digit.

lib/python/rose/apps/rose_bunch.py
arjclark: Would something like "[WARN] Not run: <command>" be suitable? Not sure if stdout or stderr is the place for it though (strictly speaking not an error but has happened as a result of one) - which do you think?

lib/python/rose/apps/rose_bunch.py
arjclark: Similar, but this is an Event rather than an Exception so as not to end up exiting part way through running through the pool.

lib/python/rose/apps/rose_bunch.py
matthewrmshin: Given that you have redirected the STDOUT and STDERR of individual commands to alternate locations, it is probably worth using the STDOUT of the built-in application to report the status of the commands. In this case, a status line to indicate that a command was never started should be sufficient.

arjclark: Having had a think about this, I think the best thing to do for stating if a command has been run or not would be to have that in the output from the main job rather than the individual commands themselves. Reasoning being, the job is acting as a controller and has (for whatever reason) decided not to run a particular command. Whereas, if there is output from a particular command that *has been run* that should go to its own out/err files.

lib/python/rose/apps/rose_bunch.py
arjclark: Having had a think, I'm going to keep this event but have it report in normal verbosity as it is something the controller should be reporting as it goes.

lib/python/rose/apps/rose_bunch.py
arjclark: Changed.

doc/rose-rug-task-run.html
arjclark: I prefer this way with the option of being able to configure the app to run in incremental mode or not - you may not want to re-do the file extract/install elements of the app every time you re-run all your commands for instance.

arjclark:
@matthewrmshin - feedback addressed and/or commented on, back to you

doc/rose-rug-task-run.html
matthewrmshin: Fair enough. For consistency with other settings (e.g. configuration metadata settings that are boolean), I think we should use lower case `true` and `false` here.

lib/python/rose/apps/rose_bunch.py
matthewrmshin: You can report a `rose.popen.RosePopenError` without exiting.

doc/rose-rug-task-run.html
arjclark: Done

doc/rose-rug-task-run.html
arjclark: Done

doc/rose-rug-task-run.html
arjclark: Done

doc/rose-rug-task-run.html
arjclark: Done

doc/rose-rug-task-run.html
arjclark: Changed

doc/rose-rug-task-run.html
arjclark: Changed as discussed

lib/python/rose/apps/rose_bunch.py
arjclark: My misunderstanding. Done now.

lib/python/rose/apps/rose_bunch.py
arjclark: Aha, that's what I was hoping there was!

lib/python/rose/apps/rose_bunch.py
arjclark: Now added in.

t/rose-task-run/31-app-bunch/app/bunch_incremental/rose-app.conf
arjclark: Now added.

t/rose-task-run/31-app-bunch/app/bunch/rose-app.conf
arjclark: Now added (bunch_bigpop)

arjclark:
@matthewrmshin - think I've now got everything :grimacing:

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,114 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+    you have configured your app to run up to 4 commands at once then you will
+    need to request 4GB of memory.</p>
+
+    <p>The following <kbd>[bunch]</kbd> settings are accepted:</p>
+
+    <dl>
+      <dt><kbd>command-format=FORMAT</kbd></dt>
+
+      <dd>Compulsory. A Pythonic <code>printf</code> style format string to
+      construct the commands to run. Insert placeholders <var>%(ARGNAME)s</var>
```
matthewrmshin: `%(ARGNAME)s` should be lower case?

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,114 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+      stop trying to run any new variants N.B. the job will wait for any
+      already running commands to finish before exiting.</dd>
+
+      <dt><kbd>incremental=true|false</kbd></dt>
+
+      <dd>Optional (default=false). If set to <var>true</var> then only failed
+      commands will be re-run on retrying running of the job. If any changes
+      are made to the configuration being run then all variants will be re-run.
+      Similarly, running the app with the <kbd>--new</kbd> option to <kbd>rose
+      task-run</kbd> will result in all commands being run.</dd>
```
matthewrmshin: Any reason why incremental mode is not the default?

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,444 @@
...
+    PREFIX_OK = "[OK]"
+    PREFIX_SKIP = "[SKIP]"
+
+
+    def run(self, app_runner, conf_tree, opts, args, uuid, work_files):
+        """ Run multiple instaces of a command using sets of specified args"""
+
+        # Allow naming of individual calls
+        self.invocation_names = conf_tree.node.get_value(
+                                            [self.BUNCH_SECTION, "names"])
```
matthewrmshin: This will not pass PEP8.

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,444 @@
...
+
+class NotRunEvent(Event):
+
+    """An event used to report commands that will not be run."""
+
+    KIND = Event.KIND_OUT
+
+    def __str__(self):
+        name, cmd = self.args
+        return " Not run %s: %s " % (name, cmd)
```
matthewrmshin: Can we have a more consistent set of reports for statuses, (preferably parse-able)? E.g.:
* Use the same style as `rose_arch` where each target is reported using a status symbol.
  * (I normally prefer this language-neutral style as we have users who are non-native English speakers.)
* Use a style such as `name: message\n\tcommand`.

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,444 @@
...
+                                     None, KeyError("command-format"))
+
+        # Set up command-instances if needed
+        instances = conf_tree.node.get_value([self.BUNCH_SECTION,
+                                              "command-instances"])
+
+        if instances:
+            try:
+                instances = range(int(instances))
+            except ValueError as err:
```
matthewrmshin: `err` not used.

lib/python/rose/apps/rose_bunch.py
```diff
@@ -0,0 +1,444 @@
...
+        if self.conn is None:
+            self.conn = sqlite3.connect(self.db_file_name, self.CONN_TIMEOUT)
+        return
+
+    def create_tables(self):
+        """Create tables as appropriate"""
+        existing = []
+
+        for row in self.conn.execute(
+            "SELECT name FROM sqlite_master WHERE type=='table' ORDER BY name"):
```
matthewrmshin: I don't think we need an ordering here.

t/rose-task-run/31-app-bunch/suite.rc
```diff
@@ -0,0 +1,39 @@
+#!jinja2
+[cylc]
+    UTC mode=True
+    [[event hooks]]
+        timeout handler=rose suite-hook --shutdown
+        timeout=1
```
matthewrmshin: Should use ISO8601 style for new test?

doc/rose-rug-task-run.html
arjclark: No reason, it can be if you think that'd be better.

doc/rose-rug-task-run.html
arjclark: Not 100% here, went with ARGNAME rather than argname as they're ARGs which would typically be presented in uppercase in command help.

lib/python/rose/apps/rose_bunch.py
arjclark: So, to be clear, you'd prefer the message to be reading like:
```
[SKIP] foo: Not run
            my_command with args
```
?

lib/python/rose/apps/rose_bunch.py
arjclark: What're you using to verify PEP8 "goodness"?

lib/python/rose/apps/rose_bunch.py
arjclark: Fair enough

t/rose-task-run/31-app-bunch/suite.rc
arjclark: Was written to be stylistically consistent with the other rose-task-run tests. I can update to ISO style if you prefer?

lib/python/rose/apps/rose_bunch.py
matthewrmshin: The `pep8` command, which is now used by our test battery on master.

arjclark: Ta, running through it now.

arjclark:
@matthewrmshin - can you confirm desired style of output is as I understand above and what your thoughts on the capitalisation of ARGNAME is and I'll make the appropriate changes.

doc/rose-rug-task-run.html
matthewrmshin: I'll stick with lower case. (I think you have used lower case for this in other part of the document.)

lib/python/rose/apps/rose_bunch.py
matthewrmshin: Status line should be as concise as possible. So, for example, if you already say SKIP, you probably don't even need *Not run*, unless it is necessary to explain why it is not run. In this built-in application, I can imagine 4 statuses: success, fail, success already in previous run, not run due to abort on failure. All we need is 4 status symbols or some kind of tags.

arjclark:
@matthewrmshin - I think the outputs should now be suitably short. Have documented accordingly as well.

matthewrmshin:
Looks OK 1.

@kaday Please sanity check.

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,118 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+    configuration.</p>
+
+    <p>The application is normally configured in the
+    <kbd>[bunch]</kbd> and <kbd>[bunch-args]</kbd> sections in
+    <var>rose-app.conf</var>.</p>
+
+    <p>Each variant of the command is run in the same working directory with
+    its output directed to separate <var>.out</var> and <var>.err</var> files
+    of the form <var>bunch.&lt;name&gt;.out</var>. Should you need separate
+    working directories your should configure your command to create the
```
kaday: you not your

doc/rose-rug-task-run.html
```diff
@@ -692,6 +692,118 @@ <h2 id="rose-task-run.util.rose_arch">Built-in Application: rose_arch</h2>
...
+
+      <dt><kbd>pool-size=N</kbd></dt>
+
+      <dd>Optional. Allows the user to limit the number of concurrently running
+      commands. If not specified then all command variations will be run at
+      the same time.</dd>
+
+      <dt><kbd>fail-mode=continue|abort</kbd></dt>
+
+      <dd>Optional (default=continue). Specify what action you want the job to
```
kaday: Doesn't read clearly to me, I think it need rewording.

kaday:
Looks okay to me and passes test battery in my environment.

PR #1699, opened 2015-09-21
Emacs rose-conf-mode: correct installation instructions & highlight rose-suite.info files

dpmatthews:
Prevent warning from emacs:
http://stackoverflow.com/questions/24779041/disable-warning-about-emacs-d-in-load-path


dpmatthews:
@stevewardle please review

stevewardle:
Looks reasonable to me - interestingly enough this is the first I've heard about it being dangerous to add .emacs.d to the load path, because in my own setup purely by coincidence I don't do it (I use sub-directories) - I suppose I must have assumed it was a sensible "default location" to tell users to install the files

PR #1708, opened 2015-09-30
rose_arch: allow source to be optional

### LINKED ISSUE #1707: rose arch to not fail if (some) sources are missing
At present, rose arch will fail if any sources are not found.

This can be a problem if a file is missing for a valid reason, for instance an observation type is unavailable for a particular cycle.

It would be very helpful if there was a syntax in rose arch to specify that a particular source should not cause rose arch to fail if the source cannot be found. Somthing like:

[arch:$ARCHIVING_PREFIX/${ROSE_TASK_CYCLE_TIME}_${ROSE_TASK_PREFIX}_aircraftsonde_odb2.gz]
source=glu_odb2/aircraftsonde.odb
skip-if-source-missing=True

This should be done on a source by source basis, and default to fail when missing, as there will be some sources that we always want the archving to fail on if they are not available.

matthewrmshin:
Using the `(NAME)` or `(GLOB)` syntax.

Close #1707.

@arjclark @kaday please review. You can now do:

```cfg
[arch:hello/]
sources=earth.txt (mars.txt) (venus.txt)
```

### DIFF
--- a/doc/rose-rug-task-run.html
+++ b/doc/rose-rug-task-run.html
@@ -589,7 +589,11 @@ opt.jobs=8
       using quotes or backslashes, like in a shell.)  Paths, if not absolute
       (beginning with a <var>/</var>), are assumed to be relative to
       <var>$ROSE_SUITE_DIR</var> or to <var>$ROSE_SUITE_DIR/PREFIX</var> if
-      <kbd>source-prefix=PREFIX</kbd> is specified.</dd>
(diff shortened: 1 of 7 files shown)


arjclark:
Looks ok to me but currently unable to test fully during system outage. Will await service restoration.

arjclark:
Looks ok to me. Getting some unrelated test battery problems but they're problems for me on master too.

Over to @kaday

kaday:
Works in my environment. Passes test battery.

PR #1718, opened 2015-10-12
rose suite-run: fix hang up of svn diff

### LINKED ISSUE #1717: rose suite-run: can hang if "svn diff" waits for input
A user has `diff-cmd = vimdiff` in their `~/.subversion/config`. If the suite is a Subversion working copy with changes, this can cause `rose suite-run` or `rose stem` to hang as `vimdiff` waits for user input.

matthewrmshin:
Call to `svn diff` may hang if user has an alternate diff command
configured that hangs. This change should force `svn diff` to use its
internal diff logic.

Fix #1717.

### DIFF
--- a/lib/python/rose/run_source_vc.py
+++ b/lib/python/rose/run_source_vc.py
@@ -43,18 +43,23 @@ def write_source_vc_info(run_source_dir, output=None, popen=None):
         handle = open(output, "wb")
     environ = dict(os.environ)
     environ["LANG"] = "C"
-    for vcs, cmds in [("svn", ["info", "status", "diff"]),
-                      ("git", ["describe", "status", "diff"])]:
+    for vcs, args_list in [
+            ("svn", [
+                ["info", "--non-interactive"],
+                ["status", "--non-interactive"],
+                ["diff", "--internal-diff", "--non-interactive"]]),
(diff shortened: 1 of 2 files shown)


lib/python/rose/run_source_vc.py
```diff
@@ -43,18 +43,23 @@ def write_source_vc_info(run_source_dir, output=None, popen=None):
         handle = open(output, "wb")
     environ = dict(os.environ)
     environ["LANG"] = "C"
-    for vcs, cmds in [("svn", ["info", "status", "diff"]),
-                      ("git", ["describe", "status", "diff"])]:
+    for vcs, args_list in [
+            ("svn", [
+                ["info", "--non-interactive"],
```
arjclark: I take it these "--non-interactive" arguments are disabling the prompt that was causing the problem?

matthewrmshin: It will help. The main fix is the `--internal-diff` option for `svn diff`.

arjclark:
Looks fine to me. No new problems in test battery on this branch vs. master.

Over to @kaday

kaday:
Looks good to me and passes all test in test-battery in my environment.

PR #1732, opened 2015-10-19
rosie disco: improve templates

### LINKED ISSUE #1142: Rosie Web should have same style as Rose Bush
Some bullet points:
- [x] Use bootstrap.
- [x] Display Rose version.

matthewrmshin:
Now use the same style as Rose Bush.

Use Bootstrap.
Display host name and Rose version.

Close #1142.


matthewrmshin:
Branch re-based.

matthewrmshin:

lib/html/static/css/dataTables.bootstrap.css
```diff
@@ -0,0 +1,319 @@
...
+table.table-bordered td:last-child {
+	border-right-width: 0;
+}
+div.dataTables_scrollHead table.table-bordered {
+	border-bottom-width: 0;
+}
+
+
+
+
```
arjclark: Rather excessive number of newlines don't you think?

matthewrmshin: This is an unmodified source file from an external library, so we should not touch it.

arjclark:
@matthewrmshin - Initial feedback from having a play:

In "search mode":

 * The Search box (as opposed to Search Term) is unclear in its purpose. It took me a while to figure out that is was filtering the results from running a test having filled in the Search Term box. Would recommend renaming to something like "filter results"

In "query mode":

 * The "group controls" box should probably be labelled something like "display group controls" and made into a toggle button for that rather than a search parameter like "all revisions". Reasoning being that you don't affect the query, you affect your ability to do something when you configure your query.

 * De-selecting "group controls" having added some in seems to create odd behaviour. Set it up with three search lines where the first line opens a single bracket at the start and the second line closes it with a single bracket, third line should be unbracketed. De-select "group controls" and you'll see the first line's bracket boxes disappear but the subsequent line ones (including blank selection boxes) remain 

 * Can't remove the first line of the search criteria - should be able to a.la rosie go. so that you can remove any line but must always have one (so only disable button when there's a single criterion left)


Swapping from "search" to "query" mode and vice-versa:

 * Not convinced the previous query/search results should remain displayed on changing mode.

Layout:

 * The search and query mode links should be more prominent, probably taking the form of buttons, and with some way of seeing quickly which mode you're in just looking at them (e.g. if both are buttons, colour the active one blue.

* Do we want to consider splitting results over multiple pages like we do for rose bush? (e.g. 300 per page)

@kaday - I think it'd be worth you having a play around at this point too to see what you think of the interface prior to the "review proper"

matthewrmshin:
@arjclark I have improved the display somewhat. Although I agree with most of the other recommendations, I don't think they are in the scope of this change, which is primarily a simple presentation change instead of a logical change.

Recommendations done:
* The search box is now removed. It was due to a bad version of jQuery DataTables.
* The group controls check box is now a button.
* The search/query links are now part of the navigation bar.

Recommendations not done:
* Removal of 1st line of query.
  * (I believe this has been discussed before. The HTML form indexing makes this very difficult on existing logic. If it becomes a big deal in the future, we can rip out the logic and try again.)
* Display of results on changing mode - `rosie go` behaves similarly.
* Paging of results - will require work on the server side. Out of scope.

arjclark:
> Recommendations not done: ...

OK, captured at #1746 for later.

kaday:
@matthewrmshin 

Some initial comments:
I think the `?` hover over for each suite id is ace.
However, I was initially confused by it and why very suite had a `?` next to it. Could you change it to one of the following ideas or something else if you can think of an alternative: Change the symbol next the suite. Change the header. Provide a hover over for the header `?` (this was the first place I looked to try to figure out what it meant).

`Show / hide columns` disappears, I would prefer it did not or if it is to disappear can you please increase the duration that it is displayed for.

In `query` mode could the `all revisions` and `group controls` either be underneath each other rather than side-by-side or could you put a gap between the `+` for adding a search query and the `all revisions` option.

In `query` if I add grouped queries and then try to remove them I can't always remove them from the display.

matthewrmshin:
The `?` column is now the `detail` column. The cell now contains a list icon.

> In query mode could the all revisions and group controls either be underneath each other rather than side-by-side or could you put a gap between the + for adding a search query and the all revisions option.

The display does not look like what is described above?

> In query if I add grouped queries and then try to remove them I can't always remove them from the display. 

I believe this is in the old logic, and is not part of this change.

kaday:
@matthewrmshin sorry my misunderstanding ignore all of my comments so far.

matthewrmshin:
The only thing that is missing is the `Show / hide columns` functionality. I am a bit reluctant to put it back in, as the ColVis library that we were using has been retired, and the old one we have does not appear to work well under Bootstrap 2. The functionality has moved to a library that requires that we move up to Bootstrap 3 (which we should do, but probably not as part of this change).

arjclark:
2 bugs currently noted:

1) Oddly, whatever I searched for last on a repo keeps getting run when I start a new session. So, e.g. search for test on repo1, swap to repo2 and so something, swap back to repo1 and it'll re-run the last search on repo1, similarly swapping back to repo2 will give you the last search on there. Navigating away and then coming back from a different window doesn't get rid of the behaviour.

2) Toggling display group controls as per the above reported "query mode" oddity now noticeable hangs my browser and is now leaving the bracket selection boxes on the last entry line (but at they are at least gone from the first two lines)

matthewrmshin:
Bug 1 is definitely on master. It comes from a set of very naive cookies settings.

matthewrmshin:
Bug 1 should now be fixed.

matthewrmshin:
Bug 2 should now be fixed.

arjclark:
Bug 2 fixed. Bug 1 still present when opening the page for a repository that's previously been searched, though if its already a problem then it isn't directly the problem of this change.

arjclark:
Confirmed 1 a problem in master.

matthewrmshin:
Note that the in-correctness of bug 1 has been resolved. The intention has always been that on loading, it repeats the previous search. It was not repeating the search per prefix. The fix is to ensure that a search is only repeated for the correct prefix.

arjclark:
> Note that the in-correctness of bug 1 has been resolved. ... It was not repeating the search per prefix. The fix is to ensure that a search is only repeated for the correct prefix.

Yup, confirmed as now working. It being intentional to repeat the previous search wasn't something I knew (its odd behaviour)

matthewrmshin:
(Otherwise, we would not have bothered with putting in a set of cookies.)

arjclark:
As an aside of sorts, are you intending on adding automated tests for this in a similar manner to what you've done for rose bush?

matthewrmshin:
I believe it already has automated tests for the data layer, but not the presentation layer. The latter is a lot more difficult (although not impossible these days) like the GUIs. (The new Rose Bush tests are tests for the data layer, not the presentation layer.)

arjclark:
Cool.

arjclark:
Looks good to me. Tests pass in my environment and manual interaction now all seems fine. Over to @kaday for review 2.

kaday:
In the `query` drop down it gives the option of `idx` where as the table uses the header of `id` can these please be consistent.

kaday:
One of my `status` has `MM` given, what does this mean? Is it correct or should it just be `M`?

matthewrmshin:
The 1st M status tells us that the suite is modified at the given revision. The 2nd M status tells us that the `rose-suite.info` file is modified at the given revision.

matthewrmshin:
Column name is now idx.

kaday:
Is the `MM` meaning documented anywhere? or could we have a hover over staying what it means?

matthewrmshin:
@kaday While I agree that the status column meaning can be documented better, it is not caused by this change. We have the same problem in other rosie clients. Switch on view status column in `rosie go` or switch on `--format=%status`, and you will be presented with exactly the same thing. If you think it needs fixing, then we should address that in a separate change.

kaday:
@matthewrmshin agreed. I am now happy that all comments have been addressed, thanks.

PR #1731, opened 2015-10-19
rose bush: jobs: cylc-suite.db task_job_logs table

matthewrmshin:
Rose Bush jobs view can now use the `task_job_logs` table in
`cylc-suite.db`.

Close #1646.

PR #1735, opened 2015-10-22
rose suite-hook: --retrieve-job-logs

matthewrmshin:
Recent version of cylc, e.g. 6.7.1, has the ability to efficiently
retrieve remote job logs. The equivalent functionality in
`rose suite-hook` should now be considered unnecessary. This change
modifies the behaviour so that the functionality will not be carried out
unless the new `--retrieve-job-logs` option is specified.

See also #1731. @arjclark @kaday please review.

### DIFF
--- a/bin/rose-suite-hook
+++ b/bin/rose-suite-hook
@@ -27,8 +27,7 @@
 #
 # DESCRIPTION
 #     Provide a common event hook for Cylc suites and tasks.
-#     * (Task event only) Pull remote task logs back to server host.
-#     * Generate suite log view.
+#     * (Task event only) Retrieve remote task logs back to server host.
 #     * Email user if --mail specified.
 #     * Shutdown suite if --shutdown specified.
 #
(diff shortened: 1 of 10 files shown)


arjclark:
Looks ok to me. Needs to be a warning in big bold letters somewhere in the changes file that there is now an implicit requirement for cylc 6.7+ if you want your logs retrieving in a suite for view via rose bush etc. May also be appropriate to put something in doc/rose-install.html under the cylc entry.

arjclark:
N.B. for the above I appreciate the existing behaviour can be replicated by adding the `--retrieve-job-logs` option to `rose suite-hook` calls but this is a change in default behaviour so one cannot simply upgrade rose without upgrading cylc and expect existing suites to work as before without changes.

As I said, a note at the start of the release notes/changes file (as we've done in past for cylc version dependent changes) will be sufficient.

Over to @kaday

kaday:
Looks okay to me, passes tests in my environment.

PR #1738, opened 2015-10-26
rosie create: meta_suite_mode should be False

### LINKED ISSUE #1737: rosie go creates suite u-ROSIE
We've just installed 2015.10.1.

Creating a suite from the GUI gives a suite named u-ROSIE (unfortunately in the MOSRS repo).

Creating one from the command line with rosie create gives the expected naming.

2015.04.1 (our previously installed version) works properly in both cases.

I've run out of time to check this properly now, but it seems the change is due to meta_suite_mode now being default True in function create in vc.py.

kaday:
Fix #1737

### DIFF
--- a/lib/python/rosie/vc.py
+++ b/lib/python/rosie/vc.py
@@ -243,7 +243,7 @@ class RosieVCClient(object):
         return id_
 
     def create(self, info_config, from_id=None, prefix=None,
-               meta_suite_mode=True):
+               meta_suite_mode=False):
         """Create a suite.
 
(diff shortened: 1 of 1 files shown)


kaday:
@matthewrmshin please review

matthewrmshin:
I can confirm that this fixes the issue in my environment.

@arjclark please review 2.

arjclark:
Looks good to me. No problems from the test battery in my environment and @matthewrmshin has reported the change as fixing the problem.

PR #1740, opened 2015-10-26
meta types: Add python_boolean meta type

arjclark:
Adds a python boolean metadata type.

Motivated by:
 * frequent using of jinja2 switches in suite.conf files and desire for suitable metadata type for that
 * desire for python boolean as part of #1696 

@kaday - please review 1
@benfitzpatrick - please review 2

### DIFF
--- a/lib/python/rose/meta_type.py
+++ b/lib/python/rose/meta_type.py
@@ -101,6 +101,18 @@ class IntegerMetaType(MetaType):
         return [True, None]
 
 
+class PythonBooleanMetaType(MetaType):
+
+    KEY = "python_boolean"
+    WARNING = "Not a valid Python boolean format (True/False): {0}"
+
+    def is_valid(self, value):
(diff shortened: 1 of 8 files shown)


lib/python/rose/meta_type.py
```diff
@@ -101,6 +101,18 @@ def is_valid(self, value):
         return [True, None]
 
 
+class PythonBooleanMetaType(MetaType):
+
```
kaday: Does this need a docstring?

arjclark: Well, the rest of the file doesn't do this and the class name is pretty much what'd go in the docstring - "metadata type for Python Boolean" - so unless you have particularly strong feelings that it should I'd say no.

kaday:
Looks okay to me, tests pass in my environment. Over to @benfitzpatrick

PR #1747, opened 2015-11-04
t: improve handling of *.pyc files

matthewrmshin:
This change brings rose test battery in line with cylc. If `lib/python/`
is writable, the test battery command will remove and rebuild `*.pyc`
files in it before running the tests. Consequently, it is no longer
necessary to run the old pyc test.


arjclark:
Looks good to me. Change is well understood from the equivalent cylc test battery functionality (see e.g. https://github.com/cylc/cylc/pull/1648

arjclark:
Over to @kaday

kaday:
Looks okay, tests pass in my environment.

PR #1748, opened 2015-11-04
t: improve robustness

matthewrmshin:
Retry delays in cylc `global.rc` can cause some tests to fail. This
change allows the tests to run in purer form without cylc `global.rc`
settings.
This change also ensures that the new rose_bunch test is able to tidy up
the suite. The NAME variable (suite name) was modified before it is
supplied to the `rose suite-clean` command.


arjclark:
Looks good to me. All passing in my environment.

arjclark:
Over to @kaday

kaday:
Looks okay, tests pass in my environment.

PR #1749, opened 2015-11-04
t: handle new version of `pgrep`

matthewrmshin:
This change will allow the test to pass on Ubuntu 14.04 boxes.

@arjclark please review. (I don't think a 2nd review is necessary for this.)


arjclark:
Looks ok to me. Test battery continuing to work in my RHEL environment. Will take your word that it works on Ubuntu 14.04 as I have no means of testing that.

Issue #1642, opened 2015-06-29
rose bush: custom title/logo/icon


matthewrmshin:
We should allow custom title, logo and/or icon to make it easier for users to distinguish between different Rose Bush instances.

PR #1750, opened 2015-11-05
rose bush and rosie disco: custom title and host

matthewrmshin:
Allow title and host to be customised for Rose Bush and Rosie Disco.

Address #1642.


matthewrmshin:

etc/rose.conf.example
```diff
@@ -106,15 +106,23 @@
 #  user-tool=ldap
 [rosa-svn]
 
-# Configuration for Rose Bush server. (Examples are default values.)
+# Configuration for Rose Bush server. (Number examples are default values.)
```
arjclark: I think it should be "Numeric" rather than "Number"

etc/rose.conf.example
```diff
@@ -106,15 +106,23 @@
...
 #  cycles-per-page=100
+## An alternate host name. (Default is the server's host name.)
+#  host=NAME
+## E.g.:
+#  host=my-wonderful-host
 ## Job list view: default number of jobs per page
 #  jobs-per-page=15
 ## Job list view: maximum number of jobs per page
 #  jobs-per-page-max=300
+## An alternate service title (Default="Rosie Suites Discovery")
```
arjclark: Given that this is rose-bush "rosie suites discovery" is appropriate as a default - its not about suite discovery.

lib/python/rose/bush.py
```diff
@@ -57,25 +58,33 @@ class RoseBushService(object):
...
+    def index(self, form=None):
         """Display a page to input user ID and suite ID."""
         # TODO: some way to allow autocomplete of user field?
+        data = {
+            "title": self.title,
+            "host": self.host_name,
+            "rose_version": self.rose_version,
+            "script": cherrypy.request.script_name,
+        }
+        if form == "json":
```
arjclark: "format" feels like it'd be a more appropriate variable name here as its the format in which to render the data.

matthewrmshin: `format` is a python built-in name. We didn't think about this when we were doing `rosie.ws`. When we wrote Rose Bush, we have chosen `form=` for all methods to avoid clashing with the built-in.

lib/python/rose/bush.py
```diff
@@ -111,6 +121,7 @@ def broadcast_events(self, user, suite, form=None):
         """List broadcasts history of a running or completed suite."""
         user_suite_dir = self._get_user_suite_dir(user, suite)
         data = {
+            "title": self.title,
```
arjclark: It feels like these repeated data dictionaries (here and in subsequent methods) feel like they should be able to be cut down to some sort of object with overrides for the settings that can change i.e. we're always using host, and rose version. Probably not an issue for here but I think the addition of title in each case highlights ongoing redundancy.

matthewrmshin: Agreed. That's the next re-factor, that I am also working on!

matthewrmshin:
Typos in `rose.conf.example` fixed.

t/lib/bash/test_header
```diff
@@ -80,11 +80,40 @@
...
+#     TEST_NUMBER
+#         Number of tests already run.
+#     TEST_SOURCE_DIR
+#         The directory containing of the current test file.
 #-------------------------------------------------------------------------------
 set -eu
 
 SIGNALS="EXIT INT"
-function FINALLY() {
+FINALLY() {
```
arjclark: was referring to these as functions just unnecessary before then?

t/rose-bush/00-basic/rose-suite.info
```diff
@@ -0,0 +1,3 @@
+owner=robert.fitzroy
```
arjclark: Nice :grin:

t/rose-bush/01-title
```diff
@@ -0,0 +1 @@
+00-basic
```
arjclark: Is this just providing the name for the title bar in rose bush?

t/rosie-disco/00-basic.t
```diff
@@ -46,30 +42,17 @@ __ROSE_CONF__
...
-done
-if port_is_busy $PORT; then
-    pass "$TEST_KEY"
-else
-    fail "$TEST_KEY"
-    kill "${ROSIE_DISCO_PID}" 2>'/dev/null'
-    wait 2>'/dev/null'
-    rm -f ~/.metomi/rosie-disco-0.0.0.0-${PORT}* 2>'/dev/null'
+rose_ws_init 'rosie' 'disco'
+if [[ -z "${TEST_ROSE_WS_PORT}" ]]; then
```
arjclark: Should we have a `rose_ws_kill` here too or will the lack of setting that variable guarantee that the server never started?

lib/python/rose/bush.py
arjclark: Fair enough. I can't think of any other appropriate synonyms that would apply here that aren't python built-ins either.

lib/python/rose/bush.py
arjclark: Sounds good.

t/lib/bash/test_header
matthewrmshin: Yes. There are subtle differences between the 2 syntaxes, but it is not not necessary for any of our usages.

t/rose-bush/01-title
matthewrmshin: This is symbolic link.

t/rosie-disco/00-basic.t
matthewrmshin: `rose_ws_init` does it on failure.

t/rose-bush/01-title
arjclark: Oh, that makes much more sense.

arjclark:
All feedback has been addressed now or in process of being addressed in future changes. Looks good to me otherwise. Test battery passing in my environment. Over to @kaday

kaday:
Looks okay to me. Test pass in my environment.

PR #1751, opened 2015-11-09
Add comparison databse to rose_ana output

stevewardle:
In the UM rose-stem suite we rely on a script to update our test files - it parses the output from the rose_ana tasks in the suite as well as looking at the suite database etc. to build up an understanding of what files need updating.  However this would be greatly simplified if rose_ana provided a database with some key comparison information instead.  This request adds that capability to rose_ana.


stevewardle:
@scwhitehouse could you take a look at this and comment please

scwhitehouse:
This is missing tests for the new functionality, and also should probably be mentioned in the documentation.

lib/python/rose/apps/rose_ana.py
```diff
@@ -48,6 +50,85 @@
...
 
+class KGODatabase(object):
+    """
+    KGO Database object, stores comparison information for rose_ana apps.
+
+    """
+    def __init__(self):
+        """Initialise the object"""
+        self.file_name = os.path.join(
+            os.getenv("ROSE_SUITE_DIR"), "log", "rose-ana-comparisons.db")
```
scwhitehouse: Is ROSE_SUITE_DIR always set?

lib/python/rose/apps/rose_ana.py
```diff
@@ -48,6 +50,85 @@
...
+                """
+                CREATE TABLE tasks (
+                task_name TEXT,
+                completed INT,
+                PRIMARY KEY(task_name))
+                """)
+            conn.commit()
+        self.unlock()
+
+    def wait_for_lock(self):
```
scwhitehouse: Am I right in thinking that this locking is purely for database creation and not just for each task updating it?

lib/python/rose/apps/rose_ana.py
```diff
@@ -72,8 +153,30 @@ def run(self, app_runner, conf_tree, opts, args, uuid, work_files):
...
+
+        # Update the database to indicate that the task succeeded
+        kgo_db.enter_task(rose_ana_task_name, 0)
+
+        # Update the database
+        for task in tasks:
+            # The primary key in the database is composed from both the
+            # rose_ana app name and the task index (to make it unique)
+            app_task = "{0} ({1})".format(rose_ana_task_name, task.name)
+            # Include an indication of what extrac/comparison was performed
```
scwhitehouse: extrac->extract

lib/python/rose/apps/rose_ana.py
```diff
@@ -48,6 +50,85 @@
...
+    def get_conn(self):
+        """Return a connection to the database if it exists"""
+        if self.conn is None:
+            self.conn = sqlite3.connect(self.file_name, timeout=60.0)
+        return self.conn
+
+    def create(self):
+        """If the databaste doesn't exist, create it"""
+        self.wait_for_lock()
+        if not os.path.exists(self.file_name):
```
scwhitehouse: Is this logic the right way around? Surely if the database already exists there's no need to lock and immediately unlock it?

lib/python/rose/apps/rose_ana.py
```diff
@@ -48,6 +50,85 @@
...
+            self, app_task, kgo_file, suite_file, status, comparison):
+        """Insert a new comparison entry to the database"""
+        conn = self.get_conn()
+        conn.execute(
+            "INSERT OR REPLACE INTO comparisons VALUES (?, ?, ?, ?, ?)",
+            [app_task, kgo_file, suite_file, status, comparison])
+        conn.commit()
+
+    def enter_task(self, app_task, status):
+        conn = self.get_conn()
```
scwhitehouse: Missing docstring

lib/python/rose/apps/rose_ana.py
stevewardle: As far as I know it should always be set, yes

lib/python/rose/apps/rose_ana.py
stevewardle: Yes, that's right - the locking is for creation as multiple tasks trying to create the database could be bad.  The updates to the database are protected by sqlite3's own locking mechanism (which has a timeout I've set to 60 seconds)

lib/python/rose/apps/rose_ana.py
stevewardle: Fixed, thanks

lib/python/rose/apps/rose_ana.py
stevewardle: You have to acquire the lock before checking if the file exists, otherwise multiple tasks could make it into the "if file doesn't exist" branch at the same time.  You'd then need an additional check just after the lock which tells the non-locking threads to bail-out... Perhaps that's better but I felt like this way around was slightly nicer?

lib/python/rose/apps/rose_ana.py
stevewardle: Thanks - fixed

stevewardle:
Having discussed this further I reworked the locking logic slightly - I think this is better

stevewardle:
I've now added a series of tests to the existing rose_ana tests in the test battery - it became clear as I did this that the method of locking just didn't work reliably, but I worked out I can use sqlite's own functionality to ensure the first task will create the database without clashing with other tasks

scwhitehouse:
This looks ok to me, I've given it a quick test myself and it seems to work. Assigning to @matthewrmshin to take a look.

matthewrmshin:
The PEP8 test fails with:

```
lib/python/rose/apps/rose_ana.py:114:22: W291 trailing whitespace
lib/python/rose/apps/rose_ana.py:115:1: W293 blank line contains whitespace
```

PR #1752, opened 2015-11-11
fcm_make built-in: fix fast location perm mode

matthewrmshin:
When using a fast location, the logic creates a temporary directory on
the fast location. This is rsync to the final destination on exit of the
`fcm make` command. The old did take into account permission mode of the
temporary directory back to the final destination, but the logic was
insufficient, as the directory permission mode is overridden by
`rsync -a`. The new logic fixes this by inquiring the permission mode of
the normally created directory at the final destination, then run the
`rsync -a` command, before setting the permission mode of the directory
to the the value of that of the normally created directory.


matthewrmshin:

arjclark:
Looks ok to me. Test battery passing in my environment.

arjclark:
Over to @kaday

kaday:
Looks okay and test battery passes in my environment.

PR #1754, opened 2015-11-12
t: check syntax of HTML files in `doc/`

matthewrmshin:
Also moved exist PEP8 test file to under `t/syntax/` so it can live with
the new test file for HTML syntax.

@arjclark please review. (I don't think this needs a 2nd.)


arjclark:
Looks good to me. Can't see any problems with the changes. Test battery passing in my environment.

PR #1755, opened 2015-11-13
t: rose_arch: archive already compressed files

matthewrmshin:
New test to prove that we are OK archiving already compressed files.


matthewrmshin:
@arjclark please review. (I don't think we need a 2nd for this.)

arjclark:
Looks ok to me. Test working in my environment. Doesn't change any behaviour and is relatively minor addition, so agree doesnt need review 2'ing.

PR #1759, opened 2015-11-19
rose bush: job entry: don't truncate entry name

matthewrmshin:
Don't bother truncating task name. A long name will automatically move itself down any way.


matthewrmshin:
This should fix the issue raised by @stevewardle yesterday.

@arjclark please review. (I don't think this needs a 2nd, but feel free to assign one if you think that is necessary.)

arjclark:
Looks good to me. @stevewardle is happy this fixes the issue. Tested against one of the UM rose-stem suites.

PR #1761, opened 2015-11-23
Improve test-battery doc

matthewrmshin:
@arjclark please review. (Only a doc change for a developer command, so 1 reviewer is probably enough.)


arjclark:
Looks ok to me.

PR #1762, opened 2015-11-25
rose bush: Fix sqlite query for filtering jobs by status.

arjclark:
Filtering jobs by status in rose bush (2015.11.1) is currently broken. This fixes the query so it will now work.

Can be tested by viewing suite results mot-aa834 in my area and filtering jobs using each checkbutton.

@kaday please review 2


matthewrmshin:
Looks good.

kaday:
Tested and works, looks okay.

PR #1767, opened 2015-12-03
rose bush: job entry: display job.out/err any way

matthewrmshin:
Display links to `job.out` and `job.err` even if not registered in the cylc suite run database (but display in grey with a trailing `?`) for completed task jobs.


matthewrmshin:
@arjclark @kaday please review (and/or comment). I think it is better to have a link than not, but may generate dislikes from some users.

arjclark:
Based on the feedback I've been dealing from users for the past week this sounds like a good idea to me. Will review tomorrow.

arjclark:
Looks good to me. Have tested manually to see what it looks like in rose bush and think its ok.

I wonder though, as a possible way to make it better, if we couldn't have something like the logic which we use to grey out and de-activate links to empty files for putting the missing link in.

i.e. - pseudocode:
```
if job.out not in job_logs_table:
    if size_of_file(expected/path/to/job.out) > 0:
        insert working (non-greyed but blue) link with question mark on end
```
This might be the best of all worlds as the file should be present but hasn't been recorded for whatever reason but will only put in the link if the file is actually there rather than speculatively so.

arjclark:
@kaday - have a look see what you think of the interface change. I've simulated this with my "matt_test" suite

matthewrmshin:
Now look at the file system where possible. (Will this create a load?)

kaday:
Looks good, test battery passes in my environment.

PR #1768, opened 2015-12-04
rose_arch: optional target

matthewrmshin:
New syntax to indicate a target may have no source.
New status for target with no source.
doc: optional target and more on rose_arch output.

See also #1707 and #1708.

### DIFF
--- a/doc/rose-rug-task-run.html
+++ b/doc/rose-rug-task-run.html
@@ -531,6 +531,11 @@ opt.jobs=8
     <kbd>[arch:TARGET]</kbd> section would override those in the global
     <kbd>[arch]</kbd> section for the given <var>TARGET</var>.</p>
 
+    <p>A target is considered compulsory, i.e. it must have at least one source,
+    unless it is specified with the syntax <kbd>[arch:(TARGET)]</kbd>. In which
+    case, <var>TARGET</var> is considered optional. The application will skip an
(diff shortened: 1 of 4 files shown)


arjclark:
Two possible changes I'd like based on our discussion.

My concern was that if a target was considered optional then it feels that the sources within should be considered optional too. So rather than:
```
[target:(foo)]
sources=(bar)
```
being the way to write this, having [target:(foo)] should imply sources=(bar), so should just be able to write:
```
[target:(foo)]
sources=bar
```
I *think* that this is achievable with minor work by inspecting the `is_target_compulsory` option when setting the `is_source_compulsory` option.

Alternatively, if this would be a major refactor then a note in the documentation (an N.B. directly after your example) saying that for optional targets all sources must be made optional otherwise the task will fail on missing sources would be sufficient.

malcolmbrooks:
Thanks Matt, the syntax as it stands would work for me. Having either optional sources or targets as separate things might be useful in some sitations.

Most archiving, in practice, specifies single sources to a single target. In this case, an optional target is of most use.

matthewrmshin:
Optional target now implies optional sources. Looks like the simple solution works.

arjclark:
Looks good to me. Test battery passing in my environment.

Over to you @kaday

kaday:
Looks okay to me and passes the tests in my environment.

PR #1769, opened 2015-12-04
t: rosie-create 00: fix random failure

matthewrmshin:
Caused by modifying a file to be comitted with a random number that may
be exactly the same as the original file - which caused the test to fail
with a 1 in 10 chance.

@arjclark please review. (I don't think we need a 2nd.)


arjclark:
Looks good to me. Tested as working by repeatedly running the test.

Was pretty much every first time run of the test battery for me :grin: - so much for randomly random numbers...

Review 2 not necessary.

PR #1773, opened 2015-12-09
doc: rose-api: updated

matthewrmshin:
This change updates a few one-liners in this doc.

@arjclark please review. (I think a review 1 is enough.)


arjclark:
Looks ok. Not sure how informative they are but seems fine to me.

PR #1776, opened 2015-12-10
rose_arch: improve exception message on no source

matthewrmshin:
When a source glob fails to return a valid file, return the full path of
the glob including the source prefix in the exception message.

Problem reported by @malcolmbrooks.


arjclark:
Looks good to me - ultimately a relatively minor change. Test battery passing in my environment.

Over to @kaday

kaday:
Looks okay to me and passes test battery in my environment.

PR #1779, opened 2015-12-15
rose bush: cycles: fix job fails count

matthewrmshin:
Presence of statement arguments for query causes the query to do nothing.


arjclark:
Looks good to me. Manually tested as working with my "FAIL" suite. No problems from the test battery in my environment. Over to you @kaday

kaday:
Looks okay to me and test battery passes in my environment.

PR #1783, opened 2015-12-21
Rose bunch: evaluate environment varibles in settings and args

arjclark:
This allows rose bunch to evaluate environment variables passed through to its settings and args sections - particularly useful for how some of our suites are written. Adds in tests for this behaviour too.

@kaday please review 2


kaday:
Looks good to me and passes the test-battery, review2 done.

matthewrmshin:
Should use `rose.env.env_var_process` for consistency with other logic in Rose. The main problem is that `os.path.expandvars`:
* Does not fail on undefined environment variables, and leaves the string as-is - which can cause surprising behaviour.
* Does not handle backslash escape.

arjclark:
@matthewrmshin - done.

PR #1785, opened 2015-12-22
rose bush: cycles/jobs: display suite host:port

matthewrmshin:
Take advantage of cylc/cylc#1705

### DIFF
--- a/lib/html/template/rose-bush/cycles.html
+++ b/lib/html/template/rose-bush/cycles.html
@@ -57,36 +57,45 @@
 
 {% include "pager.html" -%}
 
-{% for entry in entries -%}
-<div class="row-fluid entry">
-<div class="span2">
-<a href="{{script}}/jobs/{{user}}/{{suite}}?cycles={{entry.cycle|replace('+', '%2B')}}" class="cycle" title="{{entry.cycle}}">{{entry.cycle}}</a>
-</div>
-{% for state, icon, label, title in [("active", "play", "info", "active tasks"), ("success", "ok", "success", "succeeded tasks"), ("fail", "warning-sign", "important", "failed tasks")] -%}
-{% set n_state = entry.n_states[state] -%}
-<div class="span2">
-{% if n_state -%}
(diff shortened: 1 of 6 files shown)


matthewrmshin:
@arjclark @kaday please review. New test requires cylc/cylc#1705 or it will skip.

arjclark:
I'm inclined to wait until cylc/cylc#1705 is finalised and merged before pressing go on this one in case anything changes once @hjoliver has had chance to review it properly.

arjclark:
As a first pass, looks ok and test battery works in my environment.

I'm not so sure how helpful the line:

Suite has failed tasks \, is running \, last activity 25 minutes ago

is in one of my older suites as, yes a task failed some time ago (last month), but it was subsequently suicided out and has no bearing on the current state of the suite. This has the potential to end up worrying users needlessly...

matthewrmshin:
> Suite has failed tasks , is running , last activity 25 minutes ago

This line has always been there, but was placed in the navigation bar at the top right.

I have moved it down so we can add the `host:port` information in as well. (Otherwise, the navigation bar will look too busy.)

arjclark:
> This line has always been there, but was placed in the navigation bar at the top right.

Got to say, even with those directions I've just struggled to find the equivalent in the old view (2015.12.0) - admittedly obvious once I found it though that that was what has moved down. It is definitely more prominent now for better or worse. IMHO I don't like that being there where it's now the first thing you see but I'm happy for the rest of the team to decide what they think and overrule me on this one - @metomi/metos please take a look and see what you think. (N.B. I have no objection to the "last activity" part of the entry being there, so if this were a case of the icons being kept where they were and the leading line being something like "running on <hostname>, last activity <offset> ago" I wouldn't have any objection).

arjclark:
(Sorry, above should've been @metomi/owners)

matthewrmshin:
Branch re-based.

kaday:
Passes tests in my environment.
Looks okay to me.
I am happy with the change.

PR #1787, opened 2015-12-29
rose bush: configurable logo

### LINKED ISSUE #1642: rose bush: custom title/logo/icon

matthewrmshin:
Close #1642.

### DIFF
--- a/lib/html/template/rose-bush/index.html
+++ b/lib/html/template/rose-bush/index.html
@@ -12,6 +12,9 @@ rel="stylesheet" media="screen" />
 <body>
 <div class="container-fluid">
 <div class="hero-unit">
+{% if logo -%}
+<img {{logo}} />
+{% endif -%}
 <h1>{{title}} <small>@ {{host}}</small></h1>
 <p>Browse User Suite-logs on HTTP</p>
 <p class="pull-right"><small>Rose {{rose_version}}</small></p>
--- a/lib/html/template/rose-bush/suite-base.html
+++ b/lib/html/template/rose-bush/suite-base.html
@@ -17,7 +17,12 @@ rel="stylesheet" media="screen" />
 <body>
 <div class="navbar navbar-static-top">
 <div class="navbar-inner">
-<a class="brand" href="{{script}}/">{{title}} <small>@ {{host}}</small></a>
+<a class="brand" href="{{script}}/">
+{% if logo -%}
+<img {{logo}} />
+{% endif -%}
+{{title}} <small>@ {{host}}</small>
+</a>
 <ul class="nav">
 <li><a href="{{script}}/suites/{{user}}">{{user}}</a></li>
 <li><a><strong>{{suite}}</strong></a></li>
--- a/lib/html/template/rose-bush/suites.html
+++ b/lib/html/template/rose-bush/suites.html
@@ -12,7 +12,12 @@ rel="stylesheet" media="screen" />
 <body>
 <div class="navbar navbar-static-top">
 <div class="navbar-inner">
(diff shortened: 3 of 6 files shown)


matthewrmshin:

lib/html/template/rose-bush/index.html
```diff
@@ -12,6 +12,9 @@
 <body>
 <div class="container-fluid">
 <div class="hero-unit">
+{% if logo -%}
+<img {{logo}} />
```
arjclark: Can you add in some alt text

lib/html/template/rose-bush/suite-base.html
```diff
@@ -17,7 +17,12 @@
 <body>
 <div class="navbar navbar-static-top">
 <div class="navbar-inner">
-<a class="brand" href="{{script}}/">{{title}} <small>@ {{host}}</small></a>
+<a class="brand" href="{{script}}/">
+{% if logo -%}
+<img {{logo}} />
```
arjclark: can you add some alt text

lib/html/template/rose-bush/suites.html
```diff
@@ -12,7 +12,12 @@
 <body>
 <div class="navbar navbar-static-top">
 <div class="navbar-inner">
-<a class="brand" href="{{script}}/">{{title}} <small>@ {{host}}</small></a>
+<a class="brand" href="{{script}}/">
+{% if logo -%}
+<img {{logo}} />
```
arjclark: as above

lib/python/rose/bush.py
```diff
@@ -452,6 +459,7 @@ def view(self, user, suite, path, path_in_tar=None, mode=None):
             rose_version=self.rose_version,
             script=cherrypy.request.script_name,
             time=strftime("%Y-%m-%dT%H:%M:%S+0000", gmtime()),
+            logo=self.logo,
```
arjclark: I think this repeated setting of self.logo is once again highlighting the levels of redundancy in the setting of these data dicts.

lib/html/template/rose-bush/index.html
arjclark: I note in your example below that you configure the alt text as:
```
 [rose-bush]
logo=src="rose-favicon.png" alt="Rose Logo"
```
Not really sure that this should be something that should be configured in the logo config setting.

matthewrmshin: See example in `etc/rose.conf.example` above. The variable `logo` should contain all the attributes required by the `img` tag.

lib/python/rose/bush.py
matthewrmshin: Acknowledged. I am still running a a branch to give this module a good lint.

lib/html/template/rose-bush/index.html
matthewrmshin: I did consider an alternate approach, which will break the attributes into key-value pairs, but it would require a bit more logic. The implemented approach requires the least logic, but is just as effective.

arjclark: I think I'd have preferred the alternative approach as it's less prone to user error. I'll leave it to @kaday to see what she thinks about this though.

arjclark:
Test battery passing in my environment. Code generally looks ok, though I'd like a second opinion on having to configure all html tag elements in the single rose config setting (may just be me that doesn't like this).

Over to @kaday

kaday:
Looks okay, passes test-battery, got a png to work.
In response to @arjclark comment, Matt does not think consider this a major stopper and I am indifferent. No argument seemed more persuasive, so letting it go.

PR #1791, opened 2016-01-07
rose bush: job entry: improve fuzzy time + layout

### LINKED ISSUE #1789: rose bush: jobs list: retain time delta option when paging
This requires passing the option to the server.

matthewrmshin:
Fuzzy time toggle is now preserved when paging.
Location and times are now in their own columns.
A submit number badge is now:
* a link to select all jobs of a given point/task.
* displaying the number of jobs in a given point/task.
* displaying the task status by colour of the point/task.

Close #1789.

### DIFF
--- a/lib/html/template/rose-bush/job-entry.html
+++ b/lib/html/template/rose-bush/job-entry.html
@@ -3,14 +3,14 @@
     <!-- entry: cycle, name, submit_num, files -->
     <!-- entry: status -->
     {% if entry.status and "fail" in entry.status -%}
-    <span class="label label-important">
-      <i class="icon-warning-sign icon-white" title="{{entry.status}}"></i>
-    </span>
-    {% if entry.status != "fail" and entry.status != "fail(ERR)" -%}
       <span class="label label-important">
-        {{entry.status[5:-1]}}
+        <i class="icon-warning-sign icon-white" title="{{entry.status}}"></i>
       </span>
-    {% endif -%}
+      {% if entry.status != "fail" and entry.status != "fail(ERR)" -%}
+        <span class="label label-important">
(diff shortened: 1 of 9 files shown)


matthewrmshin:
@kaday @arjclark please review.

kaday:
The code looks okay.
Test battery fails tests 3 and sometimes 5 of `t/rose-bush/06-host-port.t` but it is nothing to do with this change and @matthewrmshin is fixing it in a new pull request.
Only thing, which could be changed is the `Submitted` could be split on `T`, but this is not a show stopper for me.
Over to @arjclark

lib/html/template/rose-bush/job-entry.html
```diff
@@ -1,139 +1,178 @@
...
-{% endif -%}
-</div>
+  <div class="span4">
+    <!-- entry: cycle, name, submit_num, files -->
+    <!-- entry: status -->
+    {% if entry.status and "fail" in entry.status -%}
+    <span class="label label-important">
+      <i class="icon-warning-sign icon-white" title="{{entry.status}}"></i>
+    </span>
+    {% if entry.status != "fail" and entry.status != "fail(ERR)" -%}
```
arjclark: This if needs indent nesting as its associated endif looks in error against the elif.

arjclark:
One comment on indenting correction otherwise looks good and is testing as working via manual tests and the test battery.

lib/html/template/rose-bush/job-entry.html
matthewrmshin: Fixed.

arjclark:
Change looks good to me once I've unpicked the way github is rendering it. Test battery still passing and no problems found manually testing.

PR #1793, opened 2016-01-11
t: rose-bush 06: improve robustness

matthewrmshin:
Detect for non-zero size in port file, instead of just existence.


matthewrmshin:
@kaday please review. (I think a single review is enough for this one.)

kaday:
Looks okay to me and passes test battery.

PR #1800, opened 2016-01-13
t: rose-bush 06: improve robustness

matthewrmshin:
@kaday please review. (One review is probably enough for this.)


kaday:
It seems more robust in my environment.
Ran the test multiple times.

PR #1804, opened 2016-01-19
Add extra logic to retry rose_ana KGO database operations after failures

stevewardle:
Due to some intermittent failures with the KGO database in rose_ana we should probably add some extra protection around the SQL operations (the errors are platform/file-system related rather than problems with the code - but it's not going to hurt having a little extra checking)

### DIFF
--- a/lib/python/rose/apps/rose_ana.py
+++ b/lib/python/rose/apps/rose_ana.py
@@ -27,6 +27,7 @@ import os
 import re
 import sys
 import sqlite3
+import time
 
 # Rose modules
 import rose.config
@@ -76,8 +77,7 @@ class KGODatabase(object):
         "If the databaste doesn't exist, create it."
         conn = self.get_conn()
         # This table stores the individual comparisons
-        conn.execute(
-            """
+        sql_statement = """
             CREATE TABLE IF NOT EXISTS comparisons (
             app_task TEXT,
             kgo_file TEXT,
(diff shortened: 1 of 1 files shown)


stevewardle:
I'll copy in @scwhitehouse as he'll want to be aware of progress

matthewrmshin:
@scwhitehouse please review 1. I'll do the sanity check.

lib/python/rose/apps/rose_ana.py
```diff
@@ -85,32 +85,53 @@ def create(self):
             status TEXT,
             comparison TEXT,
             PRIMARY KEY(app_task))
-            """)
-        # And this one stores the completion status of the task
-        conn.execute(
             """
+        self.execute_sql_retry(conn, (sql_statement,))
+        # And this one stores the completion status of the task
+        sql_statement = """
```
scwhitehouse: I'm not overly enthusiastic about the reuse of the sql_statement variable name

lib/python/rose/apps/rose_ana.py
```diff
@@ -76,8 +77,7 @@ def create(self):
         "If the databaste doesn't exist, create it."
         conn = self.get_conn()
         # This table stores the individual comparisons
-        conn.execute(
-            """
+        sql_statement = """
             CREATE TABLE IF NOT EXISTS comparisons (
             app_task TEXT,
```
scwhitehouse: This is a little like magic to me, comment what this SQL command is actually trying to do?

lib/python/rose/apps/rose_ana.py
```diff
@@ -85,32 +85,53 @@ def create(self):
...
+        concurrency issues
+
+        """
+        for retry in range(retries):
+            try:
+                conn.execute(*sql_args)
+                return
+            except:
+                time.sleep(timeout)
+        raise
```
scwhitehouse: This is probably me being ignorant of a Python thing, but don't you need to specify what you're raising?

lib/python/rose/apps/rose_ana.py
stevewardle: Hmm... I can change the name of the second one if you're insisting but I don't see that it's a huge problem - the statement execution can't fail and the variable is used in both cases immediately on the next line.  (It's similar to the fairly common use of a variable called something like "msg" for error-strings to be passed to an exception)

lib/python/rose/apps/rose_ana.py
stevewardle: Okay - I'll add a comment before the various SQL statements to clarify the bits that aren't spelled out in the command

lib/python/rose/apps/rose_ana.py
stevewardle: A raise on its own like this essentially means "raise whatever the last unhandled exception was"... so in the above case the code will repeatedly retry if an exception occurs, but if it exceeds the number of retries it will hit the raise and whatever the final error was will be re-raised and propagate up the stack (in this case there's nothing to catch it outside so it will bring down the whole app task)

scwhitehouse:
Looks ok, over to Matt.

PR #1817, opened 2016-01-28
rose task-env: improve --time-delta doc

matthewrmshin:
Improve doc for future duration.

@kaday please review.


kaday:
Looks good to me.
Passes the test-battery as expected in my my environment.

PR #1819, opened 2016-01-29
Reporter timestamp

### LINKED ISSUE #1782: rose app output: display time stamps in verbose modes
For debugging and problem identification purposes it would be useful to have timings printed in the logs from the various stages in the app output.

arjclark:
Closes  #1782

@matthewrmshin - please review 1. We can assign review 2 once you're happy with the underlying change being made to rose reporter.

### DIFF
--- a/lib/python/rose/reporter.py
+++ b/lib/python/rose/reporter.py
@@ -85,11 +85,9 @@ class Reporter(object):
         msg_lines = []
 
         if verbosity >= self.VV:
-            stamp = time.strftime("%Y-%m-%dT%H:%M:%SZ : ")
+            stamp = time.strftime("%Y-%m-%dT%H:%M:%S ")
         else:
             stamp = ""
-
-        insert_newline = False
 
         if prefix:
             for line in msg.splitlines():
@@ -100,12 +98,13 @@ class Reporter(object):
                 msg_lines.append(msg_line)
         else:
             msg_line = stamp
+            should_insert_newline = False
             if clip is not None:
                 if msg.endswith("\n") and len(msg) > clip:
-                    insert_newline = True
+                    should_insert_newline = True
                     msg = msg[:-1]
                 msg_line = msg_line + msg[:clip]
-            if insert_newline:
+            if should_insert_newline:
                 msg_line = msg_line + "\n"
             else:
                 msg_line = msg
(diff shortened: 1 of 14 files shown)


lib/python/rose/reporter.py
```diff
@@ -78,6 +79,40 @@ def __init__(self, verbosity=DEFAULT, contexts=None, raise_on_exc=False):
         self.event_handler = None
         self.raise_on_exc = raise_on_exc
 
+    def format_msg(self, msg, verbosity, prefix=None, clip=None):
+        """Format a message for reporting."""
+
+        msg_lines = []
+
+        if verbosity >= self.VV:
+            stamp = time.strftime("%Y-%m-%dT%H:%M:%SZ : ")
```
matthewrmshin: If you want Z here, you need to ensure `TZ=UTC` in the environment. Otherwise, the local time zone should be used. Can we use `isodatetime` here to generate the time stamp?

lib/python/rose/reporter.py
```diff
@@ -78,6 +79,40 @@ def __init__(self, verbosity=DEFAULT, contexts=None, raise_on_exc=False):
...
+        else:
+            stamp = ""
+
+        insert_newline = False
```
matthewrmshin: This is only used within the `else:` block. Can we call the variable `should_insert_newline` to make it more boolean-like?

matthewrmshin: (I know it was copy-and-paste from before, but since we are changing the logic...)

lib/python/rose/reporter.py
matthewrmshin: (I think a single space instead of space-colon-space is sufficient as a delimiter.)

arjclark:
@matthewrmshin - I believe these changes should address your feedback.

lib/python/rose/reporter.py
```diff
@@ -78,6 +79,39 @@ def __init__(self, verbosity=DEFAULT, contexts=None, raise_on_exc=False):
         self.event_handler = None
         self.raise_on_exc = raise_on_exc
 
+    def format_msg(self, msg, verbosity, prefix=None, clip=None):
+        """Format a message for reporting."""
+
+        msg_lines = []
+
+        if verbosity >= self.VV:
+            stamp = time.strftime("%Y-%m-%dT%H:%M:%S ")
```
matthewrmshin: I think you should add `%z` in for time zone?

arjclark:
@matthewrmshin - done.

arjclark:
@kaday please review 2

kaday:
Looks okay to me.
Passes test-battery in my environment.

PR #1820, opened 2016-02-01
rose bush: view: no fuzzy time on job entry breaks

matthewrmshin:
Ensure that the variable is defined before use.

### DIFF
--- a/lib/html/template/rose-bush/job-entry.html
+++ b/lib/html/template/rose-bush/job-entry.html
@@ -27,6 +27,8 @@
     <!-- entry: status -->
     {% if no_fuzzy_time -%}
       {% set no_fuzzy_time_str = "&no_fuzzy_time=" + no_fuzzy_time -%}
+    {% else -%}
+      {% set no_fuzzy_time_str = "" -%}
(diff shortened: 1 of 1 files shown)


lib/html/template/rose-bush/job-entry.html
```diff
@@ -27,6 +27,8 @@
     <!-- entry: status -->
     {% if no_fuzzy_time -%}
       {% set no_fuzzy_time_str = "&no_fuzzy_time=" + no_fuzzy_time -%}
```
arjclark: Is the red ampersand here a markup character issue?

matthewrmshin: No, the file has `*.html` extension, and so the viewer thinks that the ampersand is the beginning of an entity that is not terminated by a semicolon. In reality, it is in a Jinja2 quoted string.

arjclark:
Looks good to me. Over to @kaday

kaday:
Passes test-battery.
Code looks okay to me.
Have been assured that it fixes the problem as both Andy and Matt have seen the evidence.

PR #1828, opened 2016-02-11
Rose: doc datetime syntax update

kaday:
Updating the syntax.


kaday:
@benfitzpatrick please review, its just a syntax update so probably does not need a second review?

PR #1833, opened 2016-03-01
Allow tolerance to be an environment variable

scwhitehouse:
A user has spotted that whilst rose-ana accepts environment variables in filenames, it doesn't for the tolerance. This fixes that issue.


scwhitehouse:
@stevewardle you should be aware of this - your rose-ana rewrite needs to make sure that environment variables are expanded correctly for all variables.

@matthewrmshin when you get a moment would you mind reviewing? And nominating a second reviewer too?

matthewrmshin:
Looks OK to me. Want to add a test? @stevewardle is probably the best person to review this properly.

scwhitehouse:
Good idea, I've added some tests for this.

matthewrmshin:
Change OK. Test OK.

matthewrmshin:
@stevewardle please review 2.

stevewardle:
Looks fine to me, and I believe we have also had confirmation from the original user that it fixes their problem

PR #1837, opened 2016-03-02
doc: update usage of some suite commands

matthewrmshin:
@arjclark please review. (1 review enough?)

### DIFF
--- a/doc/rose-rug-files.html
+++ b/doc/rose-rug-files.html
@@ -444,7 +444,9 @@ rose config rose-suite-run hosts
     <div class="slide">
       <h3 class="alwayshidden">rose config --file</h3>
 
-      <p>rose config also works for the other Rose formats. For example:</p>
+      <p>rose config also works for other Rose configuration files (and any
+      configuration files that have similar formats as a Rose configuration
+      file). For example:</p>
       <pre class="shell">
 rose config --file=/path/to/rose-app.conf env
 </pre>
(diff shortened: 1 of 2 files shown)


doc/rose-rug-files.html
```diff
@@ -444,7 +444,9 @@ <h3 id="rose-config">rose config</h3>
     <div class="slide">
       <h3 class="alwayshidden">rose config --file</h3>
 
-      <p>rose config also works for the other Rose formats. For example:</p>
+      <p>rose config also works for other Rose configuration files (and any
```
arjclark: Could you pop this in ```<kbd>``` tags like the other commands (I know it wasn't originally but while you're in here we may as well change it)

doc/rose-rug-files.html
```diff
@@ -522,17 +524,27 @@ <h3 class="alwayshidden">Other Useful Commands</h3>
...
-
-        <li><kbd>rose suite-clean</kbd>: remove items created by <code>rose
-        suite-run</code> for the suite in <var>$PWD</var> and those on remote
-        job hosts.</li>
+        <li><kbd>rose suite-scan</kbd> or <kbd>cylc scan</kbd>: list your
+        running suites.</li>
+
+        <li><kbd>cylc gscan</kbd>: GTK version of <kbd>cylc scan</kbd>.</li>
+
+        <li><kbd>rose suite-gcontrol [--name=NAME]</kbd> or <kbd>cylc gui
```
arjclark: Not sure about telling people to run cylc gui like this as it needs the --host option to make it run on the correct server - if its not provided you (currently) get a "connection failed" dialog on opening of the cylc gui - or has one of your recent cylc changes altered this behaviour?

doc/rose-rug-files.html
```diff
@@ -522,17 +524,27 @@ <h3 class="alwayshidden">Other Useful Commands</h3>
...
+        NAME</kbd>: launch the <code>cylc gui</code> for a suite called
+        <var>NAME</var>. The rose wrapper uses the base name of <var>$PWD</var>
+        as <var>NAME</var> in the absence of the <kbd>--name=NAME</kbd>
+        option.</li>
+
+        <li><kbd>rose suite-shutdown [--name=NAME]</kbd> or <kbd>cylc stop
```
arjclark: Same comment as above re. needing --host= in the current version of cylc.

arjclark:
@matthewrmshin - couple of comments above. Primarily, do these additions of the cylc commands as written reflect a recent change in cylc?

doc/rose-rug-files.html
matthewrmshin: Not since cylc-6.8.0.

doc/rose-rug-files.html
matthewrmshin: Not since cylc-6.8.0.

matthewrmshin:
Please note that `--host=HOST` is no longer necessary with cylc-6.8.0.

doc/rose-rug-files.html
matthewrmshin: Done.

arjclark:
Cool, confirmed. Looks ok to me.

arjclark:
For reference, my above comments came out of testing with an earlier version of cylc (don't ask). It'd be a concern if we **didn't** recommend the equivalent latest version of cylc to be used with the Rose release but I think this is fine and ties in with our long term goals of moving Rose functionality into cylc.

PR #1866, opened 2016-03-29
rose suite-run --reload: allow suite title change

### LINKED ISSUE #1846: rose suite-run --reload after suite `title` change
Reported by users at NIWA:  If I change the title of a running suite, `rose suite-run --reload` fails:
```
$ rose suite-run --reload 
[INFO] delete: log/rose-suite-run.conf
...
[INFO] rug: will reload on wrh-1
[FAIL] cylc reload --force rug  # return-code=1, stderr=
[FAIL] security reasons
```
and I get "connection denied" for subsequent client connection attempts.   A `cylc reload` after changing the title in the installed suite definition doesn't have this affect.  I guess `rose suite-run` is nuking the passphrase, but shouldn't be doing so before reloading a running suite (and is it hashing the suite title to make passphrase?)

matthewrmshin:
Don't call `cylc refresh` on reload mode, as it may nuke the passphrase
by calling un-register and register on the suite.

See also cylc/cylc#1774. Close #1846,

### DIFF
--- a/lib/python/rose/suite_engine_proc.py
+++ b/lib/python/rose/suite_engine_proc.py
@@ -391,7 +391,8 @@ class SuiteEngineProcessor(object):
         """Run suite engine dependent logic (at end of "rose suite-clean")."""
         raise NotImplementedError()
 
-    def cmp_suite_conf(self, suite_name, strict_mode=False, debug_mode=False):
+    def cmp_suite_conf(
+            self, suite_name, run_mode, strict_mode=False, debug_mode=False):
         """Compare current suite configuration with that in the previous run.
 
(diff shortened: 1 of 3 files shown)


matthewrmshin:
@arjclark @kaday please review. I would not agonise over this very much given that the root cause has been fixed in the related pull request in cylc.

arjclark:
Looks good to me, test battery passing in my environment. As you say, most of the fix for this is over in the cylc pull request and this is a minor change.

Over to you @kaday

kaday:
Code looks okay and test battery passes in my environment.

PR #1871, opened 2016-04-07
rose bush: cycles: don't display empty cycle

matthewrmshin:
Cycles with only waiting tasks should not be displayed.


matthewrmshin:

arjclark:
Huzzah! Looks good to me, manually tests as working and test battery passing in my environment. Over to you @kaday

kaday:
Works in my environment. Test battery all pass in my environment and code looks okay to me.

PR #1872, opened 2016-04-08
rose bush: job entry: link 0-size file any way

matthewrmshin:
There may be a delay before the batch system completes writing data to `job.err` and `job.out`. On job completion, they may appear to be zero-size, and get registered to the database as such. The files are then populated with data. This change ensures that these files get linked.

### DIFF
--- a/lib/html/template/rose-bush/job-entry.html
+++ b/lib/html/template/rose-bush/job-entry.html
@@ -67,7 +67,7 @@
             {% elif key == "02-err" -%}
               {% set key = "err" -%}
             {% endif -%}
-            {% if log.exists and log.size -%}
+            {% if log.exists -%}
               {% if log.path_in_tar -%}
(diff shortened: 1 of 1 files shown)


matthewrmshin:

kaday:
Looks okay to me.
Passes all tests in my environment.

arjclark:
Looks ok to me, simple enough change. Have you got anything to test this against knocking around (don't think the existing test-battery covers this, which ideally we should add in)?

matthewrmshin:
No, our test battery is only able to handle data. It is unable yet to test contents of templates. This is a template only change so it is not yet covered.

arjclark:
In lieu of any existing test available, have manually tested by running the following sqlite operation on suite database:
```
sqlite> update task_job_logs set size = 0;
```
Running rose bush on this branch vs. master confirms the various links are now active.

PR #1875, opened 2016-04-11
Handle SuiteNotRunning error in rosie go.

### LINKED ISSUE #1874: rosie go: crash on launching cylc gui for non-running suite
rosie go is throwing an error on trying to launch the cylc gui for a non-running suite:

rosie-crash

We should handle this error.

arjclark:
Closes #1874 

Adds a dialog to capture the SuiteNotRunning exception being thrown on trying to launch gcylc from withing rosie go for a non running suite.

@matthewrmshin - please review 1
@kaday - please review 2

### DIFF
--- a/lib/python/rosie/browser/__init__.py
+++ b/lib/python/rosie/browser/__init__.py
@@ -160,6 +160,8 @@ DIALOG_MESSAGE_DELETE_CONFIRMATION = (
     "\"{0}\". This will permanently delete all branches, " +
     "trunk and any local copy." +
     "\n\nAre you sure you wish to proceed?")
+DIALOG_MESSAGE_SUITE_NOT_RUNNING = ("Cannot launch gcontrol: " +
+                                     "suite {0} is not running.")
 DIALOG_MESSAGE_UNCOMPLETED_FILTER = "Uncompleted filter"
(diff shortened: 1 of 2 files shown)


kaday:
Looks good to me.

lib/python/rosie/browser/__init__.py
```diff
@@ -160,6 +160,8 @@
     "\"{0}\". This will permanently delete all branches, " +
     "trunk and any local copy." +
     "\n\nAre you sure you wish to proceed?")
+DIALOG_MESSAGE_SUITE_NOT_RUNNING = ("Cannot launch gcontrol: " +
+                                    "suite {0} does not appear to be running.")
```
matthewrmshin: A minor point: the message is basically the same as what is returned by `SuiteNotRunningError.__str__`. I am fine with this change as long as we think this is acceptable.

arjclark: Just tweaking slightly to use the __str__ instead to reduce redundancy.

arjclark:
@matthewrmshin - feedback addressed. If you're happy I think we're go for merge.

arjclark:
@matthewrmshin - feedback addressed. If you're happy I think we're go for merge.

PR #1885, opened 2016-04-30
t: allow tests to pass on Ubuntu 16.04LTS

matthewrmshin:
The rose-suite-scan/00 test is passing dubiously any way. After the port file
change in cylc, it is running `grep` on the wrong thing. The problem was picked
up under Ubuntu 16.04 due to it being more unstable.

Ubuntu runs dash instead of bash by default, so the double bracket syntax for
numeric comparison in the new rose_bunch test does not work. Modifying it to use
the old `test` command allows the test to work as intended.


matthewrmshin:
@arjclark please review.

arjclark:
Looks ok to me. Modified tests running fine on my RHEL 6.7 desktop. Have no way of checking for Ubuntu 16.04LTS so will take your word for it.

PR #1889, opened 2016-05-10
change command/environment scripting to script/env-script  

matthewrmshin:
Plus:
* 1 instance of `post-script` from `post-command scripting`.
* Improve rose-rug-simple `suite.rc`.
* Run `teardown` before the final set of tests under `t/rose-macro/10-trigger-check.t`.

### DIFF
--- a/t/rose-task-run/04-run-path-empty/suite.rc
+++ b/t/rose-task-run/04-run-path-empty/suite.rc
@@ -24,8 +24,8 @@ install => my_task_1
            submission timeout = PT1M
            execution timeout  = PT1M
     [[install]]
-        command scripting = """rose task-run"""
+        script = """rose task-run"""
     [[my_task_1]]
-        command scripting = """
+        script = """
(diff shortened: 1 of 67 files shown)


t/rose-task-run/04-run-path-empty/suite.rc
```diff
@@ -24,8 +24,8 @@ install => my_task_1
            submission timeout = PT1M
            execution timeout  = PT1M
     [[install]]
-        command scripting = """rose task-run"""
+        script = """rose task-run"""
```
arjclark: As you're refactoring can you remove the """ round this while you're at it

matthewrmshin: I think we should do this as a separate pull request. We do have a lot of unnecessary quoted values in our test battery suites.

arjclark:

Over to you @oliver-sanders

oliver-sanders:
Look OK, test battery successful.

PR #1895, opened 2016-05-19
rose bush: fixed search url

oliver-sanders:
Fixes problem where rose bush is not hosted from root.

@matthewrmshin Please Review
@benfitzpatrick Please Review


matthewrmshin:
Change looks good enough.

PR #1898, opened 2016-05-20
rose-stem: Prepend hostname to working copies in source trees

scwhitehouse:
To enable building on a remote host, we need to prepend the hostname to source trees which are working copies so `SOURCE_FOO=/working/copy` becomes `SOURCE_FOO=hostname:/working/copy`. This syntax is supported by fcm-make and will enable build tasks to run on remote machines whilst still including working copies on the host where the rose-stem command was invoked.

This will **not** allow rose-stem to be run using the syntax `rose stem --source=hostname:/working/copy` though, that would require more significant changes to both Rose and FCM. It should however allow the current behaviour to work while build tasks are executed on remote machines.


scwhitehouse:
@steoxley this change should be transparent to you if you just use `SOURCE_FOO` and/or `SOURCE_FOO_BASE` in fcm-make, but if you've got scripts which also use them, you'll need to be aware that the hostname will now be prepends if the variables are working copies.

scwhitehouse:
After some discussion, we've decided to keep the existing variables as they are and introduce `HOST_SOURCE_FOO` and `HOST_SOURCE_FOO_BASE` variables which include the prepended host where appropriate.

scwhitehouse:
This seems to work for us, and shouldn't affect other systems since the original variables are now unchanged. @matthewrmshin would you mind reviewing when you get a minute? Thanks.

arjclark:
@scwhitehouse - @matthewrmshin is now away till next week so there'll be a delay on getting this one reviewed.

lib/python/rose/stem.py
```diff
@@ -359,11 +360,24 @@ def _check_suite_version(self, fname):
         if not suite_rose_stem_version == ROSE_STEM_VERSION:
             raise RoseStemVersionException(suite_rose_stem_version)
 
+    def _get_localhost(self):
+        """Return the local host on which the rose-stem command was invoked."""
+        return socket.gethostname()
```
matthewrmshin: You may want to use the `HostSelector.get_local_host` method in `rose.host_select` instead.

lib/python/rose/stem.py
```diff
@@ -359,11 +360,24 @@ def _check_suite_version(self, fname):
...
+
+    def _prepend_localhost(self, url):
+        """Prepend the local hostname to urls which do not point to repository
+        locations."""
+        if ('svn:' not in url and 'fcm:' not in url and 'http:' not in url
+                and 'https:' not in url and 'svn+ssh:' not in url):
```
matthewrmshin: This may be shorter:

`if ':' in url and url.split(':', 1)[0] not in ['svn', 'fcm', 'http', 'https', 'svn+ssh']:`

matthewrmshin:
Some minor comments.

Looks OK otherwise. Tests OK (with and without global configuration) in my environment.

lib/python/rose/stem.py
scwhitehouse: I'm not sure that logic is quite correct, I think it's `if ':' not in url...` at which point it's probably unnecessary as if there's no `:` in `url` the split won't contain one of those four strings?

scwhitehouse:
I think that's the logic we need; if it contains a `:` and the part before the colon is **not** one of fcm, svn, etc, prepend the host (because the source doesn't appear to be recognised repository location, assume it's a local location), else if it doesn't contain a `:`, it's a working copy so prepend the host. 

Also changed sock for HostSelector

lib/python/rose/stem.py
matthewrmshin: In which case, it should be `if ':' not in url or ...`

scwhitehouse: Fixed!

scwhitehouse:
Using Matt's simplification of the logic it now passes the rose-stem test-battery tests, and it also sends sensible things to the UM rose-stem suite (checking the variables being set by running with `-vv`).

matthewrmshin:
@stevewardle please review 2 as discussed.

stevewardle:
Looks okay to me overall, the only things I'm not so sure about are:

-  `_get_localhost` is only called in one place (from `_prepend_localhost`) and it is only directly deferring to another single-line call, could that call to `HostSelector` not just be in-lined inside `_prepend_localhost`?
- Related to the above, since both of these (or just `_prepend_localhost` if they are combined as above) aren't using `self` and so could they be updated to be `@staticmethod`s?

matthewrmshin:
Now that @stevewardle has mentioned it... I think you can also initialise `HostSelector` once under `StemRunner.__init__`, so you can take advantage of the caching `HostSelector` does with calls to functions in the `socket` module.

scwhitehouse:
I think @matthewrmshin's comment supercedes @stevewardle's, I've initialised the `host_selector` in the `__init__` method (with hopefully sensible arguments) and then used that directly in the `_prepend_localhost` method rather than having a separate method.

stevewardle:
Looks fine to me @scwhitehouse - assuming @matthewrmshin is also happy I'll merge it

matthewrmshin:
Looks OK.

PR #1930, opened 2016-06-30
Correct rose bunch doc "instances" entries to "command-instances"

arjclark:
@benfitzpatrick - instances references corrected. Should only need 1 review as is a typo correction.


arjclark:
@benfitzpatrick - please review

PR #1935, opened 2016-07-05
t: rose task-env: ROSE_DATAC__????

matthewrmshin:
Ensure a future cycle offset generates a `ROSE_DATAC__????` environment variable.

@arjclark please review and/or reassign. Looks like we don't have a test for this documented functionality.


arjclark:
Looks ok to me, test battery working in my environment. Is only a test addition for existing functionality so don't think it needs review 2-ing.

PR #1942, opened 2016-07-12
t: modify tests for cylc/cylc#1929

matthewrmshin:
Modify tests for cylc/cylc#1929. Tests should also work with current cylc master.


matthewrmshin:
@arjclark please review.

arjclark:

PR #1944, opened 2016-07-14
t: rose-config-edit-00: fix py.test argument

matthewrmshin:
Detect missing "mock" Python module in skip logic.
Supply `${TEST_SOURCE_DIR}` as argument to `py.test`.

Follow up of #1929 and #1941.


matthewrmshin:
@oliver-sanders please sanity check. The change has been tested on the VM.

oliver-sanders:
Looks good.

PR #1947, opened 2016-07-18
file install: reduce commits to sqlite database

matthewrmshin:
Delete and update statements are now grouped together, and executed using `executemany`.

We now have a maximum of 3 commits only:
* After tables are created.
* After non-parallel installation are done.
* After parallel installation are done.

@oliver-sanders @arjclark please review, sanity check or reassign.


matthewrmshin:
In my tests, based on the 50-file test from @dpmatthews, this change offers up to 6 fold reduction in elapsed time on various file systems in my environment.

oliver-sanders:
Looks good, I'm seeing ~3 fold reduction.

arjclark:
Looks OK to me. No problems from the test-battery in my environment. Testing outlined above indicates an improvement in performance so sounds good to me.

PR #1961, opened 2016-08-16
Fix t: rose-bush/00, broken since cylc/cylc#1827

matthewrmshin:
Fix test broken since cylc/cylc#1827. The `-p` option means that we are copying the symbolic link. The `-r` option is unnecessary. @arjclark please review.


arjclark:
Looks OK to me. No problems from the test-battery in my environment, tested against head of cylc. Doesn't need review 2'ing.

PR #1968, opened 2016-08-26
rose bush: use file system for job logs listing

matthewrmshin:
Instead of using the suite runtime database table, use file system
listing and/or the listing of the members in relevant
`log/job-CYCLE.tar.gz` files.
* The change does not appear to have much performance impact for the job lists.
* It appears to be much faster when viewing a job log file.
* The only behaviour change is that jobs with their logs removed will no longer have their log files listed.

In view mode, fix no job entry display when file is TAR-GZ archived.

### DIFF
--- a/lib/html/template/rose-bush/job-entry.html
+++ b/lib/html/template/rose-bush/job-entry.html
@@ -73,7 +73,7 @@
                 {% else -%}
                   <a
                   {% if log.size == "?" or not log.size -%}class="text-muted"{% endif -%}
-                  href="{{script}}/view/{{user}}/{{suite}}?path={{log.path|replace('+', '%2B')}}&amp;path_in_tar={{log.path_in_tar}}"
+                  href="{{script}}/view/{{user}}/{{suite}}?path={{log.path|replace('+', '%2B')}}&amp;path_in_tar={{log.path_in_tar|replace('+', '%2B')}}"
                   title="{{log.size}} bytes">{{key}}{% if log.size == "?" -%}?{% endif -%}</a>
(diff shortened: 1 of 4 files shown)


arjclark:
Still reviewing - probably won't get this done properly till next week but initially looks good. My only initial concern would be that all the os listing calls are potentially bad for a rose bush instance to be running on a Lustre-esque file system. As a thought, if that is the case I wonder if we might have a "filesystem friendly" mode that has either dumb list everything or slightly smarter databased based behaviour instead (as we have now)? Don't know what you think about that @matthewrmshin?

matthewrmshin:
Branch re-based.

arjclark:
Discussed filesystem things with @matthewrmshin and am happy shouldn't present an issue. Review 1 with some manual testing looks good to me.

arjclark:
Over to you @oliver-sanders

oliver-sanders:
I'm seeing 404 errors when trying to open tar'ed files caused by a `+` symbol in the `path_in_tar` not being escaped as for the `path` (using `%2B`).

Example URL:
http://127.0.0.1:8080/view/osanders/prune?path=log/job-20000101T0000%2B01.tar.gz&path_in_tar=job/20000101T0000+01/bar/01/job.out

matthewrmshin:
The `path_in_tar` escape problem does not appear to be related to this change. It requires a simple fix to the template, so I have included it in this change.

matthewrmshin:
All should be good.

oliver-sanders:
Looks OK, tested as working.

PR #1972, opened 2016-09-01
rose bush: only count cycles with things in for cycle counts.

arjclark:
Currently, if a cycle has waiting task proxies but no tasks in it that have run then it doesn't get displayed on the cycles page of rose bush. However, it does get counted at the bottom of the page "Entries X of Y" part. This can result in rose bush pages where the cycle counts are dramatically different from those being displayed.

This tweaks the query used to generate that count.

@matthewrmshin - please review (fixes the rose bush part of the problem seen in the VER suite the other day)

### DIFF
--- a/lib/python/rose/suite_engine_procs/cylc.py
+++ b/lib/python/rose/suite_engine_procs/cylc.py
@@ -806,9 +806,10 @@ class CylcProcessor(SuiteEngineProcessor):
 
         """
         of_n_entries = 0
+        stmt = ("SELECT COUNT(DISTINCT cycle) FROM task_states where " +
+                "status is not 'waiting'")
         for row in self._db_exec(
-                self.SUITE_DB, user_name, suite_name,
-                "SELECT COUNT(DISTINCT cycle) FROM task_states"):
+                self.SUITE_DB, user_name, suite_name, stmt):
             of_n_entries = row[0]
             break
         if not of_n_entries:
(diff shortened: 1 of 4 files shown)


matthewrmshin:
I wonder if you can add a test? For reference, `t/rose-bush/07-cycles-page.t` may be a good starting point for a new test.

lib/python/rose/suite_engine_procs/cylc.py
```diff
@@ -806,9 +806,10 @@ def get_suite_cycles_summary(
 
         """
         of_n_entries = 0
+        stmt = ("SELECT COUNT(DISTINCT cycle) FROM task_states where " +
+                "status is not 'waiting'")
```
matthewrmshin: Can we stick with ALL CAPS for SQL keywords?

arjclark: Will do.

arjclark:
@matthewrmshin - test added and query changed slightly (testing showed it was still miscounting in the test case added here).

arjclark:
@matthewrmshin - fixed those two broken tests, sorry!

matthewrmshin:
Now looking good. Tests OK in my environment. @oliver-sanders please sanity check.

oliver-sanders:
Looks sane, passes tests.

PR #1974, opened 2016-09-05
t: app-run 14: improve stability

matthewrmshin:
A `namelist:` source is always considered out of date, so target is
always rebuilt, and so the modification time (via `ls -l`) of the target is not
something we should compare. The test should be stable if we only compare the files' SHA1 sums.


matthewrmshin:
@arjclark please review.

arjclark:
Looks ok to me. Tested as follows:

Ran test on master 10x - 1 failure
Ran test on branch 10x - 0 failures

Seems slightly more stable to me.

PR #1979, opened 2016-09-09
Fix breakage by cylc/cylc#1953 and cylc/cylc#1958

matthewrmshin:
cylc/cylc#1953 breaks `rose task-env` (and its derivatives). cylc/cylc#1958 breaks 2 similar tests. This change should fix both issues.

@arjclark please review or reassign.


arjclark:
Looks good to me. Tests as working against current cylc master. Review 2 not needed.

PR #1985, opened 2016-09-20
t: improve test robustness

matthewrmshin:
Some tests in Rose test battery have become less robust since
cylc/cylc#1923 is merged in. This change improves the situation by
removing some of the old logic for detecting when a suite is running.

The change also removes `rose suite-hook` tests as the tests are some
what unstable and the command itself has now been replaced by better
cylc built-in functionality.


matthewrmshin:
@oliver-sanders @benfitzpatrick please review and sanity check.

matthewrmshin:
Superseded by #1986.

PR #1988, opened 2016-09-28
rose bush: jobs list: fix paths in tar links

matthewrmshin:
Links contain random files in the tar file. This change fixes the filter
in the logic.


matthewrmshin:

arjclark:
Looks OK to me. No problems from the test-battery in my environment. Over to you @oliver-sanders

PR #1996, opened 2016-10-07
New version of rose_ana

stevewardle:
This is the new version of `rose_ana` - it moves much of the logic out of Rose itself and into modular "plugin"-like classes which can be loaded from one of 4 locations (in order of priority):
1. In the `ana` directory of the `rose_ana` app being run.
2. In the `ana` directory of the top-level of the suite.
3. In a (site specific, central) directory given by an entry in the Rose configuration.
4. In the `ana_builtin` directory of Rose itself (only one module exists here for now).

An abstract base class within `rose_ana` provides the appropriate form to use, and automatically passes all options passed to each section as a dictionary; it is therefore up to the implementation of the class to process and check the options it is expected to handle.  

It's more permissive than the previous code - treating any `Exception` raised as a failure but without completely crashing the entire app.

To aid in the transition between the old-style `rose_ana` and the new (since the syntax is not backwards compatible at all) the old `rose_ana` code still exists but has been renamed.  If the new syntax (where each section starts `[ana:`) is not detected it will automatically run the old version (and issue a warning).

The test battery has been updated (with the existing `rose_ana` tests duplicated and updated so for now it is testing both the old + new forms).  We can probably expand the testing a little (and, of course provide some updated documentation) but for now we can start discussing the implementation itself @matthewrmshin @scwhitehouse @steoxley


scwhitehouse:
This seems sensible to me, though I'd quite like some documentation!

arjclark:
I assume things like the tutorial will get updated once the principal of the change is agreed?

steoxley:
Following a few minor changes by Steve I'm happy this implementation meets our requirements. As discussed it would be useful if:

- For the grepper.FilePattern utility, only differences are reported - perhaps the first 50 (configurable) or so.
- The database aspects for known good output can be disabled by an environment variable for systems which don't make use of this. In fact, perhaps it should be off by default as more systems don't make use of this feature than do?

stevewardle:
I think the commits above address the issues raised by @steoxley -

- KGO database functionality now defaults to _off_ and is enabled via the `[rose-ana]` section of the Rose configuration.  Users can opt-in as required (presently only the UM Systems team members at the Met Office will likely want to do this).

- Reports from `grepper` are now less verbose on success (except for when higher verbosity is asked for in the call to `rose task-run`

- An output limiting option is added to the `[rose-ana]` section of the Rose configuration.  If present it must be a numerical value - the maximum number of reports `grepper` will print for each analysis task.  We can either let this default to "no limit" or sites may impose their own limit that can be overidden by users (if we do this I still suggest going with quite a high limit)

stevewardle:
Now rebased + squashed and with updates to the documentation

scwhitehouse:
rose-rug-stem.html: 

- It'd be nice to have an example of kgo_file, the syntax is unclear to me, is it kgo_file=1 or kgo_file=/home/kgo/file.txt
- log/rose-ana-comparisons.db - is this still true? Are we just having one database? I thought we were going for multiple files? I've lost the thread of this debate sorry.

rose-rug-advanced-tutorials-rose-stem.html:
- "Replace the /home/user/spaceship path in the kgo1file variable with the path to this file" needs updating

stevewardle:
@scwhitehouse  I agree there should be an example of `kgo_file` - it should be set to the index of the `files` list (and is zero-indexed being Python).  I'll ensure that's clear, although hopefully only those that need to use it will be bothered (and it defaults to being off).

The comparisons database is indeed still a single database - we're going to give it one more chance with a hopefully improved locking mechanism before we go further.  (plus, we can disable the database entirely for sensitive cases that don't really benefit such as nightly testing)

PR #1998, opened 2016-10-18
rose_prune: fix bad fail message on non-suite host

matthewrmshin:
rose_prune attempts to connect to the cylc suite run time database to
get a list of job hosts. If rose_prune was being invoked on non-suite
host, it would return a rather cryptic fail message. This change should
fix the issue.


matthewrmshin:
@arjclark @oliver-sanders please review and sanity check.

arjclark:
Looks OK to me. No problems from the test-battery in my environment. Over to you @oliver-sanders

PR #2001, opened 2016-10-27
rose macro: don't run transform macros in validate mode.

### LINKED ISSUE #1923: rose macro: transform macros running in -V and not --fix mode
rose macro would appear to be running transformer macros in validate (`rose macro -V`) mode but not running them when invoked by `rose macro --fix`.

Problem was observed with a rose.macro.MacroBase subclass containing both a validate and transform method.

oliver-sanders:
Closes #1923 

If a transform and a validate method were defined in the same class then the transform macro would be run by `rose macro -V`.

@arjclark Please review
@matthewrmshin Please review

### DIFF
--- a/t/rose-macro/25-suite-level.t
+++ b/t/rose-macro/25-suite-level.t
@@ -1,0 +1,109 @@
+#!/bin/bash
+#-------------------------------------------------------------------------------
+# (C) British Crown Copyright 2012-6 Met Office.
+#
+# This file is part of Rose, a framework for meteorological suites.
+#
+# Rose is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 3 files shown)


arjclark:
Test for this?

t/rose-macro/25-suite-level.t
```diff
@@ -0,0 +1,109 @@
...
+
+    def validate(self, config, meta_config=None):
+        node = config.get(['env', 'FOO'])
+        if node is not None or not node.is_ignored():
+            if node.value not in ['foo']:
+                self.add_report('env', 'FOO', node.value, 'Not foo enough')
+        return self.reports
+
+    def transform(self, config, meta_config=None):
+        # This is here to ensure transform macros are not run with the -V
```
oliver-sanders: This addition tests that transform macros aren't being run by `rose macro -V`

arjclark:

@matthewrmshin - please eyeball accordingly.

PR #2013, opened 2016-11-16
Fixed environment variable behaviour.

### LINKED ISSUE #2012: rose edit: spin button not working correctly with environment variables
For integer spin widgets, where an environment variable has been set for the value the variable should appear as text in the widget. At present the widget's value is being set to zero instead.

Possibly introduced by #2004 and or #2010.

oliver-sanders:
Closes #2012 
Addresses #1512 

Fixes logic introduced in #2010.

The new behaviour for environment variables in rose edit:
- Always display a text field if there is an environment variable in the value UNLESS an option widget (radio/combo) should be displayed AND the environment variable is listed as an allowed value.

@arjclark Please review / assign if necessary.


oliver-sanders:
Behaviour cleared with a.voysey

arjclark:
Looks good. Have run through with Adam and is doing what we both expect.

PR #2014, opened 2016-11-21
rosie: document search SQL wildcard characters

### LINKED ISSUE #1681: rosie suite search wildcard fake-out problem
This caught me out for a while on the MOSRS shared suite repository.  I will nominate myself for an "idiot of the month" award, but maybe it does raise an interesting (if small!) issue:

Rosie suite search seems not to support wildcard characters or regular expressions.  However, I'd argue that if you don't know that, it's natural to try, for example, entering `*` in order to see "all suites".   If I do this the search engine  returns a bunch of seemingly unrelated suites, which might suggest to me (see "idiot" above) that the wildcard search worked as intended.  However, as I've just found out, the asterisk is interpreted literally and I was seeing only suites with `access-list=*` in their `rose-suite.info` files. :grimacing: 

Would it be worth adding some small print under the web search entry, and expanding the GUI search tooltip, to explain exactly how the entered terms are used in searching?

matthewrmshin:
Also correctly support a single quoted string (with space characters) as
a search term.

Close #1681. @oliver-sanders @arjclark please review.

### DIFF
--- a/bin/rosie-go
+++ b/bin/rosie-go
@@ -38,6 +38,10 @@
 #     An address URL may contain shell meta characters, so remember to put it
 #     in quotes.
 #
+#     Search strings may contain SQL wildcard characters. E.g.
+#     % (percent) is a substitute for zero or more characters.
+#     _ (underscore) is a substitute for a single character.
+#
 #     If no argument is specified, rosie go will display a list of all your
 #     locally checked out suites.
 #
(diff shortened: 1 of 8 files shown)


oliver-sanders:
Looks fine to me. @arjclark all yours.

arjclark:
Looks good to me.

PR #2015, opened 2016-11-21
Document switch off metadata use case

### LINKED ISSUE #1220: rose config-edit: per-item option to take $VAR regardless of metadata
As an example, we have a bunch of MOS tasks that take a "learning period" in weeks as input.  If a suite only uses one MOS task it makes sense to put the learning period in the rose app config and not in the suite definition, and use metadata to control type and range (integer, this >=0).  But if a suite uses several MOS tasks they all need to share the same learning period so it is better to set $LEARNING_PERIOD once in the suite definition, and reference the suite variable in the app config file - but this can't be done without deleting the metadata for type checking etc. which is still useful information.  So, how about a right-click option for "accept environment variable" for any config item, which would disable type checking etc. if the need arises?  (or is there already some way of handling this situation?).

matthewrmshin:
Should be sufficient to close #1220.

### DIFF
--- a/doc/rose-rug-config-edit.html
+++ b/doc/rose-rug-config-edit.html
@@ -270,6 +270,19 @@ rose config-edit -C ~/my_suites/this_suite
               </tr>
             </table>
 
+            <h3 id="switc-off-metadata">Switching off metadata</h3>
+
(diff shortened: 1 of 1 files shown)


matthewrmshin:
@oliver-sanders please review.

PR #2017, opened 2016-11-22
rosie go: use getpwuid to get current user ID

matthewrmshin:
Use `pwd.getpwuid(os.getuid()).pw_name` instead of `os.getlogin()`. The latter can raise mysterious `OSError` in some environments. (Note: all other instances in our code use the former to get the current user ID.)

Reported by @dpmatthews.

### DIFF
--- a/lib/python/rosie/browser/main.py
+++ b/lib/python/rosie/browser/main.py
@@ -20,8 +20,8 @@
 """Implement "rosie go"."""
 
 import ast
-import functools
 import os
+import pwd
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


arjclark:
Looks ok. Are you (or @dpmatthews) able to elaborate more on the problem so the rest of us can understand what's going on/be aware of symptoms?

matthewrmshin:
Users have reported this in their environment. (I have also independently reproduced this in one of my virtual machines.)

```
>>> import os
>>> os.getlogin()
OSError: [Errno 2] No such file or directory
```

Whereas `pwd.getpwuid(os.getuid())` works as expected.

arjclark:
Yup good to merge.

PR #2018, opened 2016-11-22
rose macro: added --transform command line option

oliver-sanders:
Adds a command line option for running / listing only default rose macros.

@matthewrmshin @arjclark Please review.


matthewrmshin:
@scwhitehouse please review and ensure this new interface is acceptable for your use cases.

scwhitehouse:
I've discussed this with me team, and we think that maybe keeping the previous behaviour for `--fix` (running the default internal macros) and introducing a new `--transform` option (running all transform macros) might be better than changing the behaviour of `--fix` and having `--default-only`. This is really just nomenclature so I'm open to other opinions!

oliver-sanders:
How about:
``-F``, ``--fix`` Run all internal transformer macros.
``-T``, ``--transform`` Run all transform macros
``-V``, ``--validate`` Run all validate macros

Any desire for a ``--default-only`` option as well (for ``-V --default-only``)?
@metomi/core, @scwhitehouse

scwhitehouse:
I don't have a use-case myself for a default-only option for `--validate`, but it might be useful to someone. I'm not sure what your policy is on adding code just in case someone wants it - it clutters the code and adds maintenance overhead, but might make somebody's day in a "that's just what I was looking for" way. (Similarly, I've no idea if there should be a `--report` option to run all reporter macros, just for symmetry with the other options).

I'm happy without, I'll defer to your team's greater wisdom on this!

arjclark:
@oliver-sanders - I'm happy to take @scwhitehouse's opinion on this at this time as they're the only ones to have noticed so far. I guess you could talk to Mike T. for a "expert user opinion".

oliver-sanders:
@scwhitehouse @arjclark Have now changed to the CLI proposed in https://github.com/metomi/rose/pull/2018#issuecomment-262495213.

matthewrmshin:
Failing syntax test on Travis CI. Please fix.

doc/rose-command.html
```diff
@@ -1667,6 +1667,10 @@ <h2 id="rose-macro">rose macro</h2>
 
               <dd>Run only for suite level macros.</dd>
 
+              <dt><kbd>--transform, -T</kbd></dt>
+
+              <dd>Prepend all transformers to the argument list.</dd>
```
arjclark: Better to be clear and say "all transformer macros"

bin/rose-macro
```diff
@@ -50,6 +51,8 @@
 #         Reduce verbosity.
 #     --suite-only
 #         Run only for suite level macros.
+#     --transform, -T
+#         Prepend all transformers to the argument list.
```
arjclark: Better to be clear and say "all transformer macros"

PR #2024, opened 2016-12-06
rose docs: broken link detection

oliver-sanders:
Added a test to detect broken links. Detects:
- External Links
    - Page exists and returns a sensible HTTP code.
- Internal Links
    - Page exists
    - URL fragments (e.g. example.com#fragment) reference sections that are present

Fixed many broken links.

Caveats:
- At present some URLs fail for SSL reasons, the crude solution implemented is to retry the affected URLs with SSL verification turned **OFF**! As this testing will only be done on URLs we have put into the docs. If you think this is a problem I can spend some time trying to get around it.
- Does not check URL fragments on external sites (would require downloading the whole page).

@matthewrmshin @arjclark Please Review.

### DIFF
--- a/t/docs/00-urls.t
+++ b/t/docs/00-urls.t
@@ -1,0 +1,30 @@
+#!/bin/bash
+#-------------------------------------------------------------------------------
+# (C) British Crown Copyright 2012-6 Met Office.
+#
+# This file is part of Rose, a framework for meteorological suites.
+#
+# Rose is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Rose is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Rose. If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test "rose date".
+#-------------------------------------------------------------------------------
+. $(dirname $0)/test_header
+#-------------------------------------------------------------------------------
+tests 2
+#-------------------------------------------------------------------------------
+# Validate URLs in the rose docs (see test 01-validate-validator.t for defailt.
+run_pass ${TEST_KEY_BASE} python $TEST_SOURCE_DIR/lib/python/urlvalidator.py "${ROSE_HOME}/doc/"
+file_cmp "${TEST_KEY_BASE}.err" "${TEST_KEY_BASE}.err" <'/dev/null'
+
+exit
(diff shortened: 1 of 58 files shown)


t/docs/lib/python/urlvalidator.py
```diff
@@ -0,0 +1,184 @@
+"""A crude URL validator"""
```
matthewrmshin: Please add usual header.

t/docs/lib/python/urlvalidator.py
```diff
@@ -0,0 +1,184 @@
...
+    report_invalid_urls(int_errors, ext_errors, frg_errors, malformed)
+
+    if int_errors or ext_errors or frg_errors or malformed:
+        sys.exit(1)
+    sys.exit(0)
+
+if __name__ == '__main__':
+    if len(sys.argv) != 2:
+        print ('Usage: Please supply root directory to validate the contents '
+               'of.')
```
matthewrmshin: Don't think you need a usage check, given that it is only used in our tests!

t/docs/lib/python/urlvalidator.py
oliver-sanders: Done.

t/docs/lib/python/urlvalidator.py
oliver-sanders: Might as well check we have the correct number of arguments and provide some form of CLI documentation to future developers.

matthewrmshin: In which case, it ought to go to STDERR, or use `sys.exit(...)`.

oliver-sanders:
Added a test for the rose macro (advanced) tutorial.

t/docs/02-macro-tutorial.t
```diff
@@ -0,0 +1,30 @@
...
+#
+# Rose is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Rose. If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test "rose date".
```
arjclark: Needs changing to what the test is.

t/docs/01-validate-validator.t
```diff
@@ -0,0 +1,39 @@
...
+#
+# Rose is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Rose. If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test "rose date".
```
arjclark: Needs changing to what the test is

t/docs/00-urls.t
```diff
@@ -0,0 +1,30 @@
...
+#
+# Rose is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Rose. If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test "rose date".
```
arjclark: Needs changing to what the test is

PR #2031, opened 2016-12-22
rose suite-clean: improve parent directory clean

matthewrmshin:
For suite under sub-directory hierarchy, clean all the way to the root
where possible.


matthewrmshin:
> `rose suite-clean` left behind the empty directory `cylc-run/foo`.

@oliver-sanders It doesn't in my environment, so we'll need to figure out why it does so in your environment. (The new test should also pick up on this if it does.)

PR #2034, opened 2017-01-12
edit: prevent radio buttons from filling full page width

### LINKED ISSUE #1936: rose edit: radio button clickable area extends beyond label
Radio buttons in rose edit are sensitive up to the right hand side of the window which can cause problems for users focusing on a window by clicking on what they assume to be empty grey space.

oliver-sanders:
Close #1936 

Turned out to be a really simple fix, see issue for reproduction instructions.


matthewrmshin:
2nd review not necessary for this trivial change.

PR #2035, opened 2017-01-13
Correct mistake in Jinja2 part of training slides.

arjclark:
Mistakenly left an `endfor` in rather than an `endif` in for the refactored jinja2 part of the taught training slides.


matthewrmshin:
Test failure not related.

PR #2039, opened 2017-01-17
rose app-upgrade: improve remove diagnostic

matthewrmshin:
Print old value of removed setting in diagnostic instead of `None`.

(Problem reported by @paulcresswell via email.)


t/rose-app-upgrade/09-opt.t
```diff
@@ -67,15 +67,17 @@ TEST_KEY=$TEST_KEY_BASE-upgrade-add
 run_pass "$TEST_KEY" rose app-upgrade --non-interactive \
  --meta-path=../rose-meta/ -C ../config 0.2
 file_cmp "$TEST_KEY.out" "$TEST_KEY.out" <<'__OUTPUT__'
-[U] Upgrade_0.1-0.2: changes: 4
+[U] Upgrade_0.1-0.2: changes: 5
```
oliver-sanders: Was this wrong before?

matthewrmshin: It wasn't showing the removal of `env=A` in the optional configuration before, so I guess it was wrong before.

arjclark:
Looks ok to me.

PR #2043, opened 2017-01-23
run default macros on the rose-suite.info file.

### LINKED ISSUE #1766: rosie: metadata further development request
- Command line `rose-suite.info` to use the `sort-key` provided by the metadata.
- Command line `rose-suite.info` to add the trigger ignored `!!` for suite if it would fail with `[FAIL]         State should be trig-ignored`

oliver-sanders:
Close #1766 

`rose macro` now runs default macros on the rose-suite.info file.

### DIFF
--- a/bin/rose-macro
+++ b/bin/rose-macro
@@ -26,6 +26,14 @@
 # DESCRIPTION
 #     List or run macros associated with a suite or application.
 #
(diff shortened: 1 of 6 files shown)


matthewrmshin:
(A syntax test is failing on Travis CI.)

PR #2047, opened 2017-02-07
Fixes for some issues with rose_ana KGO database

stevewardle:
Two slight problems have been identified with this functionality (though the impact is limited, only to those making use of the KGO database).  Two different problems exist:

**1.** The `rose_ana` re-write moved to using a buffer for the statment (to avoid many repeated calls to locking and/or commits to the database.  Unfortunately this defeats the point of the `tasks` database table - the original idea was that this table would initially be set to `1` for each task, and updated to `0` when the task finished (allowing scripts interfacing with the database the means to understand when something unexpected had stopped the task completing)...  Due to the use of the buffer everything is done in one go, meaning an unexpected error results in _nothing_ being written instead.  The solution is to have 2 calls per `rose_ana` task instead (one to set the `tasks` table entry at the start)

**2.** At some point during the writing of the analysis modules both externally and for the `grepper.py` class, the `_kgo_db_add_file_pair` method was somehow forgotten/overlooked and tasks instead directly called `self.parent.kgo_db.add_comparison`... this meant losing the behaviour of making the table key unique by appending the `rose_ana` task name.  Rather than fix the call in `grepper.py` and elsewhere it seems a little redundant having the `_kgo_db_*` methods at all in the app class, so these are now dropped

### DIFF
--- a/lib/python/rose/apps/rose_ana.py
+++ b/lib/python/rose/apps/rose_ana.py
@@ -51,12 +51,12 @@ class KGODatabase(object):
     # (as it must uniquely identify each row)
     CREATE_COMPARISON_TABLE = """
         CREATE TABLE IF NOT EXISTS comparisons (
-        app_task TEXT,
+        comp_task TEXT,
         kgo_file TEXT,
         suite_file TEXT,
         status TEXT,
         comparison TEXT,
-        PRIMARY KEY(app_task))
+        PRIMARY KEY(comp_task))
         """
     # This SQL command ensures a "tasks" table exists in the database
     # and then populates it with a pair of columns (the task name and
@@ -79,7 +79,7 @@ class KGODatabase(object):
         self.task_name = "task_name not set"
 
     def enter_comparison(
-            self, app_task, kgo_file, suite_file, status, comparison):
+            self, comp_task, kgo_file, suite_file, status, comparison):
         """Add a command to insert a new comparison entry to the database."""
         # This SQL command indicates that a single "row" is to be entered into
         # the "comparisons" table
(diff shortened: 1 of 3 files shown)


stevewardle:
Okay passing the built-in tests - I'll run it on a proper case in our UM rose suite as well to make sure it solves the problems we were seeing before.  If possible it would be great if this went on (to at least `rose-next` at the Met Office) before 28th Feb - since I currently have had to commit a few workarounds to the UM trunk; it'd be good to be able to revert those before the official release - @matthewrmshin any chance this is possible?

stevewardle:
Apologies for changing it after the review request - it's a trivial change as I spotted that the message in `stdout` looked silly being at the start + end of the app run

lib/python/rose/apps/rose_ana.py
```diff
@@ -80,7 +85,11 @@ def enter_comparison(
         # the "comparisons" table
         sql_statement = (
             "INSERT OR REPLACE INTO comparisons VALUES (?, ?, ?, ?, ?)")
-        sql_args = [app_task, kgo_file, suite_file, status, comparison]
+        # Prepend the task_name onto each entry, to try and ensure it is
+        # unique (the individual comparison names may not be, but the rose
+        # task name + the comparison task name should)
+        sql_args = [self.task_name + " - " + app_task,
```
stevewardle: I can see that it looks that way, but despite the argument names both being `app_task`, one is set by `rose_ana` itself to the Rose task name (the one which is saved to `self.task_name`) but the other one (passed to `enter_comparison`) is supplied by the user, and can be anything - though by convention it uses the section name from the app...  I could perhaps rename the arguments to make this more obvious?

lib/python/rose/apps/rose_ana.py
```diff
@@ -101,7 +112,8 @@ def database_lock(self, lockfile, reporter=None):
         if reporter is not None:
             reporter("Acquired DB lock at: " + time.asctime())
         lock.write("{0}".format(os.getpid()))
-        reporter("Writing results to KGO Database...")
+        if reporter is not None:
+            reporter("Writing to KGO Database...")
```
stevewardle: No I think this is as intended (I just realised it ought to be protected in the same way the other statements are)

PR #2050, opened 2017-02-10
rose suite-clean --only=extglob

### LINKED ISSUE #2029: rose suite-clean --only=extglob
Can the `--only=ITEM` option of `rose suite-clean` support `extglob` (e.g. for negative match)?

matthewrmshin:
Remove Python logic for cleaning localhost - as the Python standard library `glob` cannot handle extended glob. Use the same bash-based logic as remote job host for localhost to implement this functionality and for consistency.

Close #2029.


matthewrmshin:
Doc updated.

PR #2064, opened 2017-04-03
Tweaks to rose-ana based on feedback during integration

### LINKED ISSUE #2054: rose ana: grepper-report-limit
grepper-report-limit can apparently only be set in the user's `.metomi/rose.conf` file. e.g.:

```cfg
[rose-ana]
grepper-report-limit=20
```

It would be useful if this could also be set via Jinja2 in `rose-stem/rose-suite.conf`. The expectation being that if not set here a particular rose-stem implementation would default this to whatever value it deemed sensible. @stevewardle

stevewardle:
Now that the new `rose-ana` has started to see some use a few cases have emerged which could do with addressing...  In particular; it seems desirable to have a way of setting some of the more "global" options from the suite rather than only from within a user's config.... At the same time as this we'll integrate another optional setting that has been requested. 

Closes #2054

### DIFF
--- a/doc/rose-configuration.html
+++ b/doc/rose-configuration.html
@@ -1214,13 +1214,13 @@ source=http://host/svn-repos/path/to/bar.txt
               of key-value option pairs; just like other standard Rose apps. 
               However the availability of options for a given section is 
               specified and controlled by the <i>class</i> rather than the 
-              meta-data.  This makes it easy to provide your own analysis
+              meta-data. This makes it easy to provide your own analysis
               modules without requiring changes to Rose itself.
             </p>
 
             <p>Therefore you should consult the documentation or source code
               of the analysis module you wish to use for details of which 
-              options it supports.  Additionally, some special treatment is 
+              options it supports. Additionally, some special treatment is 
               applied to all options depending on what they contain:
             </p>
 
@@ -1242,7 +1242,7 @@ source=http://host/svn-repos/path/to/bar.txt
 
             <p>The app may also define a configuration section; 
               <code>[ana:config]</code>, whose key-value pairs define app-wide
-              settings that are passed through to the analysis classes.  In 
+              settings that are passed through to the analysis classes. In 
               the same way that the task options are dependent on the class
               definition; interpretation of the <code>config</code> options is
               done by the class(es), so their documentation or source code 
--- a/doc/rose-configuration.html
+++ b/doc/rose-configuration.html
@@ -1239,6 +1239,16 @@ source=http://host/svn-repos/path/to/bar.txt
                 <code>rose task-run</code>
               </li>
             </ul>
+
+            <p>The app may also define a configuration section; 
+              <code>[ana:config]</code>, whose key-value pairs define app-wide
+              settings that are passed through to the analysis classes.  In 
+              the same way that the task options are dependent on the class
+              definition; interpretation of the <code>config</code> options is
+              done by the class(es), so their documentation or source code 
+              should be consulted for details.
+            </p>
+
            </div>
 
           <div id="footer" class="panel-footer">
--- a/doc/rose-rug-stem.html
+++ b/doc/rose-rug-stem.html
@@ -358,6 +358,37 @@ rose stem --group=mygroup --group=myothergroup
 
             <div class="slide">
               <h3 class="alwayshidden">rose ana III</h3>
+              <p>As well as sections defining analysis tasks, <samp>rose_ana</samp>
+              apps allow for one additional section for storing global 
+              configuration settings for the app.  Just like the tasks themselves
+              these options and their effects are dependent on which analysis 
+              tasks are used by the app.</p>
+
+              <p>Therefore we will here present an example using the built-in
+              <code>grepper</code> class.  An app may begin with a section like
+              this:</p>
+
+              <pre class="prettyprint lang-rose_conf" >
+  [ana:config]
+  grepper-report-limit=5
+  skip-if-all-files-missing=.true.
+</pre>
+             <p>Each of these modifies the behaviour of <code>grepper</code>. 
+             The first option suppresses printed output for each analysis task
+             one the specified number of lines have been printed (in this case
+             5 lines).  The second option causes <samp>rose_ana</samp> to skip
+             any <code>grepper</code> tasks which compare files in the case that
+             <i>both</i> files do not exist.</p>
+
+             <p>Note that any options given to this section may instead be 
+             specified in the <code>rose-ana</code> section of the user or site 
+             configuration.  In the case that the same configuration option
+             appears in both locations the one contained in the app file will
+             take precedence.</p>
+            </div>
+
+            <div class="slide">
+              <h3 class="alwayshidden">rose ana IV</h3>
 
               <p>It is possible to add additional analysis modules to 
               <samp>rose ana</samp> by placing an appropriately formatted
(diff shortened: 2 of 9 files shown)


stevewardle:
@steoxley Can you take another look at the functionality please - assuming this is suitable I'll update testing and docs... Briefly what I've done is:
 * `AnalysisTask` object now has a `skipped` attribute rather than just `passed` - if `skipped` is `True` `rose_ana` will report on it differently as well as not considering it a failure overall
 * Whilst poking around I realised passing a reference to the parent app into the `AnalysisTask` objects is super bad practice (shame on me!), so I refactored it to pass references to the specific things it needs instead
 * As discussed I added a sort of "overall summary" line right at the end of the output which gives total counts of passed, failed and skipped tests

stevewardle:
@matthewrmshin I think this is ready for review - can you nominate someone?  Also should I rebase + squash _before_  or _after_ the review?

matthewrmshin:
@stevewardle Best to rebase and squash before review, to ensure that change being reviewed is as close to HEAD as possible.

lib/python/rose/apps/ana_builtin/grepper.py
```diff
@@ -41,20 +45,34 @@ def run_command_and_check(self):
...
-            self.parent.reporter(stdout, prefix="[INFO] ")
+            self.reporter(stdout, prefix="[INFO] ")
             self.passed = True
         else:
-            self.parent.reporter(stderr, prefix="[FAIL] ")
+            self.reporter(stderr, prefix="[FAIL] ")
+
+    def check_for_skip(self):
+        # If the users config options specified that the task should be
+        # ignored if all of its files were missing, set skipped attribute here
```
matthewrmshin: Can probably turn comment into docstring?

lib/python/rose/apps/ana_builtin/grepper.py
```diff
@@ -41,20 +45,34 @@ def run_command_and_check(self):
...
+        if self.skip_if_missing and len(self.files) > 0:
```
matthewrmshin: Can just do:

```python
if self.skip_if_missing and self.files:
```

lib/python/rose/apps/ana_builtin/grepper.py
```diff
@@ -41,20 +45,34 @@ def run_command_and_check(self):
...
+            if all([not os.path.exists(fname) for fname in self.files]):
```
matthewrmshin: My instinct tells me that the following may be faster - but only if Python is clever enough to not hit the file system any more if *any* file exists.

```python
if not any([os.path.exists(fname) for fname in self.files]):
```

stevewardle: I'm fairly sure Python _isn't_ clever enough for that - the comprehension will get evaluated in full and then the `any` will only scan the result as far as it needs to find a `True`...

However I think I'll still make this change because your version reads better

lib/python/rose/apps/ana_builtin/grepper.py
stevewardle: I usually go with the call to `len` for readability and because of the effective in-built type checking that enforces... if `self.files` were ever accidentally _not_ a list/iterable then my version will raise a  `TypeError: object of type X has no len()`, whereas your version will continue (with the status of the check varying depending on what type `self.files` happens to be)...

Although I concede that in this instance we are in direct control of the contents of `self.files` so perhaps the above is a moot point

lib/python/rose/apps/ana_builtin/grepper.py
stevewardle: Can do

doc/rose-rug-stem.html
```diff
@@ -358,6 +358,37 @@ <h3 class="alwayshidden">rose ana II</h3>
...
+              this:</p>
+
+              <pre class="prettyprint lang-rose_conf" >
+  [ana:config]
+  grepper-report-limit=5
+  skip-if-all-files-missing=.true.
+</pre>
+             <p>Each of these modifies the behaviour of <code>grepper</code>. 
+             The first option suppresses printed output for each analysis task
+             one the specified number of lines have been printed (in this case
```
oliver-sanders: *once?

doc/rose-configuration.html
```diff
@@ -1239,6 +1239,16 @@ <h2 id="appendix-rose-ana-config">Appendix: rose-ana
                 <code>rose task-run</code>
               </li>
             </ul>
+
+            <p>The app may also define a configuration section; 
+              <code>[ana:config]</code>, whose key-value pairs define app-wide
+              settings that are passed through to the analysis classes.  In
```
oliver-sanders: Double spaces achieve nothing in html. I think in the past they could cause problems with strange symbols appearing but it seems to render in Firefox fine.

doc/rose-rug-stem.html
```diff
@@ -459,10 +490,33 @@ <h3 class="alwayshidden">rose ana III</h3>
...
+                  <code>.true.</code> or <code>.false.</code>; if active, any
+                  comparison done on files by <code>grepper</code> will be 
+                  skipped if <i>all</i> of those files are non-existent.  In 
+                  this case the task will return as "skipped" rather than 
+                  passed/failed.</li>
+              </ul>
             </div>
 
             <div class="slide">
-              <h3 class="alwayshidden">rose ana IV</h3>
```
oliver-sanders: Slides "rose ana III, IV and V" do not fit into the slide viewer and should be split.

lib/python/rose/apps/rose_ana.py
```diff
@@ -196,11 +197,15 @@ def run(self, app_runner, conf_tree, opts, args, uuid, work_files):
...
+        # site defaults or the user's personal config
+        self._get_global_ana_config()
+
         # If the user's config indicates that it should be used - attach
         # to the KGO database instance in case it is needed later.
-        use_kgo = self.rose_conf.get_value(["rose-ana", "kgo-database"])
+        use_kgo = self.ana_config.get("kgo-database", ".false.")
         self.kgo_db = None
-        if use_kgo is not None and use_kgo == ".true.":
+        if use_kgo == ".true.":
```
oliver-sanders: `.true.` and `.false.` could be constants.

lib/python/rose/apps/ana_builtin/grepper.py
oliver-sanders: If you use a generator then it will only be evaluated as far as is necessary.
http://www.jjinux.com/2008/06/python-any-generator-expression.html

doc/rose-rug-stem.html
stevewardle: I didn't even think of testing that - thanks, I'll re-arrange it

doc/rose-configuration.html
stevewardle: Just my regular typing instincts at work - I'll get rid of it for consistency

lib/python/rose/apps/rose_ana.py
stevewardle: Good spot, I'll borrow from there (don't need `.false.` though as I don't think I ever reference it)

lib/python/rose/apps/ana_builtin/grepper.py
stevewardle: He's right!  As can be seen:

```python
if any([time.sleep(5) is None for _ in range(1000)]):
    # Runs for 5,000 seconds
if any(time.sleep(5) is None for _ in range(1000)):
    # Exits after 5 seconds
```

I'll still update it to Matt's suggested arrangement as well

doc/rose-rug-stem.html
stevewardle: In the process of doing this I realised that the documentation had drifted slightly away from the content, so I've improved it a bit further at the same time as tidying

stevewardle:
I think that's everything... do I dismiss the review @oliver-sanders ?

PR #2067, opened 2017-04-11
Rose mpi launch

steoxley:
@matthewrmshin I have tested this in my own space and with:

`rose test-battery t/rose-mpi-launch`

and also demonstrated that without the fix the modified test fails. I decided to use file size (`ulimit -f`) rather than stack size as that is the default value that `-H` or `-S` reports. In practice `-a` is the option we would normally use, but that produces many lines of output and would have been harder to create a control.

### DIFF
--- a/bin/rose-mpi-launch
+++ b/bin/rose-mpi-launch
@@ -131,7 +131,7 @@ while (($# > 0)); do
                       err "ROSE_LAUNCHER_ULIMIT_OPTS=$ROSE_LAUNCHER_ULIMIT_OPTS"
                       :;;
                     *)
-                      ulimit -$OPT $OPTARG || err \
+                      ulimit -$OPT ${OPTARG:-} || err \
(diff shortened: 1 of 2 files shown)


scwhitehouse:
I'm unsure what the motivation for this change is, can someone explain why the change is necessary?

steoxley:
Apologies, most of the initial work and reasoning was communicated by email. The particular use case for this option, as described in the header of the script, is:

```bash
ROSE_LAUNCHER_ULIMIT_OPTS="-a -s unlimited -d unlimited -a"
```
This is meant to run the following commands:

```bash
ulimit -a
ulimit -s unlimited
ulimit -d unlimited
ulimit -a
```

i.e. show the current settings, change various settings, show the new settings.

Unfortunately the code as it stands results in:

```bash
ulimit -a $OPTARG
```
for ulimit arguments without a value (`HSa`). Since `$OPTARG` is not defined in these cases, the `set -u` in `lib/bash/rose_init` brings down the script. I believe `set -u` is a relatively new addition to this script and why this problem has only recently been identified when a user tried to use this capability.

PR #2070, opened 2017-04-28
rose suite-run: no-remote-login-shell

matthewrmshin:
Support running `rose suite-run --remote=...` without invoking the login shell.

### DIFF
--- a/lib/python/rose/suite_run.py
+++ b/lib/python/rose/suite_run.py
@@ -303,58 +303,64 @@ class SuiteRunner(Runner):
         # Install items to user@host
         conf = ResourceLocator.default().get_conf()
         auths = self.suite_engine_proc.get_tasks_auths(suite_name)
-        queue = []  # [[pipe, command, "ssh"|"rsync", auth], ...]
+        proc_queue = []  # [[proc, command, "ssh"|"rsync", auth], ...]
         for auth in sorted(auths):
             host = auth
             if "@" in auth:
                 host = auth.split("@", 1)[1]
-            command = self.popen.get_cmd("ssh", auth, "bash", "--login", "-c")
-            rose_bin = "rose"
-            for name in [host, "*"]:
-                rose_home_node = conf.get(["rose-home-at", name],
-                                          no_ignore=True)
-                if rose_home_node is not None:
-                    rose_bin = "%s/bin/rose" % (rose_home_node.value)
-                    break
+            # Remote shell
+            command = self.popen.get_cmd("ssh", "-n", auth)
+            # Provide ROSE_VERSION and CYLC_VERSION in the environment
+            shcommand = "env ROSE_VERSION=%s %s=%s" % (
+                my_rose_version, suite_engine_key, suite_engine_version)
+            # Use login shell?
+            no_login_hell = self._run_conf(
+                "remote-no-login-shell", host=host, conf_tree=conf_tree)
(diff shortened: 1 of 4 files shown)


lib/python/rose/suite_run.py
```diff
@@ -303,58 +303,64 @@ def run_impl(self, opts, args, uuid, work_files):
...
+            # Remote shell
+            command = self.popen.get_cmd("ssh", "-n", auth)
+            # Provide ROSE_VERSION and CYLC_VERSION in the environment
+            shcommand = "env ROSE_VERSION=%s %s=%s" % (
+                my_rose_version, suite_engine_key, suite_engine_version)
+            # Use login shell?
+            no_login_hell = self._run_conf(
+                "remote-no-login-shell", host=host, conf_tree=conf_tree)
+            if not no_login_hell or no_login_hell.lower() != "true":
+                shcommand += r""" bash -l -c '"$0" "$@"'"""
```
dvalters: `no_login_hell` - Enjoyed this variable name.

dvalters:
What would be a usage example of this (just for testing purposes, say)? 

If I run it with some dummy argument to `--remote=SOME_VAR` I get a strange error message:

`[FAIL] 'NoneType' object has no attribute 'startswith'`

matthewrmshin:
The `--remote=` option takes a whole list of arguments. It is an internal option, always generated and called by `rose suite-run` itself. It is not intended for users. Our test battery has a reasonable coverage of  its functionality, so it should not be too bad. (In addition, it will not be long before we migrate this functionality to cylc, so I don't think we need to worry about this too much.)

hjoliver:
@matthewrmshin - thanks for this.  Next question (sorry!) - can `rose suite-clean` get the same treatment?

matthewrmshin:
> can `rose suite-clean` get the same treatment?

I'll have a think. Unlike `rose suite-run`, `rose suite-clean` does not currently run itself on remote hosts. Do we need to do `rose_prune` and `rose host-select` as well?

hjoliver:
> Unlike rose suite-run, `rose suite-clean` does not currently run itself on remote hosts. 

I'm not sure what your point is here?  My point is, the problem (for some sites) is it invoked `bash --login` on job hosts - which is not popular with security people who want to use ssh whitelisting to control access.

> Do we need to do `rose_prune` and `rose host-select` as well?

We run `rose_prune` on the job host (if it is pruning files there) - does it have the ability to prune files on remote hosts (via a bash login shell)?  We don't use `rose host-select` here so I'm not familiar with how it worsk, but generally BoM needs to know about (and hopefully avoid) any use of `ssh JOBHOST bash --login ...` by Cylc and Rose commands.  I think the only reason to use a login shell is to increase the odds that invoked commands are in `$PATH`, no?  In which case we don't need it where said commands are in the default `$PATH` for non-interactive shells.

matthewrmshin:
There are 3 reasons, as far as rose commands are concerned, for running `bash --login` on SSH.
1. Ensure correct PATH, and other PATH related variables, e.g. PYTHONPATH.
2. Ensure environment variables are set, e.g. SCRATCH, TMPDIR, DATADIR, etc.
3. Ensure that we are able to use bash syntax.

Yes, `rose_prune` can prune remote files and directories as well.

hjoliver:
Taking this offline...

PR #2073, opened 2017-05-17
Allow `rose_ana` grepper classes to treat missing KGO files as failed tasks

stevewardle:
Currently any of the file arguments being missing will result in the task raising an exception which (although this is caught correctly) means no entry is written to the KGO database...

It has emerged that it would actually be quite useful for our KGO update script to be able to effectively "install" new KGO files in the case that they are missing, so a simple change to treat the task as simply "failed" if the KGO file is missing has been suggested...


stevewardle:
I've run this using a copy of the UM trunk which uses the grepper classes to be sure it does what is intended (I'm going to need to open a separate ticket on the UM side to apply the same functionality to our own `mule_cumf` and `um_stdout` `rose_ana` classes though)

PR #2083, opened 2017-06-26
Fix tests broken by cylc/cylc#2314

matthewrmshin:
Fix tests broken by cylc/cylc#2314.


matthewrmshin:
(Note: test battery is currently broken on master.)

PR #2087, opened 2017-07-04
Use any generator

matthewrmshin:


dvalters:
Looks ok - this is to reduce memory overhead presumably? Is there any way to profile this in rose?

oliver-sanders:
> Looks ok so far - this is to reduce memory overhead presumably? Is there any way to profile this in rose?

Using generators also reduces CPU in the fail case.

We don't as yet have any sort of `rose profile-battery` command but we can always profile on the command line.

PR #2094, opened 2017-07-17
rose macro: improve feedback for invalid input

oliver-sanders:
When `rose macro` prompts users to provide values the resulting input is passes through `ast.literal_eval`. If the user provides an unquoted value they see an "Invalid entry..." notice, that is, unless the value uses any python characters, in which case they will see traceback. File system paths are a good example of this, `filename` is fine whereas `../filename` causes traceback.


matthewrmshin:
@dvalters please sanity check.

PR #2100, opened 2017-08-07
Add PyOpenSSL to Travis CI install

matthewrmshin:
In response to cylc/cylc#2355, we need to have PyOpenSSL for the Python environment in Travis CI.


matthewrmshin:
@oliver-sanders I think we now need this to get the test to pass on master.

PR #2104, opened 2017-08-10
Upgrade to isodatetime-2017.08.0

matthewrmshin:
See https://github.com/metomi/isodatetime/releases/tag/2017.08.0


matthewrmshin:
See also cylc/cylc#2389.

PR #2109, opened 2017-08-31
Remove port file related logic

matthewrmshin:
Trust the cylc 7 contact file as an indicator that a suite is running.


lib/python/rose/suite_engine_proc.py
```diff
@@ -204,7 +204,7 @@ class SuiteStillRunningError(Exception):
     FMT_HEAD = "Suite \"%(suite_name)s\" has running processes on:"
     FMT_BODY_1 = " %(host_name)s"
     FMT_BODY_M = "\n    %(host_name)s"
-    FMT_TAIL = "\nTry \"rose suite-shutdown --name=%(suite_name)s\" first?"
+    FMT_TAIL = "\nTry \"cylc stop '%(suite_name)s'\" first?"
```
dvalters: Is there a reason for switching this? i.e. Rose to recommend running a Cylc command (I mean I guess in practice people almost always have Rose + Cylc), just wondered.

matthewrmshin: `rose suite-shutdown` was an essential command for picking the correct suite host for the `cylc stop` command. It became unnecessary since cylc/cylc#1705, and I would like to encourage users to call `cylc stop` directly to reduce overhead.

dvalters:
@matthewrmshin also `/t/rosie-ls/01-many.t`?

matthewrmshin:
The tests are both OK in my environment, (but I guess it is worth checking what is causing the failure in your environment). Neither tests should be related to this change.

PR #2110, opened 2017-09-05
bunch: fix undefined variable

oliver-sanders:
Also minor style changes.

### DIFF
--- a/lib/python/rose/apps/rose_bunch.py
+++ b/lib/python/rose/apps/rose_bunch.py
@@ -146,7 +146,7 @@ class RoseBunchApp(BuiltinApp):
 
         if self.fail_mode not in self.FAIL_MODE_TYPES:
             raise ConfigValueError([self.BUNCH_SECTION, "fail-mode"],
-                                   fail_mode,
+                                   self.fail_mode,
                                    "not a valid setting")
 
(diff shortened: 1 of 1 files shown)


lib/python/rose/apps/rose_bunch.py
```diff
@@ -440,12 +432,12 @@ def check_has_succeeded(self, name):
         s_stmt = ("SELECT * FROM " + self.TABLE_COMMANDS +
                   " WHERE name==? AND status==?")
         s_stmt_args = [name, self.S_PASS]
-        has_succeeded = False
-        for s_row in self.conn.execute(s_stmt, s_stmt_args):
-            has_succeeded = True
-        return has_succeeded
+        if self.conn.execute(s_stmt, s_stmt_args):
+            return True
```
matthewrmshin: The problem is that `self.conn.execute(...)` will return a `Cursor` object even if no rows are selected.

So perhaps we can do:

```python
for _ in self.conn.execute(s_stmt, s_stmt_args):
    return True
return False
```

Or:

```python
return bool(self.conn.execute(s_stmt, s_stmt_args).fetchone())
```

May be worth adding a `LIMIT 1` clause to the SELECT statement as well?

PR #2112, opened 2017-09-13
Fix get_util_name exception.

oliver-sanders:
`os.getenv` will return `None` rather than raising `KeyError` so the `try`/`except` was failing to catch the error.

### DIFF
--- a/lib/python/rose/resource.py
+++ b/lib/python/rose/resource.py
@@ -140,9 +140,9 @@ class ResourceLocator(object):
         util = self.util
         try:
             if namespace is None:
-                namespace = os.getenv("ROSE_NS")
+                namespace = os.environ["ROSE_NS"]
             if util is None:
-                util = os.getenv("ROSE_UTIL")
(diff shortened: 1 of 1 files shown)


lib/python/rose/resource.py
```diff
@@ -140,9 +140,9 @@ def get_util_name(self, separator=" "):
         util = self.util
         try:
             if namespace is None:
-                namespace = os.getenv("ROSE_NS")
+                namespace = os.environ("ROSE_NS")
             if util is None:
-                util = os.getenv("ROSE_UTIL")
+                util = os.environ("ROSE_UTIL")
```
matthewrmshin: `os.getenv` is a function but `os.environ` is a dict, so need to update `(...)` to `[...]`.

oliver-sanders: Derp!

matthewrmshin:
@dvalters please sanity check.

PR #2114, opened 2017-09-18
Push import errors to task level in ``rose_ana``

### LINKED ISSUE #2113: Make rose_ana slightly more tolerant of analysis module import failures
Currently a user-supplied "analysis module" for ``rose_ana`` failing to import will cause the entire ``rose_ana`` task to abort.  It would be good if it was more tolerant of this situation...

At a glance I think the solution is fairly simple - the ``except`` in ``_load_analysis_modules`` needs to not stop execution, then the final part of ``_load_tasks`` should not raise an exception and instead append some sort of dummy object with the ``options`` as an attribute (for retrieval of the task name)... the lack of a ``run_analysis`` method in this dummy object should then cause the individual comparison to fail in the ``try|except`` of the main ``run`` method.

stevewardle:
Closes #2113

### DIFF
--- a/lib/python/rose/apps/rose_ana.py
+++ b/lib/python/rose/apps/rose_ana.py
@@ -151,6 +151,64 @@ class KGODatabase(object):
         self.statement_buffer = []
 
 
+class AnalysisTask(object):
+    """
+    Base class for an analysis task; all custom user tasks should inherit
+    from this class and override the "run_analysis" method to perform
+    whatever analysis is required.
+
+    """
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, parent_app, task_options):
+        """
+        Initialise the analysis task, storing the user specified options
+        dictionary and a few references to useful objects from the parent app.
+
+        """
+        self.options = task_options
+
+        # This attribute gives access to the parent task; but it is only
+        # included for backwards compatibility and will be remove in the
+        # future (please instead use the other attributes below!)
+        self.parent = parent_app
+
+        # Attributes to access some helpful/relevant parts of the parent
+        # task environment for printing, running tasks, etc.
+        self.config = parent_app.ana_config
+        self.reporter = parent_app.reporter
+        self.kgo_db = parent_app.kgo_db
+        self.popen = parent_app.app_runner.popen
+
+        self.passed = False
+        self.skipped = False
+
+    @abc.abstractmethod
+    def run_analysis(self):
+        """
+        Will be called to start the analysis code; this method should be
+        overidden by the user's class to perform the desired analysis.
+
+        """
+        msg = "Abstract analysis task class should never be called directly"
+        raise ValueError(msg)
+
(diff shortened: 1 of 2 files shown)


lib/python/rose/apps/rose_ana.py
```diff
@@ -360,6 +418,9 @@ def _load_analysis_modules(self):
                 try:
                     self.modules.add(__import__(module_name))
                 except Exception as err:
+                    # Note: We intentionally don't re-raise the exception
+                    # here, as we want to avoid a single mistake in a user
+                    # supplied method bringing down the entire task
```
matthewrmshin: Should this:
* catch only `ImportError`?
* report a warning instead of a fail?

lib/python/rose/apps/rose_ana.py
```diff
@@ -151,6 +151,64 @@ def buffer_to_db(self, reporter=None):
...
+        """
+        Options should be removed from the options dictionary as they are
+        processed; this method may then be called to catch and unknown options
+
+        """
+        unhandled = []
+        for option in self.options:
+            if option not in ["full_task_name", "description"]:
+                unhandled.append(option)
+        if len(unhandled) > 0:
```
matthewrmshin: `if unhandled:`?

stevewardle: Sure

lib/python/rose/apps/rose_ana.py
stevewardle: Hmm... I suppose I could use ``ImportError`` although I figured since using ``__import__`` felt a little bit non-standard it would be safer to catch anything - do you think it can be relied upon to ``raise`` the right exception type?

I can change the prefix to ``[WARN]`` but should that be ``KIND_ERR`` or ``KIND_OUT`` (as there isn't a ``KIND_WARN``)?

matthewrmshin: `import` uses `__import__` internally, according to the documentation of the Python's built-in functions, so it should be sufficient to catch only `ImportError`. You may want to catch `TypeError` as well (to catch non-string argument) or you can explicitly convert the argument to a `str` before passing it into the function.

Yes, you can easily change that by switching to `KIND_ERR` which will use the warning prefix by default.

lib/python/rose/apps/rose_ana.py
```diff
@@ -359,13 +417,16 @@ def _load_analysis_modules(self):
                     os.path.basename(filename))[0]
                 try:
                     self.modules.add(__import__(module_name))
-                except Exception as err:
+                except ImportError as err:
```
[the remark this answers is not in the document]
stevewardle: Are you serious @codacy-bot?  Why didn't you bring this up before?!

matthewrmshin:
Doubly approved. Normally enough for merge. However, since you have assigned @steoxley as reviewer as well, I'll let him do the honour.

PR #2119, opened 2017-10-02
Doc: remove dead cylon image link

matthewrmshin:
Remove a dead link that is causing the test battery to fail.


matthewrmshin:
(1 review is enough.)

PR #2123, opened 2017-10-20
Replace simplejson with json in standard library

matthewrmshin:
They are the same module from Python 2.6+. This change removes this unnecessary external dependency.

### DIFF
--- a/doc/rose-install.html
+++ b/doc/rose-install.html
@@ -167,10 +167,10 @@
                   <p><dfn>used by:</dfn> Rosie web service
                   client.</p>
 
-                  <p><dfn>tested with version:</dfn> 2.2.1.</p>
+                  <p><dfn>tested with version:</dfn> 2.4.2.</p>
 
                   <p><dfn>minimum version required:</dfn>
-                  2.2.1.</p>
+                  2.4.2.</p>
 
                   <p><dfn>remark:</dfn> Development of this module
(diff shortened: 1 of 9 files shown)


t/rose-suite-restart/02-basic.t
```diff
@@ -20,7 +20,7 @@
 # Test "rose suite-restart", basic usage.
 #-------------------------------------------------------------------------------
 . "$(dirname "$0")/test_header"
-tests 4
+tests 3
```
matthewrmshin: Fixes test failure caused by cylc/cylc#2437.

doc/rose-install.html
```diff
@@ -167,10 +167,10 @@ <h2 id="requirements">System Requirements</h2>
                   <p><dfn>used by:</dfn> Rosie web service
                   client.</p>
 
-                  <p><dfn>tested with version:</dfn> 2.2.1.</p>
+                  <p><dfn>tested with version:</dfn> 2.4.2.</p>
 
                   <p><dfn>minimum version required:</dfn>
-                  2.2.1.</p>
+                  2.4.2.</p>
```
matthewrmshin: This is to make it consistent with the Cylc HTTPS client.

matthewrmshin:
@dvalters please do a quick sanity check.

PR #2124, opened 2017-10-23
Fix selection expanded after change.

oliver-sanders:
The expanded selection was incorrectly updated in the event that the `TreeStore` was modified in a manner which caused a re-ordering of rows which in-turn affected a re-ordering of sections. This is a presentation layer bug as the underlying data-structure is not affected.

#### The steps to reproduce are somewhat convoluted:

1. Open a widget which inherits from the `BaseSummaryDataPanel` class.
    - Note: The data-structure will require at least 4 columns.
    - Note: At least three sections must be present.
2. Group by one column.
3. Expand a section within it ensuring that:
    - The section has 3 or more items.
    - In the section, at least two columns where at least two values are present.
    - The section is not the first in the panel.
4. Sort by **an even number** of columns (at least one of which must contain multiple values).
6. Change a value in one of the remaining columns (multiple values must be present) to the lowest possible value (this way it is certain to change the sorting).

### DIFF
--- a/lib/python/rose/config_editor/panelwidget/summary_data.py
+++ b/lib/python/rose/config_editor/panelwidget/summary_data.py
@@ -248,8 +248,14 @@ class BaseSummaryDataPanel(gtk.VBox):
         if variables is not None:
             self.variables = variables
         old_cols = set(self.column_names)
-        expanded_rows = []
-        self._view.map_expanded_rows(lambda r, d: expanded_rows.append(d))
+
+        # Generate a set of expanded sections (one level only).
+        expanded_sections = set()
+        model = self._view.get_model()
+        self._view.map_expanded_rows(
+            lambda r, p:
+            expanded_sections.add(model.get_value(model.get_iter(p), 0)))
+
         start_path, start_column = self._view.get_cursor()
         should_redraw = self.update_tree_model()
         if should_redraw:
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Test failure fixed by #2123.

oliver-sanders:
@arjclark Can you confirm this resolves the issue you reported?

arjclark:
@oliver-sanders - tested and looks like this keeps the correct group open.

Although the ordering of the groups themselves changes (so if you have two similar groups open you may miss the redraw), this change does address the immediate gotcha of ungrouping things that weren't there previously. I'm not sure how hard/feasible it would be to maintain the ordering of the groupings themselves though?

oliver-sanders:
> I'm not sure how hard/feasible it would be to maintain the ordering of the groupings themselves though?

Short Answer:

Unfortunately sorting is handled internally within GTK. If the re-ordering is an issue the best you can do is to sort by the top level field (in the reported case this would be "packages").

Long Answer:

I had a go at writing a manual iterator to work through the GTK data-structure (gtk.TreeStore via gtk.TreeModel) - one could then manually swap items as desired to effectively define an extra sorting layer. Unfortunately the sorting is handled internally in GTK creating a barrier between the stored and represented data meaning that this approach won't work.

arjclark:
@oliver-sanders - thanks for the clarification. GTK vs. TreeViews/Stores once again proves a nuisance 🙁 

Am happy for this to be the fix to close the reported issue.

oliver-sanders:
I'll wait for #2123 before merging this.

oliver-sanders:
`./t/syntax/00-lib.t` is fixed by #2128

PR #2128, opened 2017-10-25
test-battery: pycodestyle

oliver-sanders:
The python `pep8` package has been renamed `pycodestyle` to avoid confusion.

This fixes `t/syntax/00-lib.t` which will fail with newer versions of `pep8` due to a depreciation notice.


matthewrmshin:
(Would this work on our site?)

oliver-sanders:
No, hence I have left the `pep8` test in for compatibility.

lib/python/rosie/ws.py
```diff
@@ -109,7 +109,7 @@ def index(self, *_):
         """Provide the index page."""
         try:
             return self._render()
-        except:
+        except (KeyError, AttributeError, jinja2.exceptions.TemplateError):
```
oliver-sanders: `jinja2.exceptions.TemplateError` might be sufficient here?

lib/python/rose/run.py
```diff
@@ -202,8 +202,8 @@ def run(self, opts, args):
         finally:
             # Close handle on specific log file
             try:
-                self.event_handler.contexts.get(uuid).handle.close()
-            except:
+                self.event_handler.contexts[uuid].handle.close()
+            except (KeyError, IOError, AttributeError):
```
oliver-sanders: possibly `TypeError` as well?

lib/python/rosie/ws.py
matthewrmshin: Most likely. The client will get a 500 in the worst case.

lib/python/rose/run.py
matthewrmshin: I can't see how.

matthewrmshin:
@stevewardle do you mind doing a sanity check for us?

stevewardle:
Looks perfectly reasonable to me

PR #2132, opened 2017-12-01
Pylint fixes

matthewrmshin:
Unused imports and variables, `Popen` with `stdin` pointing to `/dev/null` by default and other stuffs

### DIFF
--- a/lib/python/rose/app_run.py
+++ b/lib/python/rose/app_run.py
@@ -448,7 +448,9 @@ class AppRunner(Runner):
         self.handle_event("command: %s" % command)
         if opts.install_only_mode:
             return
-        self.popen(command, shell=True, stdout=sys.stdout, stderr=sys.stderr)
+        self.popen(
+            command, shell=True, stdout=sys.stdout, stderr=sys.stderr,
+            stdin=sys.stdin)
 
 
 def main():
@@ -466,7 +468,7 @@ def main():
         if opts.debug_mode:
             traceback.print_exc(exc)
         if isinstance(exc, RosePopenError):
-            sys.exit(exc.rc)
+            sys.exit(exc.ret_code)
         else:
             sys.exit(1)
 
--- a/lib/python/rose/config_editor/panelwidget/summary_data.py
+++ b/lib/python/rose/config_editor/panelwidget/summary_data.py
@@ -174,10 +174,8 @@ class BaseSummaryDataPanel(gtk.VBox):
 
     def update_tree_model(self):
         """Construct a data model of other page data."""
-        sub_sect_names = self.sections.keys()
-        sub_var_names = []
         self.var_id_map = {}
-        for section, variables in self.variables.items():
+        for variables in self.variables.values():
             for variable in variables:
                 self.var_id_map[variable.metadata["id"]] = variable
         data_rows, column_names = self.get_model_data()
@@ -238,7 +236,7 @@ class BaseSummaryDataPanel(gtk.VBox):
 
     def set_focus_node_id(self, node_id):
         """Set the focus on a particular node id, if possible."""
-        section, option = self.util.get_section_option_from_id(node_id)
+        section = self.util.get_section_option_from_id(node_id)[0]
         self.scroll_to_section(section)
 
     def update(self, sections=None, variables=None):
@@ -256,7 +254,6 @@ class BaseSummaryDataPanel(gtk.VBox):
             lambda r, p:
             expanded_sections.add(model.get_value(model.get_iter(p), 0)))
 
-        start_path, start_column = self._view.get_cursor()
         should_redraw = self.update_tree_model()
         if should_redraw:
             self.add_new_columns(self._view, self.column_names)
@@ -382,9 +379,9 @@ class BaseSummaryDataPanel(gtk.VBox):
         pathinfo = treeview.get_path_at_pos(int(event.x),
                                             int(event.y))
         if pathinfo is not None:
-            path, col, cell_x, cell_y = pathinfo
+            path, col = pathinfo[0:2]
             if event.button == 3:
-                model, rows = self._view.get_selection().get_selected_rows()
+                rows = self._view.get_selection().get_selected_rows()[1]
                 if len(rows) > 1:
                     # Multiple selection.
                     self._popup_tree_multi_menu(event)
--- a/lib/python/rose/config_editor/panelwidget/summary_data.py
+++ b/lib/python/rose/config_editor/panelwidget/summary_data.py
@@ -72,7 +72,7 @@ class BaseSummaryDataPanel(gtk.VBox):
             multiple_selection=True)
         self._view.set_rules_hint(True)
         self.sort_util = rose.gtk.util.TreeModelSortUtil(
-            lambda: self._view.get_model(), multi_sort_num=2)
+            self._view.get_model, multi_sort_num=2)
         self._view.show()
         self._view.connect("button-press-event",
                            self._handle_button_press_event)
@@ -708,9 +708,7 @@ class BaseSummaryDataPanel(gtk.VBox):
 
     def _sort_row_data(self, row1, row2, sort_index, descending=False):
         fac = (-1 if descending else 1)
-        x = row1[sort_index]
-        y = row2[sort_index]
-        return fac * self.sort_util.cmp_(x, y)
+        return fac * self.sort_util.cmp_(row1[sort_index], row2[sort_index])
 
     def _handle_group_change(self, combobox):
         model = combobox.get_model()
--- a/lib/python/rose/config_editor/plugin/um/widget/stash_add.py
+++ b/lib/python/rose/config_editor/plugin/um/widget/stash_add.py
@@ -114,8 +114,7 @@ class AddStashDiagnosticsPanelv1(gtk.VBox):
         self._view.show()
         self._view.connect("button-press-event",
                            self._handle_button_press_event)
-        self._view.connect("cursor-changed",
-                           lambda v: self._update_control_sensitivity())
+        self._view.connect("cursor-changed", self._update_control_sensitivity)
         self._window = gtk.ScrolledWindow()
         self._window.set_policy(gtk.POLICY_AUTOMATIC, gtk.POLICY_AUTOMATIC)
         self.generate_tree_view(is_startup=True)
@@ -689,7 +688,7 @@ class AddStashDiagnosticsPanelv1(gtk.VBox):
         self._should_show_meta_column_titles = widget.get_active()
         self.generate_tree_view()
 
-    def _update_control_sensitivity(self):
+    def _update_control_sensitivity(self, _=None):
         section, item = self._get_current_section_item()
         self._add_button.set_sensitive(section is not None and
                                        item is not None)
--- a/lib/python/rose/config_editor/plugin/um/widget/stash_add.py
+++ b/lib/python/rose/config_editor/plugin/um/widget/stash_add.py
@@ -175,7 +175,6 @@ class AddStashDiagnosticsPanelv1(gtk.VBox):
         columns = ["Section", "Item", "Description", "?", "#"]
         sections = self.stash_lookup.keys()
         sections.sort(self.sort_util.cmp_)
-        mod_markup = rose.config_editor.SUMMARY_DATA_PANEL_MODIFIED_MARKUP
         props_excess = [self.STASH_PARSE_DESC_OPT, self.STASH_PARSE_ITEM_OPT,
                         self.STASH_PARSE_SECT_OPT]
         for section in sections:
@@ -206,7 +205,6 @@ class AddStashDiagnosticsPanelv1(gtk.VBox):
         else:
             col_types = []
         self._store = gtk.TreeStore(*col_types)
-        parent_iter_ = None
         for i, row_data in enumerate(data_rows):
             if rows_are_descendants is None:
                 self._store.append(None, row_data)
@@ -306,7 +304,6 @@ class AddStashDiagnosticsPanelv1(gtk.VBox):
         item_col_index = self.column_names.index("Item")
         streq_info_index = self.column_names.index("?")
         num_streqs_index = self.column_names.index("#")
-        parent_iter_stack = []
         # For speed, pass in the relevant indices here.
         user_data = (sect_col_index, item_col_index,
                      streq_info_index, num_streqs_index)
@@ -453,16 +450,15 @@ class AddStashDiagnosticsPanelv1(gtk.VBox):
         return filter_hbox
 
     def _get_current_section_item(self):
-        # Return the current highlighted section (or None) and item (or None).
-        current_path, current_column = self._view.get_cursor()
+        """Return the current highlighted section and item."""
+        current_path = self._view.get_cursor()[0]
         if current_path is None:
             return (None, None)
         current_iter = self._view.get_model().get_iter(current_path)
         return self._get_section_item_from_iter(current_iter)
 
     def _get_section_item_col_indices(self):
-        # Return the column indices of the STASH section and item.
-        model = self._view.get_model()
+        """Return the column indices of the STASH section and item."""
         sect_index = 0
         if self.group_index is not None and self.group_index != sect_index:
             sect_index = 1
@@ -477,7 +473,7 @@ class AddStashDiagnosticsPanelv1(gtk.VBox):
         return sect_index, item_index
 
     def _get_section_item_from_iter(self, iter_):
-        # Return the STASH section and item numbers for this row.
+        """Return the STASH section and item numbers for this row."""
         sect_index, item_index = self._get_section_item_col_indices()
         model = self._view.get_model()
         section = model.get_value(iter_, sect_index)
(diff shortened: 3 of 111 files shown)


matthewrmshin:
Something is very wrong with Codacy!

lib/python/rose/meta_type.py
```diff
@@ -185,8 +185,8 @@ def is_valid(self, value):
         quote_segs = value.split('"')
         if len(quote_segs) < 3 or quote_segs[0] or quote_segs[-1]:
             return [False, self.WARNING.format(repr(value))]
-        for i, s in enumerate(quote_segs):
-            num_end_esc = len(s) - len(s.rstrip("\\"))
+        for i, seg in enumerate(quote_segs):
```
oliver-sanders: Single character variable name.

lib/python/rose/popen.py
```diff
@@ -175,29 +175,31 @@ def run_bg(self, *args, **kwargs):
         stdin = kwargs.get("stdin")
         if isinstance(stdin, str):
             kwargs["stdin"] = PIPE
+        elif stdin is None:
+            kwargs["stdin"] = open(os.devnull)
```
oliver-sanders: There shouldn't be the risk of editors being called via RosePopen, is there any other foreseeable issue?

lib/python/rose/reporter.py
```diff
@@ -311,12 +308,14 @@ def write(self, message):
         self._send_pending_messages()
 
     def _send_pending_messages(self):
-        for message in list(self._messages_pending):
+        while self._messages_pending:
+            message = self._messages_pending[0]
```
oliver-sanders:
```python
message = self._messages_pending.pop()
```

lib/python/rose/task_env.py
```diff
@@ -81,16 +81,17 @@ def get_prepend_paths(event_handler=None, path_root=None, path_glob_args=[],
...
-        if path_glob_arg is None:
-            continue
-        if "=" in path_glob_arg:
-            name, value = path_glob_arg.split("=", 1)
-        else:
-            name, value = "PATH", path_glob_arg
-        if name not in path_globs_map:
-            path_globs_map[name] = []
-        path_globs_map[name].append(value)
+    if path_glob_args:
```
oliver-sanders: What is the function of this?

lib/python/rosie/browser/history.py
```diff
@@ -175,64 +167,38 @@ def __init__(self, hist_path=rosie.browser.HISTORY_LOCATION,
...
+                try:
+                    search_mode = ast.literal_eval(tail)
+                except (SyntaxError, ValueError):
+                    search_mode = False
+                h_type, details_str = head.split(',', 1)
+                history.append(HistoryItem(h_type, details_str, search_mode))
         else:
-            f = open(self.hist_path, 'w')
-            f.close()
+            open(self.hist_path, 'w').close()
```
oliver-sanders: What is the purpose of this?

lib/python/rosie/browser/util.py
```diff
@@ -149,8 +145,8 @@ def __init__(self, number=0, is_end=False, change_hook=None,
         self.show_controls = show_controls
         self.combo_box = gtk.combo_box_new_text()
         self.char = self.CHARS[is_end]
-        for n in range(max(number, self.MAX_NUM) + 1):
-            self.combo_box.append_text(self.char * n)
+        for i in range(max(number, self.MAX_NUM) + 1):
```
oliver-sanders: Single character variable name.

lib/python/rose/config_editor/panelwidget/summary_data.py
```diff
@@ -174,10 +174,8 @@ def _get_control_widget_hbox(self):
 
     def update_tree_model(self):
         """Construct a data model of other page data."""
-        sub_sect_names = self.sections.keys()
-        sub_var_names = []
         self.var_id_map = {}
-        for section, variables in self.variables.items():
+        for variables in self.variables.values():
```
sadielbartholomew: Whilst checking through the changes to this file I've spotted three other single-character names for statement variables between this and the next line change ('i' starting line 200, 'j' starting 203 and another 'i' starting 221) that should be changed to comply with pylint.

lib/python/rosie/browser/util.py
matthewrmshin: The character *i* is actually allowed as an index variable.

lib/python/rose/config_editor/valuewidget/character.py
```diff
@@ -18,13 +18,11 @@
...
 # -----------------------------------------------------------------------------
 
-import re
-
 import pygtk
 pygtk.require('2.0')
 import gtk
 
-import rose.config_editor
+import rose
```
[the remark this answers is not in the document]
matthewrmshin: No, it is actually used.

lib/python/rose/app_run.py
```diff
@@ -448,7 +448,9 @@ def _command(self, conf_tree, opts, args):
         self.handle_event("command: %s" % command)
         if opts.install_only_mode:
             return
-        self.popen(command, shell=True, stdout=sys.stdout, stderr=sys.stderr)
+        self.popen(
+            command, shell=True, stdout=sys.stdout, stderr=sys.stderr,
+            stdin=sys.stdin)
```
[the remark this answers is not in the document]
matthewrmshin: The wrapper has a slightly different interface.

lib/python/rosie/browser/history.py
matthewrmshin: It creates an empty file.

lib/python/rose/config_editor/plugin/um/widget/stash_add.py
```diff
@@ -110,19 +110,19 @@ def __init__(self, stash_lookup, request_lookup,
...
         self._view.set_rules_hint(True)
         self.sort_util = rose.gtk.util.TreeModelSortUtil(
-            lambda: self._view.get_model(), 2)
+            self._view.get_model, 2)
         self._view.show()
         self._view.connect("button-press-event",
                            self._handle_button_press_event)
         self._view.connect("cursor-changed",
-                           lambda v: self._update_control_widget_sensitivity())
+                           lambda v: self._update_control_sensitivity())
```
sadielbartholomew: 117-118 to just``self._view.connect("cursor-changed", self._update_control_sensitivity)`` instead?

lib/python/rose/config_editor/updater.py
```diff
@@ -359,7 +356,7 @@ def update_ignored_statuses(self, namespace):
             sect, opt = self.util.get_section_option_from_id(setting_id)
             if opt is None:
                 sect_vars = config_data.vars.now.get(sect, [])
-                ns = self.data.helper.get_default_namespace_for_section(
+                ns = self.data.helper.get_default_section_namespace(
```
sadielbartholomew: Rename 'ns' so it has 3 characters +.

lib/python/rose/config_editor/util.py
```diff
@@ -111,7 +105,6 @@ def __str__(self):
 def launch_node_info_dialog(node, changes, search_function):
     """Launch a dialog displaying attributes of a variable or section."""
     title = node.__class__.__name__ + " " + node.metadata['id']
-    safe_str = rose.gtk.util.safe_str
```
sadielbartholomew: Spotted another 'i' variable below this, starting on line 127.

lib/python/rose/config_editor/util.py
```diff
@@ -201,12 +194,12 @@ def wrap_string(text, maxlen=72, indent0=0, maxlines=4, sep=","):
...
-    x_sort_key = x[0]
-    x_id = x[1]
-    y_sort_key = y[0]
-    y_id = y[1]
+def null_cmp(x_item, y_item):
+    """Compares sort_key and then id of the tuples x_item/y_item."""
+    x_sort_key = x_item[0]
+    x_id = x_item[1]
+    y_sort_key = y_item[0]
+    y_id = y_item[1]
```
sadielbartholomew: What about ``x_sort_key, x_id = x_item[:2]`` (and then the y equivalent) to halve the line count for 199-202? Or is this considered bad style/practice?

lib/python/rose/config_editor/valuewidget/array/entry.py
```diff
@@ -132,8 +128,6 @@ def set_focus_index(self, focus_index=None):
         value_array = rose.variable.array_split(self.value)
         text = ''
         for i, val in enumerate(value_array):
```
sadielbartholomew: There's another variable named 'i' here.

lib/python/rose/app_run.py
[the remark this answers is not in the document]
matthewrmshin: The wrapper, see later, will redirect STDIN from `/dev/null` by default. This forces users to redirect from `sys.stdin` explicitly if required (like in this case), which is a safer usage.

lib/python/rose/gtk/dialog.py
```diff
@@ -253,25 +251,23 @@ def _process(cmd_args, stdout=sys.stdout, stderr=sys.stderr):
         func = cmd_args.pop(0)
         try:
             func(*cmd_args)
-        except Exception as e:
-            stderr.write(type(e).__name__ + ": " + str(e) + "\n")
-            text1 = stderr.read()
+        except Exception as exc:
+            stderr.write(type(exc).__name__ + ": " + str(exc) + "\n")
+            stderr.read()
```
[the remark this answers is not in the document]
matthewrmshin: No idea, but looks like the same magic as below.

lib/python/rose/reporter.py
matthewrmshin: The lead message only gets `del` on successful queuing.

lib/python/rose/task_env.py
matthewrmshin: The argument was set to the dangerous default `[]`. The default is now `None`, so need to use an `if` block to protect the loop.

lib/python/rose/config_editor/valuewidget/array/row.py
```diff
@@ -360,9 +357,9 @@ def _normalise_width_chars(self, widget):
             while child_list:
                 child = child_list.pop()
                 if isinstance(child, gtk.Entry) and hasattr(child, 'get_text'):
-                    w = len(child.get_text())
-                    if w > max_width.get(i, -1):
-                        max_width.update({i: w})
+                    width = len(child.get_text())
+                    if width > max_width.get(i, -1):
```
sadielbartholomew: Is 'i' okay here or does this also need renaming?

lib/python/rose/config_editor/valuewidget/format.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, value, metadata, set_value, hook, arg_str=None):
         self.entry_table = gtk.Table(rows=num_entries + 1, columns=1)
         self.entry_table.show()
         self.entries = []
-        for r, format_name in enumerate(value.split()):
+        for format_name in value.split():
```
sadielbartholomew: I can also see a lot of unnecessary lambda functions below this, on lines 64-69, which you might want to remove.

lib/python/rose/config_editor/variable.py
```diff
@@ -186,8 +180,8 @@ def generate_valuewidget(self, variable, override_custom=False,
                                           set_value,
                                           hook_object,
                                           custom_arg)
-            except Exception as e:
-                self.handle_bad_valuewidget(str(e), variable, set_value)
+            except Exception as exc:
```
sadielbartholomew: Can this type of error be given a specific exception class instead?

lib/python/rose/gtk/dialog.py
```diff
@@ -253,25 +251,23 @@ def _process(cmd_args, stdout=sys.stdout, stderr=sys.stderr):
...
-                         stderr=subprocess.PIPE)
-    for line in iter(p.stdout.readline, ""):
+    proc = Popen(cmd_args, stdout=PIPE, stderr=PIPE)
+    for line in iter(proc.stdout.readline, ""):
         stdout.write(line)
-    for line in iter(p.stderr.readline, ""):
+    for line in iter(proc.stderr.readline, ""):
         stderr.write(line)
-    p.wait()
-    text0 = stdout.read()  # Magically keep it alive.
```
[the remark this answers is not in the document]
sadielbartholomew: Yes I second the use of '!?', if code style tests were able to pass or fail comments for merit of content this one might not make the cut...

lib/python/rose/macros/__init__.py
```diff
@@ -42,8 +42,7 @@ class DefaultTransforms(rose.macro.MacroTransformerCollection):
     def __init__(self):
```
sadielbartholomew: Not related to the current PR changes but the 'inspect' module seems to be an unused import for this file unless it is utilised in a non-obvious way (not sure how to comment against the relevant line as I cannot seem to comment the usual way on a line in expanded view.)

lib/python/rose/meta_type.py
```diff
@@ -206,15 +206,15 @@ def transform(self, value):
...
 
 def meta_type_transform(value, meta_type):
-    c = MetaType.get_meta_type(meta_type)
-    c = c()
+    item = MetaType.get_meta_type(meta_type)
+    item = item()
     try:
-        return c.transform(value)
+        return item.transform(value)
     except Exception:
```
sadielbartholomew: Is a more specific exception class possible here - ``except NotImplementedError:`` maybe?

lib/python/rose/host_select.py
```diff
@@ -151,7 +151,7 @@ def get_local_host_strs(self):
                     continue
                 self.local_host_strs.append(item)
                 try:
-                    for addrinfo_item in socket.getaddrinfo(item, None):
+                    for addrinfo_item in getaddrinfo(item, None):
```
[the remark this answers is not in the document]
oliver-sanders: Now done.

lib/python/rosie/browser/history.py
oliver-sanders: I mistook the indent level!

lib/python/rosie/browser/util.py
oliver-sanders: That's awfully letterist!

lib/python/rose/config_editor/valuewidget/array/mixed.py
```diff
@@ -88,9 +87,8 @@ def __init__(self, value, metadata, set_value, hook, arg_str=None):
         self.entry_table.connect('focus-in-event',
                                  self.hook.trigger_scroll)
         self.entry_table.show()
-        r = 0
-        for r in range(self.num_rows):
-            self.insert_row(r)
+        for i in range(self.num_rows):
```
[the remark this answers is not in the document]
matthewrmshin: A loop is clearer here - and since we are not interested in the return value of `map`.

lib/python/rose/config_editor/variable.py
matthewrmshin: A bit more difficult, because `widget` can be anything.

lib/python/rose/meta_type.py
matthewrmshin: I think `item.transform` may do anything.

lib/python/rose/metadata_check.py
```diff
@@ -102,9 +102,8 @@ def _check_macro(value, module_files=None, meta_dir=None):
         return
     try:
         macros = rose.variable.array_split(value, only_this_delim=",")
-    except Exception as e:
-        return INVALID_SYNTAX.format(e)
-    bad_macros = []
+    except Exception as exc:
```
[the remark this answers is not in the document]
matthewrmshin: (You are right on this one - as the main failure is with a method that does a lot of regular expression matchings, but I think I'll give it a miss for now, in case the change breaks something else.)

lib/python/rose/task_env.py
```diff
@@ -127,31 +128,29 @@ def main():
...
-    except Exception as e:
-        report(e)
+            path = os.pathsep.join(prepend_paths + orig_paths)
+            report(str(EnvExportEvent(key, path)) + "\n", level=0)
+    except Exception as exc:
+        report(exc)
         if opts.debug_mode:
-            traceback.print_exc(e)
+            traceback.print_exc(exc)
         sys.exit(1)
```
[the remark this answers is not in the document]
matthewrmshin: Yes. But in due course.

lib/python/rose/gtk/dialog.py
```diff
@@ -262,7 +262,7 @@ def _process(cmd_args, stdout=sys.stdout, stderr=sys.stderr):
     for line in iter(proc.stderr.readline, ""):
         stderr.write(line)
     proc.wait()
-    stdout.read()  # Magically keep it alive.
+    stdout.read()  # Magically keep it alive!?
```
sadielbartholomew: Technically in feedback the intention was to convey that we should improve or remove this comment, but the humour is appreciated(!)

matthewrmshin: The problem is, I don't really know (and/or do not yet want to find out) what we are trying to magically keep alive. It probably should have a *FIXME* label.

The statement before this is `proc.wait`, so in theory, the child process is already finished, so are we trying to keep something else alive?

oliver-sanders:
Dismissing (erroneous) codacy review and merging as double approved.

lib/python/rose/gtk/dialog.py
oliver-sanders: Fine, we will deal with this with future gui work.

PR #2135, opened 2018-01-02
Happy New Year

oliver-sanders:


oliver-sanders:
`t/rosie.db/00-query.py` passes for me locally, on Travis it appears `%%` values are being returned as `%`.

PR #2139, opened 2018-01-15
rosa svn-pre-commit: prevent file at branch level

matthewrmshin:
Pre-commit should prevent adding a file at branch level.
Post-commit should no longer die if this happens.


matthewrmshin:
Yes, I am slowly moving our tests towards safety - always quoting strings that may contain spaces and may be used as file system locations. `${NAME}` is also preferable to `$NAME` as it improves readability and reduces the chance of accidentally joining `NAME` together with subsequent strings.

sadielbartholomew:
Ah I see, so it is the other tests that are the issue if anything! I will make note of this for future for when writing my own tests as part of development work.

PR #2138, opened 2018-01-12
rose-arch: Raise error for duplicate targets.

oliver-sanders:
It is possible to pass rose arch a pair of duplicate targets using the `[arch:(target)]` syntax e.g:

```
[arch:foo]

[arch:(foo)]
```

This PR catches the configuration error before it surfaces from sqlite to provide a more helpful error.

PR #2143, opened 2018-01-25
Fix docstring typo from #2138 merge

sadielbartholomew:
Quick fix for trivial feedback change so @oliver-sanders' PR #2138 could be merged immediately given he is away on leave.


matthewrmshin:
(Trivial change. One review is enough.)

PR #2145, opened 2018-01-25
Improve suite still running error message

### LINKED ISSUE #2142: rose suite-run: pgrep for suite processes?
The above functionality was removed by #2109, but the PR doesn't really explain the reason behind the change.   Now `rose suite-run` just trusts the suite contact file on whether or not a suite is already running.  The error message, however, suggests that actual suite processes were found even if the suite was killed leaving an orphaned contact file:
``` 
% rose suite-run
[FAIL] Suite "foo" has running processes on: blob
[FAIL] Try "cylc stop 'foo'" first?
```
This confuses some users.  Could we either:
 * restore the check for processes on the suite host as recorded in the contact file
 * or at least add a more accurate message, i.e. that the is running *according to the contact file*; therefore stop the suite, or kill its processes, OR remove the orphaned contact file.

matthewrmshin:
Now include information of the location of the contact file and relevant information from within.

Close #2142.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,4 +31,4 @@ install:
     - sudo apt-get install -y subversion
 
 script: 
-    - rose test-battery --state=save -j 5 || (echo -e "\n\nRerunning Failed Tests...\n\n"; rose test-battery --state=save,failed -j 5)
+    - rose test-battery --state=save -j 5 || (echo -e "\n\nRerunning Failed Tests...\n\n"; rose test-battery -v --state=save,failed -j 5)
--- a/lib/python/rose/suite_engine_proc.py
+++ b/lib/python/rose/suite_engine_proc.py
@@ -201,21 +201,11 @@ class SuiteStillRunningError(Exception):
 
     """An exception raised when a suite is still running."""
 
-    FMT_HEAD = "Suite \"%(suite_name)s\" has running processes on:"
-    FMT_BODY_1 = " %(host_name)s"
-    FMT_BODY_M = "\n    %(host_name)s"
-    FMT_TAIL = "\nTry \"cylc stop '%(suite_name)s'\" first?"
+    FMT_HEAD = "Suite \"%(suite_name)s\" appears to be running:\n"
 
     def __str__(self):
-        suite_name, host_names = self.args
-        ret = self.FMT_HEAD % {"suite_name": suite_name}
-        if len(host_names) == 1:
-            ret += self.FMT_BODY_1 % {"host_name": host_names[0]}
-        else:
-            for host_name in host_names:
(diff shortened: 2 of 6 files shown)


lib/python/rose/suite_engine_procs/cylc.py
```diff
@@ -76,6 +79,27 @@ def check_global_conf_compat(self):
...
+        fname = self.get_suite_dir(suite_name, ".service", "contact")
+        try:
+            lines_str = open(fname).read()
+        except IOError:
+            return  # No contact file, suite not running
+        else:
+            extras = ["Contact info from: \"%s\"\n" % fname]
+            for line in lines_str.splitlines(True):  # splitlines keep ends
+                if line.split("=")[0] in self.CONTACT_KEYS:
+                    extras.append("\t%s" % line)
```
oliver-sanders: This tab character only results in a single space on the terminal for me, perhaps use spaces instead?

```
$ rose suite-run
[INFO] chdir: /suite/path/
[FAIL] Suite "suite-name" appears to be running:
[FAIL] Contact info from: "/suite/run/dir/.service/contact"
[FAIL] 	CYLC_SUITE_HOST=suite-host
[FAIL] 	CYLC_SUITE_OWNER=user
[FAIL] 	CYLC_SUITE_PORT=12345
[FAIL] 	CYLC_SUITE_PROCESS=19836 python cylc/path suite-name
[FAIL] Try "cylc stop 'suite-name'" first?
```

.travis.yml
```diff
@@ -31,4 +31,4 @@ install:
     - sudo apt-get install -y subversion
 
 script: 
-    - rose test-battery --state=save -j 5 || (echo -e "\n\nRerunning Failed Tests...\n\n"; rose test-battery --state=save,failed -j 5)
+    - rose test-battery --state=save -j 5 || (echo -e "\n\nRerunning Failed Tests...\n\n"; rose test-battery -v --state=save,failed -j 5)
```
oliver-sanders: Is the `-v` purposeful?

matthewrmshin: Left for debug, probably still useful to capture future failures.

lib/python/rose/suite_engine_procs/cylc.py
matthewrmshin: Replaced tab with spaces.

lib/python/rose/suite_engine_proc.py
```diff
@@ -383,9 +373,7 @@ def check_global_conf_compat(self):
 
     def check_suite_not_running(self, suite_name):
         """Raise SuiteStillRunningError if suite is still running."""
```
sadielbartholomew: Can we adjust this docstring to account for the change in exception?

PR #2146, opened 2018-01-25
rose bush: fix template: hierarchical suite names

### LINKED ISSUE #2133: Rose Bush:  issues with top links when suite name contains one or more '/' characters
Hi,
We were investigating using Rose suites with (supposedly valid) cylc suite names containing forward slash `/`
Rose suite-run and Cylc both work as expected, the suites cycle ok.
Rose bush does not work properly for suites with `/` in their name

We rsync the ~/cylc-run directory to a web server
Rose Bush shows the 'cycles list' and 'task jobs list' when clicked on from the list.
When viewing further pages 'inside' a suite, the top menu links (cycles list, task jobs list, cylc files, rose files, broadcast list etc) give a `Not Found
The requested URL /rose-bush/view/<user>/<suite name part1>/<suite name part 2> was not found on this server.` 
Note: The links work for suites that do not have `/` in their name.

matthewrmshin:
Use `?&suite=SUITE` instead of `/SUITE` in links.

Fix #2133.

### DIFF
--- a/lib/html/template/rose-bush/broadcast-events.html
+++ b/lib/html/template/rose-bush/broadcast-events.html
@@ -6,7 +6,7 @@
 
 <div class="row"><div class="col-md-12">
 <ul class="list-inline pull-right">
-  <li><a href="{{script}}/broadcast_states/{{user}}/{{suite|replace("/", "%2F")}}">states</a></li>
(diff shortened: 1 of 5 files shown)


matthewrmshin:
It is to make it easier to add more stuff at the head of the query string in the future. Or to swap things around.

PR #2152, opened 2018-02-09
Fix test

matthewrmshin:
Long running of test may cause suite to be killed by timeout, which may
cause it to leave behind a contact file, resulting in `rose suite-clean`
dying.


matthewrmshin:
(The other failure on master was due to broken cylc user guide links. I have fixed that separately.)

PR #2160, opened 2018-03-01
bin/rose fix

oliver-sanders:
On certain platforms the `wc` command output is polluted by whitespace which breaks the alias system in `bin/rose`.


matthewrmshin:
(Test failure unrelated. However, I think we should replace the hyperlinks to FCM to Github's metomi/fcm to improve robustness and reduce the hits to MO's servers.)

oliver-sanders:
These links all disappear in the new documentation.

matthewrmshin:
(One review sufficient here.)

Issue #2163, opened 2018-03-05
Ability to validate a suite.rc through rose suite-run without installing


ColemanTom:
I think it would be useful to have an option to `rose suite-run` which bypasses
installing the suite, so that files in the real running area are not touched. A
simple `cylc validate` wouldn't work because rose inserts jinja variables into
the `suite.rc` file.

This in particular in feel is useful for a CI framework as one of the
integration level tests. Semi-related to https://github.com/metomi/rose/issues/1908

NOTE: I am submitting a merge request for this at the same time as raising this Issue
as I think it would be useful (and it helped me learn a bit more about how
rose works).

PR #2164, opened 2018-03-05
Add validate suite only install option

ColemanTom:
This closes https://github.com/metomi/rose/issues/2163

This installs the suite into a temporary area, and then does a cylc validate. This is to ensure the suite as a whole is still good without modifying any "real" official area.

From the second commit message:  
The --validate-suite-only option bypasses the remote and local installation
of a suite. Instead it will create a temporary copy and only perform the
`cylc validate` command, cleaning up the temporary folder afterwards. No
directories or files are updated in the `~/cylc-run` area on disk.

The benefit of this option is that it allows people to have a simple test
through a CI framework on any pull request, or prior to packaging a system
up for automated deployment. At bare minimum, it is nice to ensure that
a change has not broken the suite.rc setup without any 'permanent' changes
to a disk.

NOTE: I would suggest have a look with whitespace changes hidden first to
see what code has actually changed, and what is just indentation for the try/finally
block.


ColemanTom:
I will fix up the pycodestyle problems soon, but I need to head home so it will have to wait. But, it can be discussed irrespectively.

ColemanTom:
I've updated based on pep8, so the linting passes now. I did notice that the tests I added to rose suite run (25-...) did not run. Is there somewhere I was meant to add them to, or is there a configuration issue? I'm going to assume I missed something.

matthewrmshin:
(The new tests are OK. The tests associated with remote installation are correctly skipped.)

PR #2165, opened 2018-03-07
bush: Only serve information with global read permission.

oliver-sanders:
To facilitate running ad-hock rose bush web-servers.

Is checking the full directory tree overkill?


oliver-sanders:
Closing as superseded by #2167.

PR #2167, opened 2018-03-15
bush: Only serve files from log dir + suite.rc[.processed]

oliver-sanders:
Solution to prevent misuse of rose bush until #2165 is merged / closed.

### DIFF
--- a/lib/python/rose/bush.py
+++ b/lib/python/rose/bush.py
@@ -474,6 +474,7 @@ class RoseBushService(object):
     def get_file(self, user, suite, path, path_in_tar=None, mode=None):
         """Returns file information / content or a cherrypy response."""
         f_name = self._get_user_suite_dir(user, suite, path)
+        self._check_file_path(path)
         conf = ResourceLocator.default().get_conf()
         view_size_max = int(conf.get_value(
             ["rose-bush", "view-size-max"], self.VIEW_SIZE_MAX))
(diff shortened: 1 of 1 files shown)


oliver-sanders:
At the moment there is some duplicate checking of parameters. Rather a parameter is checked twice than not checked at all. To prettify things we would need either to:

* Begin a larger refactor.
* Turn the logic on its head, checking parameters as they arrive in the exposed methods (this will probably result in some code duplication as we would need to form then reform paths).

sadielbartholomew:
@oliver-sanders: fair enough, I agree priority-wise it is not crucial to remove duplication & I believe you if you say it is not possible to do with a few changes. I would suggest creating an issue for this but it seems relatively unimportant.

Sorry if I was holding things up with that comment, I was just awaiting a reply.

oliver-sanders:
> I would suggest creating an issue for this but it seems relatively unimportant.

It would be nice to have a general tidy up of rose bush in general, the exposed methods are much too long and should be broken down. We might be able to add some automatic checking of arguments...

oliver-sanders:
We now have to think about porting this to https://github.com/cylc/cylc/pull/2614

sadielbartholomew:
Yes, perhaps once it is migrated we can go in & do the tidy up. Seems like a good time.

sadielbartholomew:
Yes, indeed - I will add it to the checklist so I do not miss it.

PR #2170, opened 2018-03-28
Rose user guide rose tutorial

oliver-sanders:
~**Built on #2166, hold review until merged (will resolve conflicts after this merge).**~

Rose tutorial for the new documentation:

* [x] What is rose
* [x] Rose applications
* [x] Rose metadata
* [x] Rose suite configurations
* [x] Rosie & FCM (needs finalisation)
* [x] Additional rose topics (built-in apps, rose utilities)

Main objectives for this content:

* Explain rose configurations [in a generic sense] (purpose, format).
* Convey the relationship between rose applications, suites and rose configurations.
* Clarify the purpose of rose applications (namely encapsulation).
* Clarify the purpose(s) of rose metadata.
* Clarify the relationship between rose suite configurations and cylc suites (cylc suites with rose configurations).
* Introduce and reinforce rose terminology.
* Introduce the tools for evaluating rose configurations
    * [x] `rose metadata-check`
    * [ ] ~`rose config`~
* Introduce the tools for running rose configurations
    * [x] `rose app-run`
    * [x] `rose suite-run`
    * [x] `rose task-run`

To do in a future PR:

* rose config?
* rose config edit section?
* rosie go section?
* rose stem section

~Ready for preliminary review (structure, approaches, etc).~

PR #2177, opened 2018-04-25
Rose user guide rose stem

oliver-sanders:
**Built on #2170, merge after**

Migrate the rose stem user guide into the new rose documentation. Some documentation moved into the codebase, otherwise a direct translation.

### DIFF
--- a/etc/rose.conf.example
+++ b/etc/rose.conf.example
@@ -318,6 +318,11 @@ path-prepend.PYTHONPATH=
 [rose-ana]
 # Items to prepend to the search path for user methods.
 method-path=/path/1 /path2
+# :default: .false.
+#
+# Turns on the :ref:`Rose Ana Comparison Database
+# <The Rose Ana Comparison Database>`.
+kgo-database=.true.
 
 
 # Configuration related to the database of the Rosie web service server.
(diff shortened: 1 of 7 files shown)


stevewardle:
It all looks fine to me, with the exception of the docstring at the top of `grepper.py`... Although perhaps I'm just unfamiliar with the markdown syntax... it says `rose_ana[ana:ANALYSIS_CLASS]'` - I assume that translates to the `[rose_ana]` section of a user `rose.conf` file?  The example of using this in the rest of the documentation is correct I just can't see how this comment relates to it...  Likewise the rest of that sentence/paragraph reads a little oddly... the 3 places it can go are:

- In the `[rose_ana]` section of the site configuration
- In the `[rose_ana]` section of a user's personal `rose.conf`
- In the `[ana:config]` section of a `rose_ana` app

oliver-sanders:
Hows this?

> I assume that translates to the `[rose_ana]` section of a user `rose.conf` file?

It translates to the `[rose-ana]` section in the `rose.conf` file, is this meant to be `[rose_ana]`?

oliver-sanders:
Is `[ana:config]` different to `[ana:ANALYSIS_CLASS]`?

stevewardle:
I would say so... the sections in the `rose_ana` files are based on the name of the class they are meant to call (I guess that might be where "Analysis Clask" in that comes from) but the section that can contain the "global" settings must be called `[ana:config]`

PR #2179, opened 2018-04-26
Rose user guide slides

oliver-sanders:
~**WIP**~

Add a slide functionality using the hieroglyph (http://docs.hieroglyph.io/en/latest/) slide builder.

It makes the markup a bit messy but on the whole seems to be the best option. Extending hieroglyph to implement a nicer interface might be possible but it's probably quite a lot of work.

### DIFF
--- a/sphinx/_static/js/minicylc.js
+++ b/sphinx/_static/js/minicylc.js
@@ -282,9 +282,11 @@ class MiniCylc {
 
 
 // Activate minicylc.
-$('.minicylc').each(function() {
-    var obj = this;
-    $(this).find('object:first').on('load', function() {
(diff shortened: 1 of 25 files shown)


sphinx/_static/js/minicylc.js
```diff
@@ -282,9 +282,11 @@ class MiniCylc {
 
 
 // Activate minicylc.
-$('.minicylc').each(function() {
-    var obj = this;
-    $(this).find('object:first').on('load', function() {
-        new MiniCylc(obj).run();
+$(document).ready(function() {
```
oliver-sanders: This is just general good practice.

PR #2183, opened 2018-05-04
Rose user guide erratum 2

oliver-sanders:
Minor corrections, highlights:

* Fix tutorial inconsistencies
* Some minor re-structuring to ensure pages are correctly placed in the document structure (look at the sidebar)
* Use `xxxxxxxxx` in the documentation in place of the api-key.
* Include the required html redirects in the git repo rather than creating at deploy time.


sphinx/tutorial/cylc/runtime/introduction.rst
```diff
@@ -313,13 +262,65 @@ Where Do All The Files Go?
          :align: center
          :scale: 75%
 
+
+Running A Suite
```
sadielbartholomew: I agree that this section is slightly better placed a bit further down the 'Cylc Tutorial » Runtime » Introduction' main page, but it now comes under the 'Where Do All The Files Go?' high-level heading, & it doesn't really fit there. I think it might be clearer to put it just above 'Where Do All The Files Go?', perhaps grouping it with 'The Cylc GUI' as a section 'Running and Monitoring A Suite' or similar. I would also consider putting 'Suite Files' (excluding the practical if that would count as being part of this) at the top of the 'Where Do All The Files Go?' section so that the 'work' and 'job log' directories are introduced briefly on a general level with the other generated files before they are described in detail in sub-sections.

sphinx/tutorial/cylc/runtime/introduction.rst
```diff
@@ -340,6 +341,8 @@ Suite Files
    ``share/``
       The :term:`share directory` is a place where :term:`tasks <task>` can
       write files which are intended to be shared within that cycle.
+   ``work/``
+      A directory hierarchy containing tasks :term:`work directory`.
```
sadielbartholomew: The 'term' / glossary link just sits weirdly at the end of the sentence here. I suggest rephrasing as in the the ``share/`` description item just above to incorporate the term naturally.

sphinx/tutorial/cylc/scheduling/datetime-cycling.rst
```diff
@@ -183,21 +183,23 @@ Recurrence Formats
...
-   start on the first midnight *after* the initial cycle point.
+      ``2000-01-01T12``, ``2000-01-02T12``, ``2000-01-03T12``, ...
 
-   * ``T00/P1D``
-   * ``T00``
+      This can easily be adjusted. Both of the following recurrences induce a
+      start on the first midnight *after* the initial cycle point.
+
+      * ``T00/P1D``
+      * ``T00``
```
sadielbartholomew: Perhaps add a short final sentence along the lines of 'This is why *explicit* recurrences such as ``T00`` are preferred to / safer than *implicit* ones such as ``P1D``' to hit the point home.

sphinx/tutorial/cylc/runtime/configuration-consolidation/index.rst
```diff
@@ -68,20 +68,23 @@ Consolidating Configuration
...
                   """
 
 .. nextslide::
 
 Cylc offers three ways of consolidating configurations to help improve the
 structure of a suite and avoid duplication.
 
-* Families
-* Jinja2
-* Parameterisation
```
sadielbartholomew: A note for future: as discussed in person, the 'three approaches' are now appearing on the index (left-hand) menu which is great, but upon clicking the links to go one of the pages from the section link bullets under the 'The Three Approaches' section, the relevant pages are not highlighted and also the menu collapses to the 'Runtime' level, which is not ideal. Strangely it works as desired (highlighted, with no collapse) when accessed from *this* list, though readers will usually use the other one given the flow here!

sphinx/tutorial/cylc/runtime/introduction.rst
oliver-sanders: The motive of this change is so that the start focuses on running tasks and the end on suites. Before we talked about tasks, then running suites, then back to tasks again. We can't order the content differently to the contents.

sadielbartholomew: Ah, I see. This motivation wasn't immediately obvious to me, but then again I didn't read the page from fresh. As long as there is a clear reason for placing certain sections in the given order, that is fine.

sphinx/tutorial/cylc/scheduling/datetime-cycling.rst
oliver-sanders: I agree, this section is kinda strange so I've re-written it, see what you think. I haven't expressed a solid preference as it's down to the application.

sphinx/tutorial/cylc/runtime/introduction.rst
```diff
@@ -342,7 +342,8 @@ Suite Files
       The :term:`share directory` is a place where :term:`tasks <task>` can
       write files which are intended to be shared within that cycle.
    ``work/``
-      A directory hierarchy containing tasks :term:`work directory`.
+      A directory hierarchy containing task's :term:`work directories
```
sadielbartholomew: Ah, this is what was meant! My apologies. The apostrophe does make it more obvious :)

sphinx/tutorial/cylc/scheduling/datetime-cycling.rst
sadielbartholomew: Very true, it will be the context that dictates which is 'better'. Regardless was worth emphasising the difference as it is I think quite subtle, and the re-write is very clear and complete. Nice touch with the suite.rc excerpts and diagrams!

sphinx/tutorial/cylc/runtime/introduction.rst
sadielbartholomew: Apologies, I misunderstood what was meant here looking at your feedback change. The apostrophe added does make it more obvious :)

sadielbartholomew:
(Re-approving after feedback, though Travis CI seems to have stalled for the last commit so re-running it).

PR #2186, opened 2018-05-14
Test tidy up

matthewrmshin:
Remove stray output.
Tidy up suite run time directory.
Fix trap syntax.


matthewrmshin:
Just minor changes to tests. One review should be sufficient.

PR #2190, opened 2018-05-24
Rose user guide: versioning

oliver-sanders:
Add functionality to `rose make-docs` to build versioned documentation.

Documentation is built incrementally so new versions can be added without having to rebuild old ones i.e:

```console
$ git checkout 2018.05.0
$ rose make-docs
$ tree doc -L 1
doc
|-- 2018.05.0
|-- doc -> 2018.05.0
|-- index.html
`-- versions.json
$ git checkout some-future-version
$ rose make-docs
$ tree doc -L 1
doc
|-- 2018.05.0
|-- checkout some-future-version
|-- doc -> checkout some-future-version
|-- index.html
`-- versions.json
```

As the documentation is built incrementally it is non-destructive in version control terms meaning we can do:

```console
$ git push upstream gh-pages  # no need to -f force
```

This would make auto-deployment of documentation much nicer.

Static HTML redirects are created by `rose make-docs` including a legacy re-direct for `doc/rose.html`.

The version switcher AJAXes the versions.json file to get a dictionary of available versions and formats, this AJAX means that the version selector will only work when served from behind a web server (or very permissive browser - cross site scripting).

### DIFF
--- a/404.md
+++ b/404.md
@@ -1,5 +1,5 @@
 [comment]: <> (This is the 404 page as used on the Github Pages site.)
 
-## The Rose Documentation Has Moved
+## 404 Page Not Found - The Rose Documentation Has Moved
 
 Please update your bookmarks to: http://metomi.github.io/rose
--- a/404.md
+++ b/404.md
@@ -1,0 +1,5 @@
+[comment]: <> (This is the 404 page as used on the Github Pages site.)
+
+## The Rose Documentation Has Moved
+
+Please update your bookmarks to: http://metomi.github.io/rose
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,15 @@
+# This file configures the Jekyll static web builder (https://jekyllrb.com/) as
+# used by Github Pages.
+
+# Jekyll treats directories beginning with underscores as having special meaning
+# and omits them from the build. This can be overridden by manually including
+# directories using "include".
+# https://help.github.com/articles/files-that-start-with-an-underscore-are-missing/
+# NOTE: Only the directory name should be specified not the full path.
+include: [
+    "_downloads",
+    "_images",
+    "_modules",
+    "_sources",
+    "_static"
+]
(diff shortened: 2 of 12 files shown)


404.md
```diff
@@ -0,0 +1,5 @@
+[comment]: <> (This is the 404 page as used on the Github Pages site.)
+
+## 404 Page Not Found - The Rose Documentation Has Moved
+
+Please update your bookmarks to: http://metomi.github.io/rose
```
oliver-sanders: We could write this file in `rose make-docs` if we don't want it kicking about in the repo.

_config.yml
```diff
@@ -0,0 +1,15 @@
...
+# directories using "include".
+# https://help.github.com/articles/files-that-start-with-an-underscore-are-missing/
+# NOTE: Only the directory name should be specified not the full path.
+include: [
+    "_downloads",
+    "_images",
+    "_modules",
+    "_sources",
+    "_static"
+]
```
oliver-sanders: Same applies to the Jekyll config too.

404.md
sadielbartholomew: I think it is fine as it is.

etc/deploy-docs
```diff
@@ -0,0 +1,81 @@
...
+    git rm -rf $DOCUMENTATION_FILES     # remove tracked files
+    git clean -xf $DOCUMENTATION_FILES  # remove untracked files
+    mv doc "$TEMP_DOCS_LABEL"  # permit the alias "doc"
+    mv "$TEMP_DOCS_LABEL"/* .
+    git clean -xf "$TEMP_DOCS_LABEL"
+    git add * -f
+
+    # commit and push
+    git commit -m "${ROSE_VERSION}"
+    # git push "${REMOTE}" gh-pages
```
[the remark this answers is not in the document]
sadielbartholomew: I suspected it was some sort of safety thing like this. Thanks for explaining.

sadielbartholomew:
(Re-approved after final commit which is trivial hence fine to merge PR without first reviewer re-approval.)

PR #2141, opened 2018-01-19
metadata-check: ignore GTK widgets if no display

matthewrmshin:
Allow check to pass for GTK widgets if there is no display in the environment.

Problem reported by @scwhitehouse.

PR #2192, opened 2018-05-29
more reliable check for display capability

oliver-sanders:
Further to #2141, #2169 

It would appear that the measures put in place still do not cover all cases. Note that `import gtk` in `rose.metadata_check` is equivalent to `import rose.gtk`.

For reference the offending function in `gtk` is `set_interactive` which is called on import in `gtk`: https://developer.gnome.org/pygtk/stable/gtk-functions.html#function-gtk--set-interactive

### DIFF
--- a/lib/python/rose/gtk/__init__.py
+++ b/lib/python/rose/gtk/__init__.py
@@ -1,0 +1,8 @@
+try:
+    import pygtk
+    pygtk.require('2.0')
+    import gtk
+except (ImportError, RuntimeError):
+    INTERACTIVE_ENABLED = False
+else:
+    INTERACTIVE_ENABLED = True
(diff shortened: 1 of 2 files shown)


matthewrmshin:
@sadielbartholomew please sanity check.

lib/python/rose/gtk/__init__.py
```diff
@@ -0,0 +1,8 @@
+try:
+    import pygtk
+    pygtk.require('2.0')
+    import gtk
+except (ImportError, RuntimeError):
```
sadielbartholomew: ``.require()`` seems to throw an ``AssertionError`` when the 'required version is not found on system', so shouldn't this be added to exceptions?

oliver-sanders: Good point!

Issue #2103, opened 2017-08-08
Setting remote root-dir{work} no longer works with csh


MartinDix:
Our site still has a few users with csh as their login shell.

Our server site configuration sets
```
root-dir{share}=raijin*=/short/$PROJECT/$USER
root-dir{share/cycle}=raijin*=/short/$PROJECT/$USER
root-dir{work}=raijin*=/short/$PROJECT/$USER
```

With rose 2017.01.0 the command that's run was
```
 ssh -oBatchMode=yes -oConnectTimeout=10 raijin.nci.org.au bash --login -c \'ROSE_VERSION=2017.01.0\ /projects/access/bin/rose\ suite-run\ -v\ -v\ --name=au-aa398\ --install-only\ --run=run\ --remote=uuid=8f2d273a-8420-4486-ba73-3907d85de521,root-dir{share}=/short/$PROJECT/$USER,root-dir{share/cycle}=/short/$PROJECT/$USER,root-dir{work}=/short/$PROJECT/$USER\'
```
This meant that ```{share}``` etc was interpreted by bash and everything worked ok.

WIth 2015.05.0 (and remote-no-login-shell=false) the command that's run is
```
ssh -oBatchMode=yes -oConnectTimeout=10 -n raijin.nci.org.au env\ ROSE_VERSION=2017.05.0\ CYLC_VERSION=7.4.0\ bash\ -l\ -c\ \'\"$0\"\ \"$@\"\'\ /projects/access/bin/rose\ suite-run\ -vv\ -n\ au-aa398\ --install-only\ --run=run\ --remote=uuid=df365ae4-cd0f-424a-9c39-a950bdb9cd68,root-dir{share}=/short/$PROJECT/$USER,root-dir{share/cycle}=/short/$PROJECT/$USER,root-dir{work}=/short/$PROJECT/$USER
```
Here it seems that the arguments are interpreted by the remote login shell before being passed to bash and so the {} get dropped by csh. rose then actually gets arguments like root-dirshare and so all the work and share directories end up in the home filesytem.

I think the best approach is to use ```re.escape``` on ```root-dir{share}``` etc. This seems to work with all shells and would mean we could set remote-no-login-shell=true.

Aa far as I can tell, nothing in the rose test-battery tests whether these directories are correctly created remotely. Would it be worth me trying to write a test for this?

matthewrmshin:
`pipes.quote` is probably better than `re.escape` for the purpose.

[...]

PR #2194, opened 2018-06-01
Fix root-dir settings for remote machine.

aosprey:
This relates to #2103, which causes a similar issue for us. 

To repeat, we set: `root-dir=login*.archer.ac.uk=$DATADIR`. But in the submission command `$DATADIR` is interpreted on the rose/cylc server machine, which has no knowledge of what `$DATADIR` should be on the HPC. And so the suite directory is not created in the correct place. This doesn't seem to be an issue with the shell (as in #2103), as I use bash on both machines. 

Adding quotes around the value of the `root-dir` settings fixes it for us.

### DIFF
--- a/lib/python/rose/suite_run.py
+++ b/lib/python/rose/suite_run.py
@@ -24,6 +24,7 @@ from datetime import datetime
 from fnmatch import fnmatchcase
 from glob import glob
 import os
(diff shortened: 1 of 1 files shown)


matthewrmshin:

PR #2199, opened 2018-06-18
docs: update rose and cylc descriptions

oliver-sanders:


sphinx/tutorial/rose/index.rst
```diff
@@ -3,7 +3,11 @@
 Rose Tutorial
 =============
 
-Rose is a system for creating, editing and running Rose configurations.
+Rose is a system for writing, editing and running application configurations.
```
matthewrmshin: How about *toolkit* instead of *system*?

PR #2202, opened 2018-06-26
rose suite-hook --mail: obsolete message

matthewrmshin:
It will now add an obsolete message to all emails it sends.

### DIFF
--- a/lib/python/rose/suite_hook.py
+++ b/lib/python/rose/suite_hook.py
@@ -73,7 +73,9 @@ class RoseSuiteHook(object):
         # Send email notification if required
         email_exc = None
         if should_mail:
-            text = ""
+            text = (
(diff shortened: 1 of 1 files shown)


oliver-sanders:
We could do with providing a bit more information on this topic. We could put this somewhere in the Rose Documentation or just add it to the message.

We should:
* Explain that `rose suite-hook --mail` (with no arguments?) doesn't do anything cylc doesn't
* List the task states for which cylc will send a message (by default)
* Explain how to add states to this list

sadielbartholomew:
I agree with Oliver. A bit more detail would be ideal.

oliver-sanders:
@matthewrmshin

sadielbartholomew:
Bump.

matthewrmshin:
Delaying this for further discussions. In order to satisfy your comments, we'll need to discuss stuffs that currently belong local site configurations.

PR #2203, opened 2018-06-29
docs: fix tutorial solution

oliver-sanders:
See https://github.com/cylc/cylc/issues/2707#issuecomment-400938301


oliver-sanders:
> Kicking myself that I didn't spot this one!

Me too, especially as we've now run three training sessions using this material!

PR #2205, opened 2018-07-02
fix ., .. versions and ensure .html url endings

oliver-sanders:
Apparently extglob behaviour with dot-files is not completely standard across bash versions. Solve by removing `.git` from the the `DOC_VERSIONS` pattern as it is not needed there anyway.

### DIFF
--- a/bin/rose-make-docs
+++ b/bin/rose-make-docs
@@ -85,7 +85,6 @@ SPHINX_PATH=sphinx
 DOCS_DIR="${ROSE_HOME}/doc"
 # documentation output directory for this version
 BUILD_DIR="${DOCS_DIR}/${ROSE_VERSION}"
-# glob for documented rose versions
 # glob for documentation formats within a rose version
 DOC_FORMATS=!(*.html|doctrees)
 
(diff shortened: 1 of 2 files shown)


bin/rose-make-docs
```diff
@@ -142,6 +141,7 @@ done
 if [[ -z "${BUILDS}" ]]; then
     BUILDS='html'
 fi
+DOC_VERSIONS=!(*.html|doc|versions.json|"${DEFAULT_ALIAS}"|CHANGES.md|404.md|_config.yml)
```
sadielbartholomew: A few minor things RE this line:

1. Now it has been moved below (new) line 106, which sets ``DEFAULT_ALIAS='doc'``, with no code allowing this to be set otherwise, so 'doc' is globbed twice I believe. It's inconsequential but not ideal.
2. The comment associated with this has been left behind & is feeling very lonely.

oliver-sanders: 1. Yes `doc` is in the glob twice, I don't think there's any consequence to this. I put `DEFAULT_ALIAS` in to make it easier to change this link later (maybe we might want to change it to "stable" or "latest" as used by other projects).
2. Aw, we don't want any lonely comments.

sadielbartholomew: Ah I see, in that case it's fine.

Issue #2206, opened 2018-07-11
editing quoted types in a raw entry widget


oliver-sanders:
There is some strange behaviour with rose edit and quoted data types.

*rose-app.conf*
```ini
[env]
foo='$foo'
```

*meta/rose-meta.conf*
```ini
[env:foo=foo]
type=quoted
```

To replicate open `rose edit`, paste in the value `'bar'` and click save. The result will be `bar` within three sets of single quotes:

*rose-app.conf*
```ini
[env]
foo='''bar'''
```

This is because the environment variable has forced the value to be rendered in a raw entry widget.

oliver-sanders:
#2207 greatly reduces the prominence of this issue.

PR #2207, opened 2018-07-11
env vars in quoted types

oliver-sanders:
Partly addresses #2206

When environment variables are present in values `rose edit` safely falls back to `raw`. Unfortunately this isn't actually safe with quoted data types (#2206).

To solve don't fall back to the `raw` entry widget if the metadata type is a text type e.g. `quoted`. This does not solve #2206 as one could still edit the field from a raw widget (i.e. with metadata turned off) but this should prevent the overwhelming majority of users encountering the problem.

Also two bolt-on fixes:

* Document the multi line string data type.
* Add `python_boolean` support to `LogicalArrayValueWidget` (seems strange not to support it).

### DIFF
--- a/demo/rose-config-edit/demo_meta/app/01-types/meta/rose-meta.conf
+++ b/demo/rose-config-edit/demo_meta/app/01-types/meta/rose-meta.conf
@@ -18,6 +18,11 @@ description=A logical array of any length
 length=:
 type=boolean
 
+[namelist:nl1=my_python_boolean_array]
+description=A boolean array of length 3
+length=3
+type=python_boolean
+
 [namelist:nl1=my_char]
 description=A character variable
 type=character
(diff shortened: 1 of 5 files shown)


sadielbartholomew:
Sensible partial fix, though:
* we should add ``my_python_boolean_array`` into the tests ``t/rose-macro`` & ``t/rose-metadata-check/01-type-length.t`` to keep these up-to-date
* can we test to confirm that quoted data types are treated safely with raw-entry widgets (to the extent guaranteed in this PR) or can it wait until a full solution to #2206?

oliver-sanders:
> we should add my_python_boolean_array into the tests

This PR relates only to the widget that `rose edit` loads to display the config. We don't have any automatic tests for this *(though in a future web GUI world we should definitely look into visual diffs etc)*, there is, however, a manual test battery in `etc/demo` which I've added the setting to.

> can we test to confirm that quoted data types are treated safely with raw-entry widgets

No, that must wait on #2206.

sadielbartholomew:
Fair enough @oliver-sanders, thanks for clarifying. As long as testing is as thorough as practicable I am happy.

PR #2210, opened 2018-07-16
document rose-app.conf|opts

oliver-sanders:
Another setting which has escaped documentation.


oliver-sanders:
After a bit of testing it seems that if you can use `opts` in the `rose-suite.conf` file similarly to the `rose-app.conf` file, makes sense.

PR #2216, opened 2018-07-25
docs: permit documenting rose objects via intersphinx

oliver-sanders:
This PR allows rose objects to be referenced from other Sphinx projects (i.e. Cylc).

Invalid references to rose objects from other projects will result in warnings which will cause a build failure if run in strict mode (same behaviour as referencing rose objects locally).

Unfortunately this is rather hard to test without creating a dummy sphinx project to reference from. Once we start using this in the Cylc documentation (and vice versa) it will be tested (by virtue of being used and failing on error).

Minimal example sphinx project to demonstrate intersphinx referencing.

*conf.py*
```python
import os
import sys

# add extensions directory to python path
sys.path.append(os.path.abspath('ext'))

# register sphinx extensions
extensions = [
    'sphinx.ext.intersphinx',
    'rose_domain'
]

# mapping of project name: project url
intersphinx_mapping = {'myrosedoc': ('http:/<url path to>/rose/doc/html', None)}

# name of the index file
master_doc = 'index'
```
*index.rst*
```rst
Check out the :rose:file:`myrosedoc:rose-app.conf` file!
```

*ext/*

* *rose_domain.py*

### DIFF
--- a/sphinx/ext/rose_domain.py
+++ b/sphinx/ext/rose_domain.py
@@ -138,6 +138,11 @@ OPT_ARG_REGEX = re.compile(  # Regex for splitting domains and arguments
     r'(?:[^=]+)?)'       # Configuration setting.
     r'(?:=(.*))?$'       # Argument.
 )
+CROSS_DOMAIN_REGEX = re.compile(  # Catch leading target from inter-sphinx ref.
+    r'^([\w\-]+)'  # inter-sphinx mapping
+    r':'            # colon divider
+    r'([\w\-].*)$'  # rose object reference
+)
 
 
 def tokenise_namespace(namespace):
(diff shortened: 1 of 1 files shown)


sadielbartholomew:
The HTML & man pages build fine, but for ``make slides`` I get:

```
Sphinx error:
Builder name slides not registered or available through entry point
make: *** [slides] Error 2
```
& with a clean slate (after removing the ``doc`` dir) I also find for ``make latexpdf`` I get no ``.pdf`` file, & ``Fatal error occurred, no output PDF file produced!`` in the produced ``texput.log``.

There seems to be something amiss in these cases.

oliver-sanders:
To build slides you will need to add Hieroglyph to the list of sphinx extensions:

```python
# register sphinx extensions
extensions = [
    'sphinx.ext.intersphinx',
    'hieroglyph',
    'rose_domain'
]
```

If you are getting no `.pdf` file at all it's probably something wrong with your LaTeX environment as the document should build even if the intersphinx reference doesn't work.

sadielbartholomew:
Yes, that has solved it in both cases. Thanks.

Issue #2212, opened 2018-07-16
rose tutorial: odds and ends


oliver-sanders:
* [ ] Extensions for practical exercises
* [ ] [Add a date-time manipulation practical](https://github.com/metomi/rose/blame/2018.06.0/sphinx/tutorial/rose/furthertopics/date-time-manipulation.rst#L121)
* [ ] [Migrate ana database docs into the code base](https://github.com/metomi/rose/blame/2018.06.0/sphinx/tutorial/rose/furthertopics/rose-stem.rst#L306)
* [ ] Simplify trigger tutorial (see TODO notices on file)
* [ ] Simplify metadata tutorial (see TODO notices on file)
* [ ] [An example filetree for a rosie suite with rose applications for reference](https://github.com/metomi/rose/blame/2018.06.0/sphinx/tutorial/rose/summary.rst#L75)

PR #2218, opened 2018-08-02
command-format-less rose_bunch

TomekTrzeciak:
This PR makes rose_bunch compatible with the default command lookup logic.

**Motivation:**

The `[bunch]commad-format` template is sometimes awkward to use. Firstly, it is expanded for environment variables, which makes it inconvenient for writing pieces of shell code. One might need to define variables in more complex commands (e.g.,  `for VAR in ...`), but kludges like `${VAR%}` are necessary to prevent their premature expansion in the command body.

Secondly, it makes factoring out bunching functionality out of the apps harder. One may reasonably want to have, say, a single app for deterministic and ensemble runs with a single `[command]default` for both cases and enable bunching through an optional config or app import on top of the main app config.

**Implementation:**

This PR makes `[bunch]commad-format` setting optional rather than compulsory. If it is not specified then the command to run is determined following the default ``mode=command`` logic. In this case `[bunch-args]` are passed in the environment.

### DIFF
--- a/lib/python/rose/apps/rose_bunch.py
+++ b/lib/python/rose/apps/rose_bunch.py
@@ -27,8 +27,7 @@ from time import sleep
 
 from rose.app_run import (
     BuiltinApp,
-    ConfigValueError,
-    CompulsoryConfigValueError)
+    ConfigValueError)
 from rose.popen import RosePopenError
 import rose.job_runner
 from rose.reporter import Event
@@ -156,15 +155,19 @@ class RoseBunchApp(BuiltinApp):
             self.incremental = rose.env.env_var_process(self.incremental)
 
         multi_args = conf_tree.node.get_value([self.ARGS_SECTION], {})
+        bunch_args_names = []
+        bunch_args_values = []
         for key, val in multi_args.items():
-            multi_args[key].value = rose.env.env_var_process(val.value)
+            bunch_args_names.append(key)
+            bunch_args_values.append(
+                shlex.split(rose.env.env_var_process(val.value)))
 
-        self.format = True
+        self.isformatted = True
         self.command = rose.env.env_var_process(
             conf_tree.node.get_value([self.BUNCH_SECTION, "command-format"]))
 
         if not self.command:
-            self.format = False
+            self.isformatted = False
             self.command = app_runner.get_command(conf_tree, opts, args)
 
         if not self.command:
@@ -188,18 +191,16 @@ class RoseBunchApp(BuiltinApp):
             if instances:
                 arglength = len(instances)
             else:
-                item, val = sorted(multi_args.items())[0]
-                arglength = len(shlex.split(val.value))
+                arglength = len(bunch_args_values[0])
             self.invocation_names = range(0, arglength)
         else:
             arglength = len(self.invocation_names)
 
-        for item, val in sorted(multi_args.items()):
-            if len(shlex.split(val.value)) != arglength:
-                raise ConfigValueError([self.ARGS_SECTION, item],
-                                       conf_tree.node.get_value(
-                                       [self.ARGS_SECTION, item]),
-                                       "inconsistent arg lengths")
+        if any(len(item) != arglength for item in bunch_args_values):
+            raise ConfigValueError([self.ARGS_SECTION, item],
+                                   conf_tree.node.get_value(
+                                   [self.ARGS_SECTION, item]),
+                                   "inconsistent arg lengths")
 
         if conf_tree.node.get_value([self.ARGS_SECTION, "command-instances"]):
             raise ConfigValueError([self.ARGS_SECTION, "command-instances"],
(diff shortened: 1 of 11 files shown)


matthewrmshin:
@TomekTrzeciak Please have a look at CONTRIBUTING.md and add your name.

TomekTrzeciak:
@matthewrmshin, I added myself in CONTRIBUTING.md. I think this is ready for review, tests pass and even codacy is happy! :grin:

sphinx/api/built-in/rose_bunch.rst
```diff
@@ -49,28 +49,33 @@ Configuration
 
 The application is normally configured in the
 :rose:conf:`rose_bunch[bunch]` and :rose:conf:`rose_bunch[bunch-args]`
-sections of the :rose:file:`rose-app.conf` file.
+sections of the :rose:file:`rose-app.conf` file, but `rose_bunch[command]`
```
oliver-sanders: We use the `:rose:conf:` role to link the setting back to the API reference. This way we can detect broken references etc.

```rst
:rose:conf:`rose_bunch[command]`
```

sphinx/api/built-in/rose_bunch.rst
```diff
@@ -49,28 +49,33 @@ Configuration
...
+         are made accessible to the command as environment variables.
+
 
       .. rose:conf:: command-instances=N
 
          Allows the user to specify an integer value for the number of
          instances of a command they want to run. This generates the values
          used by the ``%(command-instances)s`` value in command-format.
+         If `[bunch]command-format` is not specified then the command instance
+         is passed as the ``COMMAND_INSTANCES`` environment variable instead.
```
oliver-sanders: The `COMMAND_INSTANCES` environment variable can go in the environment variable section then be referenced as
```rst
:envvar:`COMMAND_INSTANCES`
```

lib/python/rose/apps/rose_bunch.py
```diff
@@ -179,25 +191,29 @@ def run(self, app_runner, conf_tree, opts, args, uuid, work_files):
...
             arglength = len(self.invocation_names)
 
-        for item, val in sorted(multi_args.items()):
-            if len(shlex.split(val.value)) != arglength:
-                raise ConfigValueError([self.ARGS_SECTION, item],
-                                       conf_tree.node.get_value(
-                                       [self.ARGS_SECTION, item]),
-                                       "inconsistent arg lengths")
+        if any(len(item) != arglength for item in bunch_args_values):
+            raise ConfigValueError([self.ARGS_SECTION, item],
```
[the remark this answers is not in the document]
TomekTrzeciak: Ah, good catch, thanks! Back to the for loop then...

lib/python/rose/apps/rose_bunch.py
```diff
@@ -216,13 +232,17 @@ def run(self, app_runner, conf_tree, opts, args, uuid, work_files):
             self.dao = None
 
         commands = {}
-        for index, name in enumerate(self.invocation_names):
-            invocation = RoseBunchCmd(name, self.command_format, index)
-            for key, vals in sorted(multi_args.items()):
-                invocation.argsdict[key] = shlex.split(vals.value)[index]
+        for vals in zip(range(arglength), self.invocation_names,
```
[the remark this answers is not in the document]
TomekTrzeciak: Maybe and probably wouldn't be that hard to implement. One could argue that `izip_longest` type of passing arguments could be useful too. You could have arguments lists that vary in length and pad the shorter ones with empty strings, for example. But the logic of config option validation is already a bit dense there and I wonder if it's worth to make it even more complicated unless someone really asks for it.

TomekTrzeciak:
One more possible addition occurred to me. A lot of `app-run` and `task-run` options cam be specified though environment, but `--app-mode=...` is suspiciously command line only. Would it make sense to add `ROSE_APP_MODE` to the mix and sneak it in this PR?

sphinx/api/built-in/rose_bunch.rst
TomekTrzeciak: Actually, I think this should be
```
:rose:conf:`rose-app.conf[command]`
```
I will tweak the docs.

oliver-sanders: Yes!

oliver-sanders:
As an aside [for future work], we should also think about the [`rose_bunch` tutorial](http://metomi.github.io/rose/doc/html/tutorial/rose/furthertopics/rose-bunch.html). Should the tutorial use the new approach, I think it's clearer.

If we agree on this we can punt it to #2212.

oliver-sanders:
OK, we think the globbing aspect can be kicked into another issue (#2220), for the moment `bunch_args` variables will appear raw string in Rose Edit which is acceptable even if not ideal.

PR #2221, opened 2018-08-10
rose.config: python 3 compatibility

oliver-sanders:
For the most part Rose is a standalone system utility, however, the `rose.config` module is occasionally imported. In order to support the continued use of this ahead of the planned Python3 migration (to follow GUI replacement) the `rose.conf` module and its Rose dependencies have been upgraded to support both Python 2 & 3.

#### Tripwire

Old-style `cmp` functions have now been retired, unfortunately Rose makes heavy use of them for sorting configurations. The sorting algorithm currently used in rose contains branched logic which changes the sorting algorithm based on comparisons of both elements being compared, ...

https://github.com/metomi/rose/blob/ec78ddffc59ae21335e2f33c5431c225f8942215/lib/python/rose/config.py#L1550-L1573

... because of this **the old sorting algorithm cannot be ported to Python 3**.

*Ok, cannot is probably a little strong, perhaps more like should not*

#### Solution?

I've implemented a new-style rich comparison algorithm, this is not a direct replacement for the old sorting system, there are differences in results. I think the new system yields better results but the diffs could potentially cause issues for users (see the diff in t/rose-namelist-dump/02-type.t)?

We will have to take the pain of this one with the Python3 upgrade?.

#### Going Forward

In this branch I've done the bare minimum required for `rose.config` to run under Python 3, hence I have left the old sorting algorithm in-place where it is still used by other parts of rose (e.g. the Rose Edit GUI). Should we:

1) Delay this until we are ready to upgrade the entire codebase into Python 3 and take the pain then.
2) Push ahead with a mixed system (not actually as bad as it sounds?).
3) Upgrade all old style comparisons and take the pain now.

@metomi/core

### DIFF
--- a/lib/python/rose/config.py
+++ b/lib/python/rose/config.py
@@ -24,17 +24,10 @@
     import os
     from rose.config import *
 
-.. testcleanup:: rose.config
-
-    try:
-        os.remove('config.conf')
-    except OSError:
-        pass
-
 Synopsis:
     >>> # Create a config file.
     >>> with open('config.conf', 'w+') as config_file:
-    ...     config_file.write('''
+    ...     _ = config_file.write('''
     ... [foo]
     ... bar=Bar
     ... !baz=Baz
@@ -49,7 +42,7 @@ Synopsis:
 
     >>> # Retrieve config settings.
     >>> config_node.get(['foo', 'bar'])
-    {'state': '', 'comments': [], 'value': 'Bar'}
+    {'value': 'Bar', 'state': '', 'comments': []}
 
     >>> # Set new config settings.
     >>> _ = config_node.set(['foo', 'new'], 'New')
@@ -92,6 +85,7 @@ What about the standard library ConfigParser? Well, it is problematic:
 
 import copy
 import os.path
+import pprint
 import re
 from rose.env import env_var_escape
 import shlex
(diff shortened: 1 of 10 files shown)


matthewrmshin:
Should add Travis CI Python >3.5 tests for this one?

oliver-sanders:
We could but it would require a separate Travis-CI job, this wouldn't be the worst idea as over time the Py3 test component will grow but we would need to somehow split our test battery into a Py2/Py3 component without having to add `if python -c 'import sys; assert(sys.version[0]) == 3'; then skip all` type statements to the head of each test file.

sadielbartholomew:
Forgive me for failing to understand the subtlety here @oliver-sanders, but why can we not just check for Py 3 and if in use define an equivalent ``cmp()`` method, which can be done very easily? Is the crux here that the current *algorithm* making heavy use of it is not very good?

oliver-sanders:
We could, that would be option 4 in the going forward list above but it wouldn't get around the problem and would leave us in the position where rose config produces different results under different python versions.

oliver-sanders:
There might have been a 5th option to implement the rich comparison functions in the `ConfigNode` class. As these comparisons accept two arguments they could emulate the old logic, however, in order to support custom `sort order` we need to be able to provide a sort function at the point where sorting is performed rather than centrally hardcodeing it.

sadielbartholomew:
Thanks for clarifying. I'll have a think about this & see if I any of the outlined options transpires as the better solution in my view.

oliver-sanders:
If we want Python3.7 testing this can be easily done after #2222.

oliver-sanders:
@wxtim Assigning you on this as your work will supersede this PR.

matthewrmshin:
Close as superseded by #2286.

PR #2222, opened 2018-08-16
docs: auto-deploy documentation from the Travis-CI build

oliver-sanders:
#### If building a git tag (as opposed to a branch) then append the built docs to the current gh-pages branch and push back the result, GitHub will take care of the rest.

Documentation is built incrementally atop what has gone before so changes are not destructive, i.e we don't need to do a force push.

Every now and again we will want to remove older versions from the documentation, this can be done manually:
* Check out the `gh-pages` branch and delete the relevant commit.
* If the relevant commit is the first one, delete the relevant directory and squash the commit with the next one.

Example in motion:

* Example of Travis building a regular branch
   * TB output: https://travis-ci.org/oliver-sanders/rose/builds/416776038
* Example of Travis building a tag (release)
    * TB output: https://travis-ci.org/oliver-sanders/rose/builds/416776730
    * gh-pages branch: http://oliver-sanders.github.io/rose/doc/html/index.html

#### Before this can work on metomi/rose

We need a GitHub o-auth token with the permission "public_repo". This then has to be provided to Travis-Ci (via the dashboard) as the environment variable `GH_TOKEN`.

### DIFF
--- a/bin/rose-make-docs
+++ b/bin/rose-make-docs
@@ -37,7 +37,7 @@
 #         errors.
 #     --debug
 #         Run `make` with the --debug option.
-#     --default-version=VERSION
+#     --default-version VERSION
 #         By default the current version is symlinked as the default version,
 #         provide an alternative version to override this.
 #         If set to 'none' then the default version will remain unchanged.
--- a/bin/rose-make-docs
+++ b/bin/rose-make-docs
@@ -40,6 +40,7 @@
 #     --default-version=VERSION
 #         By default the current version is symlinked as the default version,
 #         provide an alternative version to override this.
+#         If set to 'none' then the default version will remain unchanged.
 #
 # BUILD
 #     The format(s) to build the documentation in - default html.
@@ -278,11 +279,13 @@ if make ${DEBUG} -C "${SPHINX_PATH}" ${BUILDS} ${SPHINX_OPTS}; then
     # switching pane
     version_file "${DOCS_DIR}" > "${DOCS_DIR}/versions.json"
     # symlink this version as the default
-    (
-        cd "${DOCS_DIR}"
-        rm "${DEFAULT_ALIAS}" 2>/dev/null || true
-        ln -s "${DEFAULT_VERSION:-$ROSE_VERSION}" "${DEFAULT_ALIAS}"
-    )
+    if [[ "${DEFAULT_VERSION}" != 'none' ]]; then
+        (
+            cd "${DOCS_DIR}"
+            rm -f "${DEFAULT_ALIAS}" 2>/dev/null
+            ln -s "${DEFAULT_VERSION:-$ROSE_VERSION}" "${DEFAULT_ALIAS}"
+        )
+    fi
     # symlink landing pages
     html_redirect "${DEFAULT_ALIAS}/html/index.html" 'doc/index.html'
     html_redirect "html/index.html" "doc/${ROSE_VERSION}/index.html"
--- a/etc/deploy-docs
+++ b/etc/deploy-docs
@@ -36,7 +36,7 @@ initial_deployment () {  # Create a new gh-pages branch from scratch.
     git clean -xf doc
 
     # build the documentation
-    rose make-docs html slides
+    rose make-docs html slides pdf
 
     # create the gh-pages branch
     git branch -D gh-pages || true
@@ -58,19 +58,20 @@ initial_deployment () {  # Create a new gh-pages branch from scratch.
 
 subsequent_deployment () {  # Add new version to documentation.
     # update gh-pages branch
-    git pull "${REMOTE}" gh-pages -f
+    git fetch "${REMOTE}" gh-pages -f
+
     git clean -xf doc
 
     # copy gh-pahes branch to a temporary directory
     TMP_DOCS_DIR="$(mktemp -d)"
-    git archive 'gh-pages' | (cd "${TMP_DOCS_DIR}" && tar -xf -)
+    git archive 'upstream/gh-pages' | (cd "${TMP_DOCS_DIR}" && tar -xf -)
 
     # append new documentation to the temporary directory
     ln -s "$TMP_DOCS_DIR" doc
-    rose make-docs html slides
+    rose make-docs html slides pdf
 
     # apply changes to gh-pages branch
-    git checkout gh-pages
+    git checkout -B gh-pages upstream/gh-pages
     rsync -av "${TMP_DOCS_DIR}/" .  # NOTE: the trailing slash is required
     rm -rf "${TMP_DOCS_DIR}"
     git add * -f
--- a/etc/travis-scripts
+++ b/etc/travis-scripts
@@ -143,7 +143,7 @@ build-docs () {
     if ! rose make-docs --venv --dev --strict clean html slides pdf \
             ${DEFAULT_VERSION:-} >out 2>&1; then
         # output is a bit long, only output it if the docs fail to build
-        cat out >2
+        cat out >&2
         exit 1
     fi
 
--- a/etc/travis-scripts
+++ b/etc/travis-scripts
@@ -1,0 +1,169 @@
+#!/usr/bin/env bash
+#-------------------------------------------------------------------------------
+# (C) British Crown Copyright 2012-8 Met Office.
+#
+# This file is part of Rose, a framework for meteorological suites.
+#
+# Rose is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Rose is distributed in the hope that it will be useful,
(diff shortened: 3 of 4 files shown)


etc/travis-scripts
```diff
@@ -0,0 +1,156 @@
...
+        tail -n 1)"
+    if [[ "${LATEST_TAG}" != "${TRAVIS_TAG}" ]]; then
+        DEFAULT_VERSION="--default-version none"
+    fi
+
+    # build the documentation
+    if ! rose make-docs --venv --dev --strict clean html slides pdf \
+            ${DEFAULT_VERSION:-} >out 2>&1; then
+        # output is a bit long, only output it if the docs fail to build
+        cat out
```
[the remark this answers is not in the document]
oliver-sanders: Contains both 1 & 2, we could cat to &2 but Travis doesn't distinguish so it makes no difference either way.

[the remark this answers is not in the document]
matthewrmshin: (Agreed. However it would look neater to redirect error to standard error before an `exit 1`. Sorry, but my brain is wired to expect this pattern.)

oliver-sanders:
De-conflicted.

matthewrmshin:
@sadielbartholomew please review, but do not merged until we have sorted out the admin.

bin/rose-make-docs
```diff
@@ -40,6 +40,7 @@
 #     --default-version=VERSION
 #         By default the current version is symlinked as the default version,
 #         provide an alternative version to override this.
+#         If set to 'none' then the default version will remain unchanged.
```
sadielbartholomew: Throughout ``travis-scripts`` ``--default-version none`` is used as opposed to ``--default-version=none`` as documented. I've tested these out & only the former syntax seems to work, so ideally we can update this docstring accordingly to prevent confusion.

etc/travis-scripts
```diff
@@ -0,0 +1,169 @@
...
+#-------------------------------------------------------------------------------
+set -eu
+set -o xtrace  # for ease of debugging travis builds
+shopt -s extglob
+
+
+export PATH="${PWD}/bin:${HOME}/fcm-master/bin:${HOME}/cylc-master/bin:$PATH"
+
+TMP_DOCS_DIR="${TMPDIR:-/var/tmp}/temp-docs-dir"
+DOCUMENTATION_FILES=!(doc|404*|CHANGES.md|_config.yml)
```
sadielbartholomew: Since (as briefly discussed in person) a ``bash -n travis-scripts`` vaildation is confused by this extended globbing line, consider appending a comment here identifying this as such (if I may be pedantic).

etc/travis-scripts
```diff
@@ -0,0 +1,169 @@
...
+    args=("$@")
+
+    # package managers
+    sudo apt-get -q update
+    sudo apt-get install -qy python-pip
+
+    # Cylc
+    if grep 'cylc' <<< "${args[@]}"; then
+        # install Cylc
+        wget 'https://github.com/cylc/cylc/archive/master.tar.gz' -O \
```
sadielbartholomew: Now we have migrated from the monolithic ``.yml`` file, it might be useful to move some/all of the site link definitions (or at least the bases e.g. ``https://github.com/[cylc/cylc/]``) from which we extract files etc. to the head of the file so that they are immediately explicit? Then again it might be clearer to define them within the functions they are used only... I'll leave it for consideration.

etc/deploy-docs
```diff
@@ -58,19 +58,20 @@ initial_deployment () {  # Create a new gh-pages branch from scratch.
 
 subsequent_deployment () {  # Add new version to documentation.
     # update gh-pages branch
-    git pull "${REMOTE}" gh-pages -f
+    git fetch "${REMOTE}" gh-pages -f
```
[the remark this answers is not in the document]
oliver-sanders: Because we branch from the upstream version by name:

```bash
git checkout -B gh-pages upstream/gh-pages
```

bin/rose-make-docs
oliver-sanders: Well spotted!

etc/travis-scripts
oliver-sanders: No idea why `bash -n` doesn't like the extglobs, perhaps it doesn't support them. For proper linting of shell script use something like shellcheck.

etc/travis-scripts
oliver-sanders: `https://github.com/cylc/cylc/` is only used once so I don't think we would have much to gain, of have I misunderstood?

sadielbartholomew: There is not much duplication, true, but the links are constants that are fundamental to the scripts, so putting them at the head of the script might be more appropriate & would make it clear where we are sourcing the codebase installations from. There is not much to gain now, admittedly, but may be useful for future reference. We can leave it be, though.

oliver-sanders:
Update this against #2235 if it is merged first.

oliver-sanders:
Time to create a metomi "actor" account which we can use to perform actions on the metomi repositories?

oliver-sanders:
As GitHub Pages builds from branches in the GitHub repository it is necessary to use a personal access token, no changes to Travis-CI will facilitate this.

Out of interest there is a script called [travis-sphinx](https://github.com/syntaf/travis-sphinx) which [some other projects](https://github.com/search?l=YAML&q=pip+install+travis-sphinx&type=Code) are using. It does roughly the same thing as this PR but pushes in force mode.

**Our Options:**

* Auto-deploy as in this PR.
* Auto-deploy using this code but building the docs elsewhere (no real change):
     * On a different CI platform;
     * On Institutional or personal hardware.
* Continue to build docs on institutional or personal hardware and deploy manually.
* Move documentation elsewhere e.g. readthedocs.
     * URLs wouls have to be re-directed.
     * Would tie us in with the readthedocs theme.
     * Would need to research if we could continue to use our slide builder there.

oliver-sanders:
Rebased onto latest Rose.

etc/travis-scripts
```diff
@@ -0,0 +1,169 @@
...
+deploy-docs () {
+    # produce a gh-pages branch ready for docs deployment
+    git checkout -B 'gh-pages' 'upstream/gh-pages'
+    rsync -aqv "${TMP_DOCS_DIR}/" .
+    git add -f ./*
+    git commit -q -m "${TRAVIS_TAG}"
+
+    # push gh-pages branch to GitHub where it will be auto-deployed to ghpages
+    set +o xtrace  # prevent leakage of the GH_TOKEN env var
+    git push "https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git" \
```
matthewrmshin: This will still reveal the `GH_TOKEN` variable if we do  `ps -f` listing while the `git push` command is running?

matthewrmshin:
Apologies. Push back for now as we cannot see how the Personal Access Token can be deployed properly.

kinow:
Hey :wave: ! Just passing by, noticed this part

>Every now and again we will want to remove older versions from the documentation, this can be done manually:
>
>    Check out the gh-pages branch and delete the relevant commit.
>    If the relevant commit is the first one, delete the relevant directory and squash the commit with the next one.

FWIW, I use [`ghp-import` for my site](https://github.com/davisp/ghp-import). As I am not using Jekyll, in order to deploy to GitHub pages, I have a small script that builds calling a static site generator for Python. This puts the contents in a folder like `./blog`... then I just execute:

```bash
$ ghp-import -n -m "Publishing kinoshita.eti.br" -p -r origin -b gh-pages blog
```

The important flags are `-p` to push to the branch after committing, `-r` to specify which git remote repo. The `gh-pages` is the branch, and `blog` is the local folder to push.

Have no idea if that'd be helpful. Maybe there's something in `ghp-import` that's helpful :wave:

oliver-sanders:
There is a sphinx-build wrapper thinggy called [travis-sphinx](https://github.com/syntaf/travis-sphinx) () which sits atop `ghp-import`. I looked into it but it seemed that implementing the required logic was probably simpler. Unfortunately `ghp-import` has the same issue with personal access tokens, we are waiting on GitHub to come up with a cleaner solution.

matthewrmshin:
I'll close this one down for now - until we find a way that does not require a personal access token.

oliver-sanders:
> I'll close this one down for now - until we find a way that does not require a personal access token.

Wahoo, there is now an alternative!

We can now use "deploy-keys", these are specific to a repository, to work for purpose they need to be given write permission.

Here's a GH action which does the job - https://github.com/peaceiris/actions-gh-pages#️-create-ssh-deploy-key

The permissions of the key are a little broader than I would desire (would be nice to be able to specify write for the `gh-pages` branch only, and also to ban force pushes), however, the tokens are repo specific and have a much less scary scope than personal access tokens.

I'm testing it on the [cylc-sphinx-extensions repo](https://github.com/cylc/cylc-sphinx-extensions/pull/19)

PR #2231, opened 2018-09-13
docs: add tip regarding nature of script setting

sadielbartholomew:
A user required clarification about the ``script`` task runtime setting, as used in the simple case of running a custom script under ``bin/``. They were unable to get a very basic suite running because the ``script`` setting for their tasks was set-up incorrectly & they could not see why.

They pointed out that the example provided in the Cylc Tutorial docs (Runtime -> Introduction -> 'The ``script`` Setting' & then the Runtime practicals) made it seem like this setting was for providing **just a script name** that would be run for the task in question, as one could easily assume from the setting name ``script``.

I've had a look & the tutorial would greatly benefit from clarification in this regard.

### DIFF
--- a/sphinx/tutorial/cylc/runtime/introduction.rst
+++ b/sphinx/tutorial/cylc/runtime/introduction.rst
@@ -95,6 +95,14 @@ We can also call other scripts or executables in this way, e.g:
        [[hello_world]]
            script = hello_world
 
+.. tip::
+
+   In the above example we run our custom executable by supplying just its
+   name to the ``script`` setting. This only works because we set up
+   the file so it can be executed in this way on the command line.
+
+   This is not the case in general. The ``script`` setting is for
+   *bash instructions*, not the name of a file to run.
 
 .. _tutorial-tasks-and-jobs:
 


oliver-sanders:
I think this is likely to lead people astray as they are likely to think that they could do `python hello_world`.

Easiest solution is probably to explain in the previous slide that Cylc executes the `script` in a bash shell then provide a Python example (e.g. `python -c 'print "Hello World!"'`) to drive the point home.

matthewrmshin:
> (e.g. `python -c 'print "Hello World!"'`)

Should be:

`python -c 'print("Hello World!")'`

For Python 3 compat.

sadielbartholomew:
> I think this is likely to lead people astray as they are likely to think that they could do python hello_world (i.e. get in a twist over $PATH and $PYTHONPATH).

Well anticipated. Admittedly, I thought ``python hello_world`` would work, so it seems I do not quite understand myself the role of the two env variables. I'll look this up & try to understand.

Regardless the main concern here is the execution of a separate script *as kept under* *``bin/``* and how it could easily be misconstrued from the current tutorial description that the setting is just where you specify that script by name to run it, so the Python example you suggest may not be that helpful as it is just a code string on the command line instead of a script file. (Does that makes sense?)

Once I've got my head around the function of e.g. ``PYTHONPATH``, I'll try to update the tip accordingly. If it still seems unclear or downright unhelpful, I am happy to change to your example, but ideally we can keep the focus on script files.

>  probably to explain in the previous slide

I imagine you referred to sildes by mistake, but just to be clear this concerns the HTML (etc) docs only. The *slides* & hence training material remain unchanged.

oliver-sanders:
> `PYTHONPATH`

We don't need to go into Python specific stuff at this stage, the point is to convey that it's just a shell call.

> but just to be clear this concerns the HTML

The slides are generated from the same RST file.

sadielbartholomew:
> The slides are generated from the same RST file.

Ah true, my bad. I assumed that admonitions would not appear on the slides, as I thought I had checked whether the new tip would show up & it did not, though I must have mistakenly checked the current docs. One can't actually view the slides from a branch build it seems.

sadielbartholomew:
Getting a Travis CI failure of unknown origin, seemingly unrelated to the PR change:
```
Exception occurred:
  File "/home/travis/build/metomi/rose/venv/lib/python2.7/site-packages/sphinx/writers/html.py", line 897, in unknown_visit
    raise NotImplementedError('Unknown node: ' + node.__class__.__name__)
NotImplementedError: Unknown node: autosummary_table
```
Any ideas?

oliver-sanders:
> Any ideas?

None whatsoever. If we could take a peak at the Sphinx error file we should get a full traceback which might help point us at the source of the error.

Try cat'ing `sphinx-err*` in the Travis temp directory, I think the following should work:

```diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,5 +41,6 @@ jobs:
             rose test-battery --state=save -j 5 ||
             (echo -e "\n\nRerunning Failed Tests...\n\n";
             rose test-battery -v --state=save,failed -j 5)
-        - script: >
-            rose make-docs --venv --dev --strict clean linkcheck doctest html slides latexpdf
+        - script:
+            - rose make-docs --venv --dev --strict clean linkcheck doctest html slides latexpdf
+            - cat /tmp/sphinx-err*
```

Or does that need to be a post script or error script or whatever...

sadielbartholomew:
Thanks @oliver-sanders, I'll have a look-see at the error file & try to fix the failure when I get a chance. Otherwise consider this on hold.

sadielbartholomew:
(#2236 merged, so I have restarted all Travis CI jobs. After these pass this should be good to go, unless anyone has further feedback.)

oliver-sanders:
Taking a read through I think this tutorial is actually pretty clear. We could streamline it a little and remove a potentially misleading reference to running "other scripts" to avoid possible confusion with Python:

```diff
--- a/sphinx/tutorial/cylc/runtime/introduction.rst
+++ b/sphinx/tutorial/cylc/runtime/introduction.rst
@@ -50,9 +50,9 @@ The ``script`` Setting
    We tell Cylc *what* to execute when a task is run using the ``script``
    setting.
 
-   This setting is interpreted as a bash script. The following example defines a
-   task called ``hello_world`` which writes ``Hello World!`` to stdout upon
-   execution.
+   The ``script`` is run as a bash script, the following example defines a
+   task called ``hello_world`` which writes the text ``Hello World!`` to
+   stdout when run:
 
 .. code-block:: cylc
 
@@ -64,7 +64,7 @@ The ``script`` Setting
 
    If you do not set the ``script`` for a task then nothing will be run.
 
-We can also call other scripts or executables in this way, e.g:
+We can also call executables in this way, e.g:
 
 .. code-block:: cylc
 
@@ -77,9 +77,10 @@ We can also call other scripts or executables in this way, e.g:
 .. ifnotslides::
 
    It is often a good idea to keep our scripts within the Cylc suite directory
-   tree rather than leaving them somewhere else on the system. If you create a
-   ``bin`` sub-directory within the :term:`suite directory` this directory will
-   be added to the path when tasks run, e.g:
+   tree rather than leaving them somewhere else on the system.
+
+   If a ``bin`` directory is present within the :term:`suite directory`
+   it will be added to the ``PATH`` when the task is run:
 
 .. code-block:: bash
    :caption: bin/hello_world
@@ -95,15 +96,6 @@ We can also call other scripts or executables in this way, e.g:
        [[hello_world]]
            script = hello_world
 
-.. tip::
-
-   In the above example we run our custom executable by supplying just its
-   name to the ``script`` setting. This only works because we set up
-   the file so it can be executed in this way on the command line.
-
-   This is not the case in general. The ``script`` setting is for
-   *bash instructions*, not the name of a file to run.
-
 .. _tutorial-tasks-and-jobs:
 
 Tasks And Jobs
```

sadielbartholomew:
Since you think this does not need amending I will close this. If we get any other users who have a similar misunderstanding, we should reconsider implementing something similar.

sadielbartholomew:
Since you think this does not need amending I will close this. If we get any other users who have a similar misunderstanding, we should reconsider implementing something similar.

PR #2234, opened 2018-09-21
Add one more way to identify localhost string

matthewrmshin:
`socket.gethostbyname_ex(target)` can give different answer to both
`socket.gethostname` and `socket.getfqdn`.

This is the case in my *new* desktop environment where a good long host name is not being recognised as a string for `localhost`.

### DIFF
--- a/lib/python/rose/host_select.py
+++ b/lib/python/rose/host_select.py
@@ -27,7 +27,8 @@ from rose.reporter import Reporter, Event
 from rose.resource import ResourceLocator
 import shlex
 import signal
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Do we need to replicate this `socket.gethostbyname_ex` approach in `remote.py`?

matthewrmshin:
When you said `remote.py`... DId you mean in `cylc.hostuserutil`? If so, it is already the case. It is Rose that has not been kept up to date.

oliver-sanders:
Argh, didn't spot which repo this was on!

PR #2235, opened 2018-09-24
rose ana test: remove dependency on pcregrep

matthewrmshin:
Replicate functionality with a small Python script.

(Background. My new desktop environment does not have `pcregrep` installed by default.)


oliver-sanders:
Test failure in the documentation unrelated.

PR #2237, opened 2018-09-26
Fix for path functions in grepper

stevewardle:
We recently made an adjustment to the way we interpret the KGO database entries produced by `rose_ana` tasks.  As a result sometimes the compared files include sym-links, and unfortunately the code in the current analysis functions calls `os.path.realpath` on the files.  The causes the sym-links to be followed and causes problems for our updating processes.  On closer inspection there's no reason why this needs to be `realpath`, so we change it here to use `abspath` instead (which is essentially the same, save for the sym-link following part)


stevewardle:
Not really... if you take a look at the UM ticket no. 4325 you'll see for the time being we've had to override `grepper.py` in our source and it has gone through testing there (on live KGO updates)

matthewrmshin:
Fair enough.

PR #2239, opened 2018-10-04
Reduce Rose-Cylc interaction part 1

matthewrmshin:
This is mainly a companion change for cylc/cylc#2693:
* It recognises that the `--host=HOST` option for most Rose-wrapped-Cylc commands should now be redundant.
* It uses and loads the suite contact file to determine whether a suite is running or not.
* It finds suite contact files to list running suites.
* It modifies `rose suite-gcontrol --all` to launch `cylc gscan` (instead of launching many `cylc gui` which can cause a machine to run out of resource).
* It modifies `rose suite-scan` to simply invoke `cylc scan` directly.
* It writes (Rose) automatic and custom environment variables to head of `suite.rc`.
  * Introduce a new `ROSE_SITE` environment/jinja2 suite variable + `site=SITE` setting - useful for site portable suites.


lib/python/rose/config_processors/jinja2.py
```diff
@@ -63,6 +77,12 @@ def process(self, conf_tree, item, orig_keys=None, orig_value=None, **_):
                 except UnboundEnvironmentVariableError as exc:
                     raise ConfigProcessError([s_key, key], node.value, exc)
                 tmp_file.write(self.ASSIGN_TEMPL % (key, value))
+            environ = kwargs.get("environ")
+            if environ:
+                tmp_file.write('[cylc]\n')
+                tmp_file.write('    [[environment]]\n')
```
oliver-sanders: Why not `[runtime][root][environment]`

matthewrmshin: The environment variables from `rose-suite.conf` are currently implemented as environment variables for the suite server program, not for tasks. So `[cylc][environment]` offers the best backward compatibility (and hence less chance of breakage). We can decide to do `[runtime][root][environment]` as well in the future, but I am not sure I want to risk it at this stage.

etc/rose.conf.example
```diff
@@ -19,6 +19,9 @@ meta-path=DIR1[:DIR2[:...]]
 #
 # URL to Rose documentation.
 rose-doc=http://metomi.github.io/rose/
+#
```
oliver-sanders: This will create a leading blank line in the docs.

oliver-sanders:
Code looks good, a good bit of waffle removed. Functionality testing to follow.

etc/rose.conf.example
matthewrmshin: Should now be fixed.

matthewrmshin:
Just added another changeset to remove more unused logic.

lib/python/rose/suite_engine_procs/cylc.py
```diff
@@ -130,19 +130,71 @@ def cmp_suite_conf(
...
+        )
+
+    def get_running_suites(self):
+        """Return a list containing the names of running suites."""
+        rootd = os.path.join(os.path.expanduser('~'), self.SUITE_DIR_REL_ROOT)
+        sub_names = ['.service', 'log', 'share', 'work', self.SUITE_CONF]
+        items = []
+        for dirpath, dnames, fnames in os.walk(rootd, followlinks=True):
+            if dirpath != rootd and any(
+                    name in dnames or name in fnames for name in sub_names):
```
sadielbartholomew: Is it perhaps more Pythonic to do ``any(name in dnames + fnames for ...)``?

lib/python/rose/suite_engine_procs/cylc.py
```diff
@@ -130,19 +130,71 @@ def cmp_suite_conf(
...
-        """Launch control GUI for a suite_name running at a host."""
-        if not self.is_suite_registered(suite_name):
-            raise SuiteNotRegisteredError(suite_name)
-        if not host:
-            host = "localhost"
-        environ = dict(os.environ)
-        if engine_version:
-            environ.update({self.get_version_env_name(): engine_version})
-        fmt = r"nohup cylc gui --host=%s %s %s 1>%s 2>&1 &"
+    def gcontrol(self, suite_name, args=None):
```
sadielbartholomew: ``gcontrol()`` & ``gscan()`` are now largely duplicates. Can we create a separate method e.g. ``launch_gui()`` for consolidation?

matthewrmshin:
I'll leave `t/rose-suite-run/{26,27,28}-*.t` as-is. Refactors of this nature should be done in separate PRs. (And hopefully, these stuffs will be removed from Rose in the not-so-distant future.)

I'll also leave `--host=localhost` for safety. It should, in theory, forces `cylc run --host=localhost`, which is most desirable for testing purpose. (We don't want the suite to run on configured suite hosts.)

lib/python/rose/suite_engine_procs/cylc.py
matthewrmshin: It is nicer syntax certainly. Not sure whether the extra `+` operation will have a penalty against the `or` short circuit logic. Not likely to have any impact here. I'll change the syntax just for the readability.

lib/python/rose/suite_engine_procs/cylc.py
matthewrmshin: Fair enough.

matthewrmshin: Updated with a service function.

lib/python/rose/suite_engine_procs/cylc.py
oliver-sanders: I don't think `or` and `+` will do the same thing here?

```python
>>> a = [1,2,3]
>>> b = [4,5,6]
>>> [x for x in a or x in b]
[1, 2, 3]
>>> [x for x in a + b]
[1, 2, 3, 4, 5, 6]
```

matthewrmshin: The original logic is more like:

```python
any(x in a or x in b for x in x_list)
# which should give the same results as:
any(x in a + b for x in x_list)
```

oliver-sanders:
This has been broken by https://github.com/cylc/cylc/pull/2759, re-starting Travis-CI build.

matthewrmshin:
OK. I'll take a look at the failed tests.

matthewrmshin:
I am out of time. I have fixed 2 of 3 failed tests, but am unable to get the tutorial forecasting suite run to pass in my environment at the moment.

oliver-sanders:
Taking a quick look it would appear the test is failing as it was taking advantage of the `cylc run DIR" behaviour (which was actually quite useful in this case):

`cylc run "${TEST_DIR}/cylc-forecasting-suite" ...`

I'll see if I can put up a fix to your branch.

oliver-sanders:
Looks like the `./t/rose-suite-restart/00-no-run.t` failure requires attention.

matthewrmshin:
`./t/rose-suite-restart/00-no-run.t` failure should now be fixed.

PR #2241, opened 2018-10-08
docs: minimal styling to incorporate logo & favicon

sadielbartholomew:
Customise the Read the Docs theme colours (mainly in the sidebar) so our documentation stands out a little, especially as the theme is becoming ever-more popular so docs often have the same style.

Not important, of course, but it completes the professional and polished feel of the docs & looks cool (see for example the Dask docs). Moreover, this was a nice web dev exercise for myself.

I went with a theme based on the blue & yellow colours of from the Rose logo (see also: Ikea, Leeds United FC etc.), but we can always debate what looks best & edit the colour codes. This choice does not clash with the colours of the code highlighting, so keeps things clean & clear.

I could get the Rose logo incorporated in the sidebar header in this PR too?


oliver-sanders:
> docs often have the same style

I kinda like that, look at all the great projects using the RTD theme (un-modified). The documentation used to be colour themed to the rose logo, there is a reason we changed that (yuck)!

We aught to get the Rose logo changed at some point, I thought about proposing something back when I was doing the Rose Docs but ran out of time, something which reflects the Cylc logo might be nice...

> I could get the Rose logo incorporated in the sidebar header in this PR too?

This might be a good idea, I did have a go but couldn't quite get it to "sit right", perhaps a more neutral background colour (https://sphinx-doc-zh.readthedocs.io/en/latest/config.html#confval-html_logo)

...

When we overhaul the Cylc docs it would be nice to have the two docs in the same format but it will be important for users to be able to tell between them. I'd suggest just changing just the colour of the box in the top left (to something neutral which better suits the logo).

sadielbartholomew:
Fair enough, it is all a matter of preference & taste, though I would say a lot of docs looking identical would frustrate me as trawling through docs, given their detailed technical nature, is monotonous enough without identical style from system to system.

I would encourage you to at least build my proposal as I think it looks nice & certainly isn't 'yuck'. But maybe a very subtle approach as you suggest is better. I'll wait for @matthewrmshin to give his opinion before I make any amendments in line with your comment.

wxtim:
Has anyone tried running these through a colour blindness simulator?

sadielbartholomew:
> I'd suggest just changing just the colour of the box in the top left (to something neutral which better suits the logo).

@oliver-sanders I'll get this PR changed in line with your suggestion, since you created the beautiful  new docs so I'm happy to grant you preference over the style. I'll also see if I can, without too much difficultly, get the current logo incorporated in a subtle way (I may just see if I can track down the font used which I recognise as a standard one & apply it directly to the sidebar header text).

@wxtim, good idea considering colour blindness. It would be premature to look into it until we are agreed on the extent of styling override to apply, but once there is that agreement we can check.

wxtim:
I liked this: I tried a colour-blindness simulator on the image and I had to go and find an Ishihara test to sit next to it to show that the app was working, the documentation looked so similar to what it does anyway...
thumbnail_Screenshot_20190327-123747

sadielbartholomew:
> I tried a colour-blindness simulator on the image

@wxtim very nice! The current docs are pretty bland colour-wise, being largely un-customised from the Read the Docs theme, which is so standard nowadays I would be extremely surprised to learn has problems relating to accessibility. So I do not think there is much concern for the case of the Rose docs as they are or as they would be with this change, but we should definitely make use of such tools when we are developing our Web UIs.

wxtim:
@oliver-sanders - might this be an easy win to review and merge?

PR #2247, opened 2018-11-01
fix doctests

oliver-sanders:
Command was not updated after it was renamed causing the doctests to always get skipped.


matthewrmshin:
(Will merge after tests complete.)

PR #2250, opened 2018-11-02
docs: add conda recipe

oliver-sanders:
Add conda environment file for installing the Python dependencies for the rose documentation builder. This is more documentation than functional change.

* Keep it simple for now, create a proper environment for Rose later.
* It might be possible to install other dependencies but these would include graphviz, latex, etc which would be a little heavy handed.
* It would be possible to add an auto-installer similar to the `--venv` option but that doesn't really seem necessary.

@matthewrmshin I've put this on next release as it's non functional and should be pretty quick to review. Feel free to bump if not.

### DIFF
--- a/bin/rose-make-docs
+++ b/bin/rose-make-docs
@@ -65,6 +65,30 @@
 #     `linkcheck`
 #         Checks external links.
 #
+# SOFTWARE ENVIRONMENT
+#     The software dependencies can be found by running the command::
+#
+#        $ rose check-software --docs
+#
+#     Rose provides two ways of installing the Python dependencies for the Rose
+#     documentation builder:
+#
+#     Virtual Environment:
+#         Rose will automatically build a Python virtual environment
+#         when the --venv flag is used with this command. The virtual
+#         environment will be created in rose/venv and will be destroyed
+#         after use. Use the --dev flag to prevent the environment being
+#         destroyed for future use. Example::
+#
+#            $ rose make-docs --venv --dev html  # create and keep a virtualenv
+#
+#     Conda
+#         An environment file for creating a conda env can be found in
+#         etc/rose-docs-env.yaml. Example usage::
+#
+#            $ conda env create -f etc/rose-docs-env.yaml  # create conda env
+#            $ source activate rose-docs  # activate conda env
+#            $ rose make-docs html  # make docs as normal
 #-----------------------------------------------------------------------------
 set -e
 set -o pipefail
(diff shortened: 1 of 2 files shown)


bin/rose-make-docs
```diff
@@ -65,6 +65,30 @@
...
+#         after use. Use the --dev flag to prevent the environment being
+#         destroyed for future use. Example::
+#
+#            $ rose make-docs --venv --dev html  # create and keep a virtualenv
+#
+#     Conda
+#         An environment file for creating a conda env can be found in
+#         etc/rose-docs-env.yaml. Example usage::
+#
+#            $ conda env create -f etx/rose-docs-env.yaml  # create conda env
```
sadielbartholomew: Whilst just documenting (& as an example), for future conda support I should point out:
1. Typo: should be ``etc/``
2. Perhaps append ``-p $ROSE_HOME`` to this command to ensure the file search path is tethered?

bin/rose-make-docs
```diff
@@ -65,6 +65,30 @@
...
+#         when the --venv flag is used with this command. The virtual
+#         environment will be created in rose/venv and will be destroyed
+#         after use. Use the --dev flag to prevent the environment being
+#         destroyed for future use. Example::
+#
+#            $ rose make-docs --venv --dev html  # create and keep a virtualenv
+#
+#     Conda
+#         An environment file for creating a conda env can be found in
+#         etc/rose-docs-env.yaml. Example usage::
```
sadielbartholomew: Isn't .yml the (standard) extension for YAML files?

oliver-sanders: Both `.yml` and `.yaml` are valid.

*Historical sidenote, quite a lot of formats are like that e.g. JPEG (`.jpg`, `.jpeg`), HTML (`.htm`, `.html`) the reason being that once upon a time not all systems permitted file extensions of more than three characters.*

bin/rose-make-docs
oliver-sanders: 1. Yep
2. Dunno, not all users will have `ROSE_HOME` defined in their environment (within rose commands it is ok to use such environment variables as they are provided by `bin/rose`). Because this command is illustrative rather than functional I think a relative path is OK, I'm not sure what the alternative would be?

bin/rose-make-docs
matthewrmshin: DOS!

bin/rose-make-docs
sadielbartholomew: Fair enough!

bin/rose-make-docs
```diff
@@ -233,43 +257,13 @@ if ! rose-check-software --doc >/dev/null; then
...
-                [Nn])
-                    echo
-                    echo 'Unable to build documentation.'
-                    exit 1
-                    ;;
-            esac
-        done
+        echo 'For information on building a Python environment for the Rose'
+        echo 'documentation builder see the "Software Dependencies" section'
+        echo 'of the help page for this command.'
```
matthewrmshin: Should redirect errors and warnings to STDERR.

PR #2253, opened 2018-11-05
rosie client: filter out urllib3 warnings

matthewrmshin:
(Similar to cylc/cylc#2688.)

### DIFF
--- a/lib/python/rosie/ws_client.py
+++ b/lib/python/rosie/ws_client.py
@@ -29,6 +29,7 @@ import json
 from multiprocessing import Pool
 import requests
 import shlex
+import sys
 from time import sleep
 
 from rosie.suite_id import SuiteId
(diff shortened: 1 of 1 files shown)


lib/python/rosie/ws_client.py
```diff
@@ -170,6 +171,27 @@ def _get(self, method, return_ok_prefixes=False, **kwargs):
...
+        # except to annoy or confuse users.
+        if sys.version_info < (2, 7, 9):
+            import warnings
+            try:
+                from requests.packages.urllib3.exceptions import (
+                    InsecurePlatformWarning)
+            except ImportError:
+                pass
+            else:
+                warnings.simplefilter("ignore", InsecurePlatformWarning)
```
oliver-sanders: In cylc/cylc#2688 the more primative `SecurityWarning` was used (which also covers `InsecureRequestWarning`), is this not needed here?

matthewrmshin: I copied the latest code in Cylc. We don't need the `InsecureRequestWarning` here because we don't have the issue where we have to call the server without verifying its SSL certificate.

PR #2255, opened 2018-11-07
commented out `modify_bg` statements

wxtim:
This is rather speculative - but I'm hoping that it might fix the RHEL7 issue
@matthewrmshin  - could you give it  a go?


wxtim:
.

PR #2257, opened 2018-11-09
Fix tests broken by cylc/cylc#2781 and cylc/cylc#2849

matthewrmshin:
Suppress debug level log entries in validate commands by not running command in verbose mode.

Modify tests to expect background `localhost` jobs to store host name entry as `hostname -f`.

See also: cylc/cylc#2781 and cylc/cylc#2849

### DIFF
--- a/etc/tutorial/consolidation-tutorial/.validate
+++ b/etc/tutorial/consolidation-tutorial/.validate
@@ -1,1 +1,1 @@
-cylc validate "$TUT_DIR" -v -v --icp=2000
+cylc validate "$TUT_DIR" --icp=2000
--- a/etc/tutorial/rose-stem/.validate
+++ b/etc/tutorial/rose-stem/.validate
@@ -1,4 +1,4 @@
 mkdir "${CYLC_RUN_DIR}/${REG}"
 echo -e '#!Jinja2\n{% set RUN_NAMES=["command_spaceship"] %}' > "${CYLC_RUN_DIR}/${REG}/suite.rc"
 cat "$TUT_DIR/rose-stem/suite.rc" >> "${CYLC_RUN_DIR}/${REG}/suite.rc"
-cylc validate -v -v "${CYLC_RUN_DIR}/${REG}" -s "SOURCE_SPACESHIP=foo"
+cylc validate "${CYLC_RUN_DIR}/${REG}" -s "SOURCE_SPACESHIP=foo"
--- a/etc/tutorial/rose-suite-tutorial/.validate
+++ b/etc/tutorial/rose-suite-tutorial/.validate
@@ -1,3 +1,3 @@
-cylc validate "$TUT_DIR" --icp=2000 -v -v
+cylc validate "$TUT_DIR" --icp=2000
 rose metadata-check -C "$TUT_DIR/app/forecast/meta"
 rose macro -C "$TUT_DIR/app/forecast" -V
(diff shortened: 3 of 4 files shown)


matthewrmshin:
@sadielbartholomew Please sanity check. This PR fixes the test battery failures that we are currently seeing on master and other PRs.

sadielbartholomew:
Sure. I've just started my local test battery.

etc/tutorial/consolidation-tutorial/.validate
```diff
@@ -1 +1 @@
-cylc validate "$TUT_DIR" -v -v --icp=2000
```
oliver-sanders: I think the verbose mode is actually functional here as the stderr is compared to `/dev/null` causing the test to fail on deprecated syntax.

matthewrmshin: Correct. Do we really need verbose mode here for all the irrelevant debug statement? Or do we just want a pass/fail validation?

matthewrmshin: cylc/cylc#2854 should improve this. `cylc validate` will now issue deprecation warnings for suite definitions and user configuration at normal verbosity level.

oliver-sanders: In this case it would be better just to exclude the site config as that isn't what we are testing anyway:

```bash
CYLC_CONF_PATH= cylc validate "$TUT_DIR" -v --icp=2000
```

oliver-sanders: **Followup:** It would appear that Cylc debug items go to stderr now which seems erroneous and breaks the "fail on warning" nature of `cylc validate` these tests were relying on.

matthewrmshin: Things considered. There is no right or wrong, apparently.
* `pycodestyle` and `pylint` print diagnostics to STDOUT, badness and all.
* `tidy` (my favourite linting utility for more than a decade) prints diagnostics to STDERR (and the prettified version of your original doc to STDOUT). I consider this the most Unix-like behaviour (although it confused some first time users).
* `rose.reporter` sends error-warning to STDERR, and info-debug to STDOUT.
  * (This is the closest to what you want?)
  * While you can *force* Python's `logging` to do the same, it is definitely not its natural way of working - because it does level-greater-than but does not really do level-range. Using the `logging` module, all I can do is choose the most logical location to send diagnostics to:
    * For a terminal based utility, the STDERR is the most logical location so we can reserve STDOUT for verbosity-agnostic output. (It almost behaves like `tidy` - if only it also prints the prettified version of the `suite.rc`, which it doesn't at the moment!)
    * For the suite server program, the suite log file is the most logical location.

oliver-sanders: > it is definitely not its natural way of working

Eh?

Currently our logging levels are:

1. Debug
2. Info
3. Warning
4. Error
5. Critical

I would naturally assume:
    * level <= 2: stdout
    * level >2: stderr

The usage of the debug level in Cylc is more akin to supplemental info. I would have thought that the two camps (print debug to stdout or stderr) would largely be divided by whether or not the output of the program needs to be parsed, in the case of a linter you probably don't want debug polluting the report, in the case of a log file, it's going to anyway so doesn't really make a difference.

matthewrmshin: Do you still want to pursue more change to this now that #2854 is merged?

oliver-sanders: Perhaps, I might take a look into `logging` at some point but not now.

PR #2258, opened 2018-11-09
Simplify logic for app_mode conditional assignment

sadielbartholomew:
Self-explanatory. Trivial hence lone reviewer sufficient.

### DIFF
--- a/lib/python/rose/app_run.py
+++ b/lib/python/rose/app_run.py
@@ -346,11 +346,8 @@ class AppRunner(Runner):
         self._poll(conf_tree)
 
         # Run the application or the command.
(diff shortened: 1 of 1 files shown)


sadielbartholomew:
Struggling to get the Travis CI ``.2`` Job to pass after several restarts, but given similar issues on another PR I am reviewing I think this is unrelated.

matthewrmshin:
Sorry, this one has totally gone off my radar.

I cannot see what is really improved by this change. If anything, it appears to change the logic slightly - as it is no longer testing for `... is None`.

Is there a particular linter that disagrees with the original code or something?

PR #2259, opened 2018-11-12
macro: document optional_config_name feature

oliver-sanders:
Add documentation for a feature quietly slid in way back (#1915).

@r-sharp pinging you in on this one, do these docs make sense?
@matthewrmshin don't allow this to be a release blocker, bump to soon if we can't get it reviewed.

### DIFF
--- a/lib/python/rose/macro.py
+++ b/lib/python/rose/macro.py
@@ -199,6 +199,24 @@ class MacroBase(object):
         ...    def report(self, config, meta_config=None):
         ...        # Perform some analysis of the config but return nothing.
         ...        pass
+
+        Keyword arguments can be used, ``rose macro`` will prompt the user to
+        provide values for these arguments when the macro is run.
+
+        >>> def validate(self, config, meta_config=None, answer=None):
+        ...     # User will be prompted to provide a value for "answer".
+        ...     return self.reports
+
+        There is a special keyword argument called ``optional_config_name``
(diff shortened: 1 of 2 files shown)


lib/python/rose/macro.py
```diff
@@ -199,6 +199,24 @@ class MacroBase(object):
...
+        ...     return self.reports
+
+        There is a special keyword argument called ``optional_config_name``
+        which is set to the name of the optional configuration a macro is
+        running under, or ``None`` if no optional configuration is being used.
+
+        >>> def report(self, config, meta_config=None,
+        ...            optional_config_name=None):
+        ...     if optional_config_name:
+        ...         print ('Macro is being run using the "%s" '
```
matthewrmshin: Please remove space between `print` and `(`.

r-sharp:
Hi Oliver,

That looks okay to me although (and it’s probably a syntax difference for you guys compared with we users….)

Where you say “… a macro is running under, or ``None`` if no optional configuration is being used”

I would have phrased it
… a macro is running on (or maybe … a macro is processing)
And the last bit I would have said “ ,or ‘’None’’ if the default configuration is being used” as to my way of reading it saying “if no optional configuration is being used” implies an app with no optional configs, not a config with no optional element applied.

It’s probably just the way I think about these things, not having to deal with them regularly, compared to you.

HTH
Roddy

oliver-sanders:
Done.

PR #2261, opened 2018-11-16
Execute python2 instead of python

matthewrmshin:
For our final feature release based on Python 2.


matthewrmshin:
(Similar to cylc/cylc#2676.)

matthewrmshin:
@oliver-sanders As discussed earlier. :snake: :snake: One review enough?

oliver-sanders:
> One review enough?

Hopefully, this should cover all functional cases! If we were really going for it we could add something to Travis `alias python=false`. I'll leave you to decide.

PR #2265, opened 2018-11-28
WIP: Extend rose_bunch to be able to run over all permutations of arguments

ColemanTom:
This was an idea I was thinking might be worth adding. If you are interested in it, then I would add tests and documentation for it.

The basic idea of this is that it would be useful to be able to provide a few arguments to rose_bunch, as is done currently, and be able to run all permutations of them.

For example. You need to prepare data for a bunch of timesteps for a variety of domains and level types. Each command may not take that long to run, but if you do them serially over just forecast hour, it starts to add up. Instead just tell rose_bunch that these are the things I want to do, and I want all combinations of them.

```
mode=rose_bunch

[bunch]
all_permutations=True
command-format=exe %(fhr)s $(domain)s %(type)s

[bunch-args]
fhr=1 2 3 4 5 6
domain=AUS QLD
type=press slv
```

versus spelling it out in full like:

```
mode=rose_bunch

[bunch]
all_permutations=True
command-format=exe %(fhr)s $(domain)s %(type)s

[bunch-args]
fhr=1 2 3 4 5 6 1 2 3 4 5 6 1 2 3 4 5 6 1 2 3 4 5 6
domain=AUS AUS AUS AUS AUS AUS AUS AUS AUS AUS AUS AUS FULL FULL FULL FULL FULL FULL FULL FULL FULL FULL FULL FULL
type=press press press press press press slv slv slv slv slv slv press press press press press press slv slv slv slv slv slv
```

Similar ideas could be `fhr` by `ensemble_member`.

What I'm proposing is the simplest implementation of a permutation creator that I can think of.


TomekTrzeciak:
Seems to be exactly what @oliver-sanders asked for in [#2218 comment](https://github.com/metomi/rose/pull/2218#discussion_r207510326).

matthewrmshin:
This change makes sense, but we cannot afford to break backward compatibility for the `all_permutations=True` option (even though it is a bit wrong).

Can we introduce a new more flexible setting and deprecate the `all_permutations=True` (with backward compatibility support)? E.g.:
* `permutations=full` == new mode proposed by this PR.
* `permutations=zip` == old `all_permutations=True`.
* Something else.

oliver-sanders:
>  Seems to be exactly what @oliver-sanders asked for in #2218 comment.

I was just thinking the same...

I agree with Matt but we can make this more generic to "cover all" possible argument formats. We might as well just map straight onto the `itertools` methods:

`argument-mode=product`  (I think what this PR is proposing)
`argument-mode=permutations`
`argument-mode=combinations_with_replacement`
`argument-mode=izip`
`argument-mode=zip-longest` (suggested in https://github.com/metomi/rose/pull/2218#discussion_r207573315)

Something along the lines of:

```python
try:
    args = itertools.getattr([self.BUNCH_SECTION, "argument-mode"])(*bunch_args_values))
except AttributeError:
    args = bunch_args_values
```

ColemanTom:
Ok. I'll update accordingly. From a quick look, extra logic will be required to handle all those options.

For example

* `izip_longest` takes a fillvalue argument which https://github.com/metomi/rose/pull/2218#discussion_r207573315 suggests should be an empty string (default is None, but it is possible someone would want to use that as a real argument value).
* `permutations` would need to ensure there is only one argument. The length I'm guessing would be the length of that argument (i.e. the default option)
* `combinations_with_replacement` I'm not sure what to do with. What `r` value would be chosen? The length of the argument? Again, also needs a check to make sure there is only one list.

ColemanTom:
I'm not sure what to do with `permutations` and `combinations_with_replacement`. They can only really take one list, but will return a list of tuples. How would they map to a single argument? Perhaps I'm not grasping their use cases?

oliver-sanders:
Thanks for doing this :+1: 

#### Zips and Products

I think you've got the sensible cases covered for now:

* `izip`
* `iziplongest`
* `product`
* This should also work for chain but I have no idea why anyone would want that, perhaps a lazy way of turning parallelisation off.

#### Combinations

As you noted to get the permutation methods working we would require a method of specifying additional arguments. We could do this in a pythonic format then extract them with a simple regex:

`argument-mode=permutations(r=2)`
`argument-mode=islice(start=5, end=10, step=2)`

This can wait until someone actually requires this functionality:

#### Maps

The `*map` methods could be useful, I'd guess the main use case here is probably argument formatting:

```python
>>> tuple(itertools.imap(lambda x, y, z: '--myarg=%s' % ((x + y) * z), [1, 2, 3], [4, 5, 6], [7, 8, 9]))
('--myarg=35', '--myarg=56', '--myarg=81')
```

It's likely users would want to do this formatting in combination with an expansion so this would probably need to be an extra feature of rose_bunch. Again this can definitely wait.

TomekTrzeciak:
Exciting to see some more development on rose_bunch :smiley: 

Now I have an itch to scratch and make it work with job_runner and inter-command-instance dependencies. One could use it then for execution of DAGs. The ugly bit is how to specify the dependencies in rose_bunch friendly format. Could use something like this:
```ini
[bunch]
names = a b c x y z
dependencies = a,b,z c,x b,c,y
```
The comma separated sequences could mean a partial order as in `tsort` command. If `names` is not specified integer values of command instance could be used.

Interestingly, there is unused `import rose.job_runner` in `rose_bunch.py` from it's very inception. @arjclark, did you plan to use it at some point?

arjclark:
@TomekTrzeciak - It was certainly on my radar, yes.

oliver-sanders:
The long term plan is for Cylc to assume this functionality effectively replacing Rose Bush. Ideally when we have [prised apart the Cylc kernel and shell](https://github.com/cylc/cylc/wiki/Possible-Cylc-Futures#kernel-shell-divide) we should be able to submit a lightweight (pre-configured) Cylc kernel as a job which would provide an advanced solution to such requirements.

I'd bump discussion of the future of "bunching" functionality over to https://github.com/cylc/cylc/issues/2754.

ColemanTom:
I've done an update with tests and some documentation. I didn't try to add to the tutorial on rose-bunch though.

matthewrmshin:
@ColemanTom If you re-base your branch against upstream/master, this should be good again.

ColemanTom:
Ah, I should have done that first! Done, and thanks.

matthewrmshin:
@sadielbartholomew please review and sanity check.

t/rose-task-run/31-app-bunch.t
```diff
@@ -163,5 +163,59 @@ for KEY in $(seq 0 3); do
...
+# Testing ROSE_BUNCH_ARGUMENT_MODE_IZIP_LONGEST lengthens all bunch-args
+#-------------------------------------------------------------------------------
+APP=bunch_argument_mode_izip_longest
+#-------------------------------------------------------------------------------
+TEST_KEY_PREFIX=argument_mode_izip_longest
+FILE=$LOG_DIR/$APP/01/job.out
+for VALUES in 1,a,9 2,,23 3,, 4,,; do
+    IFS=, read -r -a ARGS <<< "$VALUES"
+    file_grep "$TEST_KEY_PREFIX-RUN-$ARG1-$ARG2-$ARG3" \
+        "\[INFO\] [0-9]*-[0-9]*-[0-9]*T[0-9]*:[0-9]*:[0-9]*+[0:9]* echo arg1: ${ARGS[0]:-} - arg2: ${ARGS[1]:-} - arg3: ${ARGS[2]:-}"\
```
sadielbartholomew: Ideally we can also split this long line, & the same for line 211 (& shorten the first line of 177).

t/rose-task-run/31-app-bunch.t
```diff
@@ -163,5 +163,59 @@ for KEY in $(seq 0 3); do
...
+#-------------------------------------------------------------------------------
+# Testing ROSE_BUNCH_ARGUMENT_MODE_IZIP_LONGEST lengthens all bunch-args
+#-------------------------------------------------------------------------------
+APP=bunch_argument_mode_izip_longest
+#-------------------------------------------------------------------------------
+TEST_KEY_PREFIX=argument_mode_izip_longest
+FILE=$LOG_DIR/$APP/01/job.out
+for VALUES in 1,a,9 2,,23 3,, 4,,; do
+    IFS=, read -r -a ARGS <<< "$VALUES"
+    file_grep "$TEST_KEY_PREFIX-RUN-$ARG1-$ARG2-$ARG3" \
```
sadielbartholomew: I had to investigate this test, since all of the sub-tests were named identically, suggesting there might be an issue. Firstly, the arguments to the test name are those from the ``izip``-related test instead of the ones tested:

```suggestion
    file_grep "$TEST_KEY_PREFIX-RUN-${ARGS[0]:-}-${ARGS[1]:-}-${ARGS[2]:-}" \
```

However, besides this I do get the impression that these ``izip``-related tests do not test the intended behaviour; when I arbitrarily change the values provided to the ``VALUES`` in the ``for`` loop, the tests still pass. Could you please take another look at this & confirm or otherwise if this does as it should?

As a general comment, I am struggling to follow the intended behaviour for each sub-test for the two ``izip-*`` argument modes. If you had the time, it would be useful if you could add some brief comments to clarify the expected test outputs (for all three modes).

t/rose-task-run/31-app-bunch.t
ColemanTom: Would it be appropriate for me to shift the `\[INFO\] [0-9]*-[0-9]*-[0-9]*T[0-9]*:[0-9]*:[0-9]*+[0:9]*` into a different include file somewhere? It is a standard pattern for looking in log files at the info lines. It appears at least 16 times in the test suite (14 in this file, 2 in `37-app-bunch-rm-old.t`).

I am tempted to move it into the `test_header`, but, I don't think it tends to have known constants in there that could be commonly used across tests. For now I'll just put it as a constant at the top of this file itself, unless advised otherwise.

ColemanTom: Oh, the datetime component of that strings appears more, and in some other test files, it is `\\[INFO\\]....`

ColemanTom:
@sadielbartholomew Thanks for the feedback. I've done some updates which I think address your comments. Can you please have another look? Thanks.

t/rose-task-run/31-app-bunch.t
```diff
@@ -163,5 +170,72 @@ for KEY in $(seq 0 3); do
...
 #-------------------------------------------------------------------------------
+#-------------------------------------------------------------------------------
+# Testing ROSE_BUNCH_ARGUMENT_MODE_IZIP shortens all bunch-args to the length
+# of the shortest bunch-arg
+#-------------------------------------------------------------------------------
+APP=bunch_argument_mode_izip
+#-------------------------------------------------------------------------------
+TEST_KEY_PREFIX=argument_mode_izip
+FILE=$LOG_DIR/$APP/01/job.out
+ARG1=2
```
matthewrmshin: Change this to `ARG1=1` and the test will pass.

t/rose-task-run/31-app-bunch.t
sadielbartholomew: > It is a standard pattern for looking in log files at the info lines. It appears at least 16 times in the test suite

I agree it would be a good idea to move the pattern somewhere where it can be referenced to prevent duplication. It may be best to, instead of a solitary constant in the ``test_header`` as you are reluctant to add, creating an option to the ``file_grep()`` function that will adjust the behaviour of that test method accordingly in the common case of testing for log line macthes?

I'll get this PR reviewed as-is, without the above change. However, given it is distinct to the core of this PR, I suggest we get this merged as-is & then advise you to put in a separate PR for cleaning up the test suite as above, should you still wish to (it would be great if you did!).

ColemanTom:
Fixed. Sorry, I must have committed that test failure by accident when I was testing changing values.

PR #2267, opened 2018-11-30
Remove unreachable code in macro.py

kinow:
After the `if` and `else`, I think there's no way for the extra `return` to be used.


matthewrmshin:
Agreed. One review OK for this trivial change.

PR #2268, opened 2018-11-30
Name LocalCopyStatusError as exc, so that str(exc) works below

kinow:
There's a call to `str(exc)` right below the `try/catch`. I think it was copy/paste'd, as `str(exc)` would fail in runtime?

Found doing a quick analysis of the project after working on the pull request for coverage, and before preparing things for Melbourne; so apologies for not doing thorough testing.


matthewrmshin:
One review enough for this trivial change.

PR #2269, opened 2018-11-30
Batch of typos in a few files (includes exception messages)

kinow:
I think most typos are harmless, except that I updated a couple of variables that I believe are used for exception/logging. Not sure if some tests are expecting specific values - if so, once Travis points which ones need fixing, will update the commit accordingly :+1:


matthewrmshin:
1 review enough for this.

PR #2273, opened 2019-01-04
remove docs latex dependency

oliver-sanders:
When I started out writing the Cylc tutorial I though LaTeX diagrams might be useful, however, it turns out that `graphviz` and `svg` cover most use cases better and there is only one LaTeX diagram in the documentation.

This PR converts the one LaTeX diagram to an SVG (editable via Inkscape) removing the LaTeX dependency (except for PDF builds) and adds superior SVG graphic support where appropriate tools / extensions are available.

### DIFF
--- a/bin/rose-check-software
+++ b/bin/rose-check-software
@@ -315,8 +315,6 @@ def docs(check=check):
                outfile=2),
         check('graphviz-dot <dot>', command_template='-V',
                version_template='graphviz version ([^\s]+)', outfile=2),
-        check('tex', version_template=r'TeX ([^\s]+)'),
-        check('pdflatex', version_template=r'pdfTeX .*-.*-([^\s]+)'),
         # TODO remove Sphinx version dependency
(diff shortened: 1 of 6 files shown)


oliver-sanders:
Codacy complaining about use of subprocess calls. Should I add a `nosec` comment.

matthewrmshin:
If it is just for the `which` command, perhaps use (a copy of) this pure Python logic?

https://github.com/metomi/rose/blob/2536f97dfac96cc4b899ef473387dbf9900b263e/lib/python/rose/popen.py#L252-L264

(You don't need the first 2 lines because you are not dealing with absolute paths.)

matthewrmshin:
(It wasn't used in Cylc any more, but is still used in Rose (e.g. to determine if `fcm` and/or `svn` is installed or not).)

Need to do `from rose.popen import RosePopener` then call `RosePopener.which` or `RosePopener().which`. Or something like that.

PR #2274, opened 2019-01-07
pygments syntax: improve cylc lexer

oliver-sanders:
Support additional features of the Cylc suite.rc language to permit use in the Cylc documentation (cylc/cylc#2910).

Adds support for the following (see table in cylc/cylc#2752):
* Inter-Suite
* Include files
* xtriggers
* empy

Fixes:
* Line continuation characters in headings
   ```
    [foo, \
    bar]
    ```
* Cycle points in inter-cycle offsets

### DIFF
--- a/sphinx/ext/cylc_lang.py
+++ b/sphinx/ext/cylc_lang.py
@@ -175,16 +175,26 @@ class CylcLexer(RegexLexer):
         ],
 
         # Task inter-cycle offset for graphing:  foo[-P1DT1M]
+        # Legal formats: POINT, POINT[+-]OFFSET, OFFSET
         'intercycle-offset': [
-            include('integer-duration'),
+            include('cycle-point'),
+            include('integer-duration'),  # matches a subset of iso8601
             include('iso8601-duration'),
             (r'[\^\$]', INTERCYCLE_OFFSET_TOKEN),
             (r'\]', Text, '#pop')
         ],
 
+        # generic Cylc cycle point:  2000
+        'cycle-point': [
+            # validating the cycle point as a regex [effectively] requires
+            # knowledge of the cycling mode so is not *really* possible,
+            # also validating iso8601 datetimes is horrible.
+            (r'[\+\-]?[\d\:\-T]+(Z)?\b', INTERCYCLE_OFFSET_TOKEN)
+        ],
+
         # An integer duration:  +P1
         'integer-duration': [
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Automated testing for syntax files, hmmm, that might be a tricky one.

The simplest thing I can think of doing would be to come up with some devilishly convoluted examples to run through each editor in turn, then perform a visual diff on before and after screenshots.

oliver-sanders:
> If there is a way to test by giving the lexical analyser some suite.rc fragments and see what tokens come out on the other side

For `pygments`, *yes*.
For `vim`, *possibly*.
For `gedit` *god knows*.

See cylc/cylc#2915 for a discussion on testing approaches for syntax files.

oliver-sanders:
Have added standard reference files for testing lexers in https://github.com/cylc/cylc/pull/2916.

PS/ Spotted an edge case with `empy` (`@[if []]`).

sphinx/ext/cylc_lang.py
```diff
@@ -118,24 +127,37 @@ class CylcLexer(RegexLexer):
...
             include('parameterisation'),
+            (r'@\w+', GRAPH_XTRIGGER_TOKEN),
             (r'\w+', GRAPH_TASK_TOKEN),
+            (r'\!\w+', Other),
             (r'\s', Text),
             (r'=>', Operator),
-            (r'[\&\|\!]', Operator),
+            (r'[\&\|]', Operator),
             (r'[\(\)]', Punctuation),
             (r'\[', Text, 'intercycle-offset'),
```
sadielbartholomew: I have noticed one omission. Dependencies on specific cycle points e.g. ``graph = "red[2019] => blue"`` are supported as valid specification within square brackets in a graph string, where only intercycle offsets appear to be recognised by the lexer as-is.

sadielbartholomew: The good news is the lexer is now so water-tight it picked up on several missing markers in some of the Cylc documentation code excerpts. It's nearly there :clap:

oliver-sanders:
>  cycle point arithmetic has not been included

Arithmetic is fine but arbitrary whitespace is not supported.

sadielbartholomew:
> Arithmetic is fine but arbitrary whitespace is not supported.

Oh, fair enough, I simply used the example you provided in your original comment, assuming it was valid. Good!

PR #2275, opened 2019-01-07
Fix overzealous informational message about KGO in grepper class

stevewardle:
There's a bit of logic in the way the `grepper` class deals with the specification of the KGO files which leads to it spamming a message in an unhelpful way...

Currently the behaviour if the `kgo_file` input is completely omitted and the behaviour if it is provided but left completely empty are exactly the same.  As a result a `rose_ana` task which omits the setting because it doens't care about KGO at all (it may simply not use the concept of KGO) will print the message "No KGO files are present" for every single comparison made.  Clearly this is undesirable...

The solution is to reshuffle the logic so that the message is only printed in the case of the input being explicitly set to blank (with complete omission printing no messags at all)

### DIFF
--- a/lib/python/rose/apps/ana_builtin/grepper.py
+++ b/lib/python/rose/apps/ana_builtin/grepper.py
@@ -157,19 +157,18 @@ class SingleCommandStatus(AnalysisTask):
         # Parse the kgo index
         if kgo is not None:
             if kgo.strip() == "":
+                self.reporter("No KGO files are present")
                 kgo = None
             elif kgo.isdigit():
                 kgo = int(kgo)
                 if int(kgo) > len(self.files):
                     msg = "KGO index cannot be greater than number of files"
                     raise ValueError(msg)
+                self.reporter("KGO is file {0}".format(kgo + 1))
             else:
-                msg = "KGO index not recognised; must be a digit or blank"
+                msg = ("KGO index not recognised; should be either a digit or "
+                       "left blank")
                 raise ValueError(msg)
-        if kgo is not None:
-            self.reporter("KGO is file {0}".format(kgo + 1))
-        else:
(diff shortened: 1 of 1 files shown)


stevewardle:
@matthewrmshin  There's just one thing I'm trying to decide here... do you think the message itself is potentially misleading?  It says "No KGO files are present"... which I fear might sometimes come across to the user as saying that there were meant to be KGO files but none have been found.  The intent though is to convey that the _app_ has specified that there are no KGO files (and the output is just repeating this)

So perhaps it might usefully be changed to something more like  "App indicates none of the input files are KGO" or just "App specifies no KGO" or even "This test does not compare against KGO"... any thoughts?

matthewrmshin:
As discussed off line, perhaps just state that you are in KGO mode or not. Using the `diff` syntax as an analogy...

If you are in KGO mode, you will state that the expected output is from a KGO:

```diff
--- out1 (KGO)
+++ out2
...
```

If you are not in KGO mode, you will simply do:

```diff
--- out1
+++ out2
...
```

stevewardle:
> As discussed off line, perhaps just state that you are in KGO mode or not. 

I think that makes sense... ultimately the line stating that "No KGO is present" or similar is not really relevant or useful information; if one of the files is KGO it will say so, if not then it doesn't actually matter whether that's because the setting was blank or not specified (just that none of the files are KGO)

matthewrmshin:
@stevewardle is this ready for review or do you intend to add some tests as well?

stevewardle:
> @stevewardle is this ready for review or do you intend to add some tests as well?

Go for it, thanks... I'm not sure what value testing would add here as not much has actually changed

matthewrmshin:
@paulcresswell Please sanity check. Approve and merge if you can.

PR #2280, opened 2019-01-23
docs: patch hieroglyph to work with sphinx 1.8.0+

oliver-sanders:
Temporary fix to allow the documentation to be built with contemporary versions of Sphinx (note v2.0 is under development).

### DIFF
--- a/bin/rose-make-docs
+++ b/bin/rose-make-docs
@@ -177,12 +177,10 @@ venv-install () {
     venv-destroy
     virtualenv --python=python2.7 "${VENV_PATH}"
     venv-activate
-    # TODO remove Sphinx version dependency
-    # https://github.com/nyergler/hieroglyph/issues/148
-    pip install 'sphinx==1.7.9'
-    pip install sphinx_rtd_theme
-    pip install sphinxcontrib-httpdomain
-    pip install hieroglyph
+    pip install 'sphinx'
+    pip install 'sphinx_rtd_theme'
+    pip install 'sphinxcontrib-httpdomain'
+    pip install 'hieroglyph'
 }
 
 venv-deactivate () {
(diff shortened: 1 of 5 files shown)


bin/rose-make-docs
```diff
@@ -179,10 +179,10 @@ venv-install () {
     venv-activate
     # TODO remove Sphinx version dependency
```
sadielbartholomew: This TODO has just been addressed, so can be removed I think (:ballot_box_with_check:)?

sphinx/ext/hieroglyph_patch.py
```diff
@@ -0,0 +1,41 @@
...
+
+from sphinx.writers.html import HTMLTranslator
+
+from sphinx.ext.graphviz import html_visit_graphviz
+from sphinx.ext.autosummary import (autosummary_toc_visit_html,
+                                    autosummary_table_visit_html)
+from minicylc import MiniCylc
+
+
+def none(*args, **kwargs):
```
sadielbartholomew: Is this just left from development or does it have some mysterious purpose?

oliver-sanders: It's a dummy interface used below on lines 36 and 38.

oliver-sanders:
> Is that just something to do with my environment or setup?

Yes, make sure you are viewing the site through a web server (e.g. Apache) rather than viewing the file directly in the browser.

Explaination, the version selector make an AJAX call to load in a JSON file which contains the information about which versions and formats are available.

sphinx/ext/hieroglyph_patch.py
sadielbartholomew: Oh, sorry, I did check for that but missed them (can't have looked very well!)

PR #2285, opened 2019-01-25
docs: fix suite design-guide url

oliver-sanders:


matthewrmshin:
One review sufficient.

PR #2286, opened 2019-01-30
2to3 1st attempt on independent commands

wxtim:
Work in progress for Rose Python3 upgrade.

Test | Status | Notes | 
|---|---|---|
|docs| :heavy_check_mark: |
|__Rosie Apps__<br>(Deliberatly not fixed: waiting on @sadielbartholomew)|
|rosa-db-create| :heavy_multiplication_x: | PR #2288 - @sadielbartholomew |
|rosa-svn-post-commit| :heavy_multiplication_x: |PR #2288 - @sadielbartholomew|
|rosa-svn-pre-commit| :heavy_multiplication_x: | #2304|
|rose-ana| :heavy_multiplication_x:| #2304|
|rosie-checkout| :heavy_check_mark: ||
|rosie-create| :heavy_check_mark: ||
|rosie-delete| :heavy_check_mark: ||
|rosie-disco| :heavy_multiplication_x:|PR #2288 - @sadielbartholomew|
|rosie-graph| :heavy_multiplication_x:|PR #2288 - @sadielbartholomew|
|rosie-hello| :heavy_check_mark:||
|rosie-id| :heavy_check_mark: ||
|rosie-lookup| :heavy_multiplication_x:| #2304|
|rosie-ls| :heavy_multiplication_x:|partly :heavy_check_mark:, but the tests are reliant on rosie-disco|
|rosie.db| :heavy_multiplication_x: |PR #2288 - @sadielbartholomew|
|__Rose Standalone Apps__|
|rose-app-run| :heavy_check_mark:||
|rose-app-upgrade| :heavy_check_mark:||
|rose-cli-bash-completion| :heavy_check_mark:||
|rose-config| :heavy_check_mark:||
|rose-config-diff| :heavy_check_mark:||
|rose-config-dump| :heavy_check_mark:||
|rose-config-edit| :heavy_check_mark:||
|rose-date| :heavy_multiplication_x:| issue #2303  |
|rose-env-cat| :heavy_check_mark:||
|rose-help| :heavy_check_mark:||
|rose-host-select| :heavy_check_mark:||
|rose-macro| :heavy_check_mark:||
|rose-metadata-check| :heavy_check_mark:||
|rose-metadata-graph| :question: | #2301  |
|rose-metadata-gen| :heavy_check_mark:||
|rose-mpi-launch| :heavy_check_mark:||
|rose-namelist-dump| :heavy_check_mark:||
|rose-stem| :heavy_check_mark:||
|rose.c3| :heavy_check_mark:||
[...]

Issue #2271, opened 2018-12-18
rosie discovery service: migrate away from cherrypy


matthewrmshin:
[Tornado](https://www.tornadoweb.org/en/stable/) is the current framework of choice to migrate to, (but we may also want to look at [AIOHTTP](https://aiohttp.readthedocs.io/en/stable/).)

We need to do the following:
1. Find a quick route to migrate from Cherrypy to Tornado (or AIOHTTP) with minimal change. Making sure that we can run the service:
  a. As an ad-hoc server in a test environment.
  b. As a WSGI application in a system service environment.
2. Invest in further effort to make use of any useful features of the new framework, such as:
  a. Asyncio.
  b. Authentication services.
3. Migrate web service UI to use the same framework as Cylc 8 web UI.

Only point 1 above is essential at the moment. We should aim to do this in a modern version of Python 3.

See also cylc/cylc#2563.

PR #2288, opened 2019-02-04
rosie & rosa: migrate to Python 3(.6-.7) & Tornado

### LINKED ISSUE #2301: Convert Rose Metadata Graph to Python 3
rose metadata-graph was not fixed in https://github.com/metomi/rose/pull/2286 due to issues with pygraphviz.

sadielbartholomew:
Address point 1 of #2271, namely for the Rosie Discovery service, with minimal changes [:star:]:

* migrate the web server from CherryPy to Tornado;
* port necessary components (only) from Python from 2 to 3, not preserving backward compatibility, as an intermediate step, consisting of:
  * **modules** i.e. ``rosie.ws`` & its chain of undetachable module dependency imports:

    | ``rose.``                                                        | ``rosie.``                             | ``rosa.``                          |
    |----------------------------------------------------|-------------------------------|----------------------------|
    |  ``config`` <br> ``date`` <br> ``host_select`` <br> ``loc_handlers.svn`` <br> ``reporter`` <br> ``metadata_graph`` [close #2301] |  ``ws`` <br> ``suite_id`` <br>``lookup`` <br> ``ls`` <br> ``graph`` |  ``db_create`` <br> ``svn_post_commit`` <br> ``svn_pre_commit`` (not essential but fixed as a side effect) |

  * **tests**: relating to the Rosie Disco server & client:

    key: :heavy_check_mark: = passes locally.

    | ``t/rose...``                                                        | ``t/rosie...``                             | ``t/rosa...``                          |
    |----------------------------------------------------|-------------------------------|----------------------------|
    |  |  ``.db`` :heavy_check_mark: <br> ``-disco`` :heavy_check_mark:<br> ``-hello`` :heavy_check_mark: <br> ``-graph`` :heavy_check_mark: <br> ``-lookup`` :heavy_check_mark: <br> ``-ls`` :heavy_check_mark: |  ``-db-create`` :heavy_check_mark: <br> ``-svn-post-commit`` :heavy_check_mark: <br> ``-svn-pre-commit`` :heavy_check_mark: |

Also make some isolated trivial changes to clean up & fix minor issues from #2286.

#### :star: Context on changes that may seem non-minimal without explanation:

I added some new classes, but these do not change the functionality. They were created to preserve the current logic but in a way that is compatible with, & organised under, the Tornado framework:

* ``RosieDiscoServiceApplication`` organises the collection of request handlers forming the web application. A Tornado application does not *have* to be a class, but the amount of logic for defining this application merits one to keep everything clean. **Logically, the old ``RosieDiscoServiceRoot`` -> ``RosieDiscoServiceApplication`` + new ``RosieDiscoServiceRoot``**.
*  ``QueryHandler`` & ``SearchHandler`` were created as subclasses to ``RosieDiscoService`` because they are distinct API points & the most natural & ["Tornado-esque" way](https://www.tornadoweb.org/en/stable/web.html?highlight=web.RequestHandler#request-handlers) to adjust the data retrieval is to set up a base class & subclass it to override the HTTP GET. **Logically, new ``QueryHandler`` = old ``RosieDiscoService.query()`` & new ``SearchHandler`` = old ``RosieDiscoService.search()``**. I then moved the module-level function ``_query_parse_string()`` into ``QueryHandler`` as a static method, since it concerned only that class, to tidy up.

* ~~In order to keep things as simple as possible, I had to change the CLI a little, since [Tornado has a command-line parsing module](https://www.tornadoweb.org/en/stable/options.html) where options are much simpler to set-up than arguments (so ``start`` -> ``--start`` etc.). See the re-documented CLI via ``rosie disco --help`` / the docstring to  ``bin/rosie-disco`` for full detail.~~
* A few tweaks, unavoidable without adding complexity to ``ws.py``, were made to the templates, as they were doing the wrong thing otherwise (looking for a favicon ``.ico`` file that did not exist, for example), for reasons I am still unsure of.

#### TO DO list:

- [x] Get the Rosie Discovery service to run via Python 3.
- [x] Tidy up the Python 2 to 3 port to remove any unnecessary ``2to3`` tool changes.
- [x] Get a Tornado server to *start* according to the same ``rosie disco`` CLI as before...
- [x] & ditto to *stop* (cleanly) **[update: this was by far the most difficult part of this PR, taking an equivalent proportion of the time!]**.
- [x] Serve the basic service API points in the absence of data.
- Serve the service search & query API points, to return both:
  - [x] JSON data structures;
  - [x] web pages with tables populated with the results.
- [x] Ensure ~~all possible/[documented](http://metomi.github.io/rose/doc/html/api/rosie-web.html?highlight=hello#rosie.ws_client.RosieWSClient)~~ **[update: not all from docs were there originally; supporting only current/original API to minimise changes]** query arguments are functional ~~& not persistent~~ **[update: this is an issue in the current (``master``/CherryPy-variant) behaviour, hence leave as-is for this PR & register for fixing later]**.
- [x] Manual testing for functionality & correct outputs from REST inputs, as above.
- [x] Migrate automated testing.


sadielbartholomew:
> Migrate automated testing: * awaiting discussion before starting.

... Naturally, the test battery & Travis CI will now fail not only on tests relating to ``rosie disco``, but various others given the porting without back compat to Python 3 of only a small subset of Rose & Rosie modules.

@matthewrmshin I've tried to convey important context in my opening comment. I will fix the JSON ASAP. At your convenience we should discuss strategy for the testing aspect.

lib/python/rosie/ws.py
```diff
@@ -61,35 +76,89 @@ def __init__(self, *args, **kwargs):
                 ResourceLocator.default().get_util_home(
                     "lib", "html", "template", "rosie-disco")))
         db_url_map = {}
-        for key, node in rose_conf.get(["rosie-db"]).value.items():
+        for key, node in list(rose_conf.get(["rosie-db"]).value.items()):
```
oliver-sanders: 2to3 somewhat over-cautiously wraps every `items` call inside a `list`, it's only actually necessary if we change the keys of the dictionary (i.e. `pop`, `del`, `dict[newkey]=`), otherwise it's safe to remove these.

lib/python/rosie/ws.py
```diff
@@ -99,72 +168,56 @@ def __init__(self, props, prefix, db_url):
...
         self.dao = rosie.db.DAO(db_url)
+        self.root_service_url = root_service_url
+
+        # Arguments for API points and formatting
+        self.q_args = self.get_arguments("q")
+        self.s_args = self.get_arguments("s")
+        # Need to provide a default, else get MissingArgumentError w/ no arg:
+        self.format = self.get_argument("format", default=None)
 
     def __call__(self):
```
oliver-sanders: I have absolutely no idea what function this was originally supposed to serve!

lib/python/rosie/ws.py
```diff
@@ -99,72 +168,56 @@ def __init__(self, props, prefix, db_url):
...
+        if self.get_current_user():
+            data = self.HELLO % self.get_current_user()
         else:
             data = self.HELLO % "user"
-        if format == "json":
-            return json.dumps(data)
+        if self.format == "json":
+            self.write(json.dumps(data))
+            return
         return data
```
oliver-sanders: Out of interest, how come we write JSON data (and return None) but return strings?

matthewrmshin: And also... `format` is an argument of the REST API. Not sure why it has become a class variable.

lib/python/rosie/ws.py
sadielbartholomew: Nor I. There were also ``self.exposed = True`` instance variables set but not used at all in the logic...

I will remove ``__call__`` given your backing to my suspicions that it was redundant. I should have investigated whether it was at some point.

lib/python/rosie/ws.py
sadielbartholomew: Thanks for the explanation, it should be useful going forward. I wasn't being particular about the details of the Python 3 conversion at this stage of the WIP (for example a lot of blank lines have been blitzed from ``config.py`` which I need to eventually get re-added) as long as changes were vaguely sensible since I was going to do a 'tidy up' once the Disco utility is fully functional & tested under Tornado, particularly since I need to do more 2 to 3 on the tests.

lib/python/rosie/ws.py
sadielbartholomew: For my replies to both these questions, relating intrinsically to the nature of Tornado, I found nice code examples for simple apps [under an official 'demos' repo, e.g. for a 'blog' example](https://github.com/tornadoweb/tornado/blob/stable/demos/blog/blog.py) that you may find it useful to reference.

******

> how come we write JSON data (and return None) but return strings?

If JSON is requested by query, the data gets written in JSON to the output buffer, but a ``return`` is needed to bypass the ``return data`` line, otherwise both JSON & the rendered web page are served. This is not a the best way to go about that logically, but in the interest of minimising the changeset line count & logical flow I avoided recasting.

We don't actually return strings; this was an omission for the ``hello()`` method (thanks for highlighting it for me!) Data should simply be called as the argument to ``RequestHandler`` output methods such as ``self.write``, without any logical need to call ``return``, instead of returned, as for the JSON & as in the other methods rending the web pages:

https://github.com/metomi/rose/blob/98325edeff3ca0126cf6ce5f3d800dc038a03b9d/lib/python/rosie/ws.py#L232

> format is an argument of the REST API. Not sure why it has become a class variable.

In Tornado, as described further & here [quoted from the docs](https://www.tornadoweb.org/en/stable/guide/structure.html#handling-request-input), "request data in the formats used by HTML forms will be parsed for you and is made available in methods like ``get_query_argument`` and ``get_body_argument``" or as implemented here via [the Request Handler method covering both query & body args](https://www.tornadoweb.org/en/stable/web.html?highlight=get_argument#tornado.web.RequestHandler.get_argument):

https://github.com/metomi/rose/blob/98325edeff3ca0126cf6ce5f3d800dc038a03b9d/lib/python/rosie/ws.py#L177

where I assigned it to an instance variable to avoid having to call the method several times.

The ``format`` argument is not recognised as a general REST API key by keyword, as in ``cherrypy``, & needs to be fetched via such methods. You could see this via attempting to revert back to ``if format == "json":`` where it does not get recognised, brekaing the functionality.

Does that make sense? Also, if you can see a better way to process this & perhaps the other arguments, I am happy to try a new approach.

lib/python/rosie/ws.py
oliver-sanders: I've a hunch `self.exposed` might be related to the cherrypy `expose` decorator.

lib/python/rosie/ws.py
sadielbartholomew: Sorry @matthewrmshin, @oliver-sanders just elaborated on what more specifically you were referring to there in your comment. You are right, the class variable nature of some of these API arguments is inappropriate & I now know is the cause of an issue of persistent query arguments that I had been seeing & not previously understanding. Thanks both, this comment & discussion has really helped with my understanding.

sadielbartholomew:
*Update*: made some changes in response to feedback, but I still need to make some amendments to ensure all arguments are processed & do not persist across request instances due to improper class use. The 'TO DO' list has been updated to indicate progress thus far.

wxtim:
I have had a play - the functionality seems unchanged between the oper version (2019.01.00) and the test version - the only slight change I noticed was that search terms are replaced in the search box with themselves wrapped in quotes and square brackets - i.e. `Tim`, becomes `['Tim']`

bin/rosie-disco
```diff
@@ -20,22 +20,19 @@
...
 #
-#     For `rosie disco start`, if `PORT` is not specified, use port 8080.
-#
 # OPTIONS
-#     --non-interactive, --yes, -y
-#         (For stop only.) Switch off interactive prompting (=answer yes to
-#         everything)
-#     --service-root, -R
-#         (For start only.) Include web service name under root of URL.
+#     --start
```
wxtim: This is quite a big change to the way the CLI is used. Is this deliberate?

lib/python/rose/config.py
```diff
@@ -97,6 +97,7 @@
 import shlex
```
wxtim: I've made a very small number of changes to this file since you took it out of my branch, most of these are to fix doctests. There don't appear to be any conflicts.

lib/python/rosie/ws.py
```diff
@@ -61,35 +77,90 @@ def __init__(self, *args, **kwargs):
...
         self.db_url_map = db_url_map
         if not self.db_url_map:
             self.db_url_map = {}
-        for key, db_url in self.db_url_map.items():
-            setattr(self, key, RosieDiscoService(self.props, key, db_url))
 
-    @cherrypy.expose
-    def index(self, *_):
+        # Set-up the Tornado application request-handling structure.
+        service_root = r"/" + RosieDiscoServiceApplication.NS + r"/?"
```
wxtim: Is there a reason why you've called `RosieDiscoServiceApplication.NS` here and `self.NS` at ~line 101?
What does NS stand for? If I need to ask should it be more self-explanatory?

lib/python/rosie/ws.py
```diff
@@ -19,39 +19,55 @@
...
+
         self.props = {}
         rose_conf = ResourceLocator.default().get_conf()
         self.props["title"] = rose_conf.get_value(
             ["rosie-disco", "title"], self.TITLE)
         self.props["host_name"] = rose_conf.get_value(["rosie-disco", "host"])
         if self.props["host_name"] is None:
-            self.props["host_name"] = HostSelector().get_local_host()
+            self.props["host_name"] = socket.gethostname()
             if self.props["host_name"] and "." in self.props["host_name"]:
```
wxtim: Out of curiosity, in what circumstances does `socket.gethostname()` return something with a `.` in it?

sadielbartholomew:
Thanks for the review feedback, @wxtim, it will be useful going forward. I'm glad the utility seems to be doing as it should except in a few cases in your environment too.

Just to add a reply RE your summary comment, which I can't reply to directly:

> only slight change I noticed was that search terms are replaced in the search box with themselves wrapped in quotes and square brackets - i.e. Tim, becomes ['Tim']

Good spotting, I'll add that to the TO DO list for fixes to address before the PR transcends WIP status.

bin/rosie-disco
sadielbartholomew: Well noticed. It is deliberate & I have tried to explain the background in the third bullet point down under the 'Context on changes' section in my opening comment to the PR (sorry, there is a lot of content in that comment, so it is easy to miss). I'm happy to go into more detail verbally if you would like, though.

lib/python/rose/config.py
sadielbartholomew: Sure, I anticipated there might be some changes but was planning to wait until nearer a 'ready' stage to go back to make any adjustments. We should ensure that the most recent version of files you have ported end up in the final branch that gets reviewed for merging into ``master``.

lib/python/rosie/ws.py
```diff
@@ -61,35 +77,90 @@ def __init__(self, *args, **kwargs):
                 ResourceLocator.default().get_util_home(
                     "lib", "html", "template", "rosie-disco")))
         db_url_map = {}
-        for key, node in rose_conf.get(["rosie-db"]).value.items():
+        for key, node in list(rose_conf.get(["rosie-db"]).value.items()):
             if key.startswith("db.") and key[3:]:
                 db_url_map[key[3:]] = node.value
         self.db_url_map = db_url_map
         if not self.db_url_map:
```
wxtim: Another random question - given the line above, is this ever true?

lib/python/rosie/ws.py
sadielbartholomew: > Is there a reason why you've called RosieDiscoServiceApplication.NS here and self.NS at ~line 101?

There is not, no, other than that I had not noticed that both variants had been used. Good observation! I will make these consistent & since they are referring to a string i.e. a constant I think ``self.`` would be most appropriate to handle inheritance safely.

> What does NS stand for? If it's a standard tornado idiom?

I'm not actually sure! It is not related to Tornado, but to web services, or at least as coded in our codebases. I suspect it stands for 'namespace', as it has the root in a chain ``NS``, ``UTIL`` (utility, surely), ``HOST`` & then ``PORT``, as in:

https://github.com/metomi/rose/blob/e88db4259c5a31d578bc86615b2cd8b7b98ba556/lib/python/rose/ws.py#L36

where ``NS`` cases relevant to us are ``rose`` & ``cylc``.

> If I need to ask should it be more self-explanatory?

Yes, I endorse a clearer name, though quite how to shorten 'namespace' clearly I am not sure. Perhaps the clearest choice is not to shorten it.

lib/python/rosie/ws.py
```diff
@@ -189,33 +224,114 @@ def _render(self, all_revs=0, data=None, filters=None, s=None):
...
+
+class HelloHandler(RosieDiscoService):
+
+    """Writes a 'Hello' message to the current logged-in user, else 'user'."""
+
+    HELLO = "Hello %s\n"
+
+    def get(self):
+        """Say Hello on success."""
+        if self.get_current_user():
```
wxtim: As far as I can see you haven't over-ridden `tornado.web.RequestHandler.get_current_user`. As (the method in the base class](https://www.tornadoweb.org/en/stable/_modules/tornado/web.html#RequestHandler.get_current_user) is a placeholder which always returns `None` this logic will never select the `if self.get_current_user():`. Is this desirable behaviour?

lib/python/rosie/ws.py
```diff
@@ -189,33 +225,114 @@ def _render(self, all_revs=0, data=None, filters=None, s=None):
             filters=filters,
             s=s,
             data=data)
+        )
+
+
+class HelloHandler(RosieDiscoService):
```
wxtim: I've suggested a change here, and committed it - I didn't mean to, and I'm a tad surprised that i was able to do it - I wanted to offer you some changes you could choose to merge or not!

The changes are
```python
from getpass import getuser
if get_user():
  data = self.HELLO % get_user()

```

lib/python/rosie/ws.py
wxtim: do you need to shorten "namespace"? If you don't I'd suggest just commenting it.

sadielbartholomew: Agreed, let's go with ``NAMESPACE``. No ambiguity there, though we will have to presume it is the correct term given no-one we have asked at this stage is certain!

lib/python/rosie/ws.py
sadielbartholomew: From the [method description in the docs](https://docs.python.org/3/library/socket.html#socket.gethostname), it infers that a ``.`` is there only in the case of a FQDN (fully qualified domain name), which consists of a host name *and*, appended after a ``.``, the name of a domain where that host is located. There is some good & concise (& seemingly reliable) background including regarding the need cases for a domain component [here](https://kb.iu.edu/d/aiuv). So in the conditional in the subsequent lines:

https://github.com/metomi/rose/blob/9058d426113ce1d2d667964476eab742ab6e65bb/lib/python/rosie/ws.py#L70-L73

we check whether there is an attached domain & if so extract only the host name component.

lib/python/rosie/ws.py
sadielbartholomew: Good question, & well spotted. I can't say I have combed through the original code but it would be natural that it can be improved in ways such as this. Having though about the logic here, I agree that there seems to be no need for the ``if not`` statement & contained assignment, so I will remove it.

lib/python/rosie/ws.py
sadielbartholomew: (See comment under your next, associated, sub-review.)

lib/python/rosie/ws.py
sadielbartholomew: You do appear to have done so & I am also confused as to how that has been allowed! But that is no worries at all. I am unsure whether it will remain after I next push my local branch, but I will ensure the underlying problem as outlined is fixed.

Excellent point. I am not sure at this stage, but doubtful, if there is a need to import ``getuser``, but 'll play around to ensure the current user gets assigned properly via some means.

sadielbartholomew: I've just force pushed my own branch to overwrite this as I suspect the Tornado-based approach is more standard & wouldn't require the extra import. In case you were interested in how GitHub/``git`` managed the incorporation of this commit on my end, upon trying to push I got warned as follows:

```
To github.com:sadielbartholomew/rose.git
 ! [rejected]          HEAD -> i2271-rosie-disco-migrations (fetch first)
error: failed to push some refs to 'git@github.com:sadielbartholomew/rose.git'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

sadielbartholomew:
@matthewrmshin, @wxtim, @oliver-sanders: FYI, the tests are not yet sorted as I had hoped from yesterday so this is still a WIP sadly, but I will be coming in tomorrow to hopefully finalise it but at the least get it to an agreed stage where you all know how to go forward with it while I am on leave until 12th so I do not delay anything. In the meantime I will be watching this PR and can answer any questions, too.

sadielbartholomew:
@matthewrmshin, @wxtim, @oliver-sanders:

> but I will be coming in tomorrow

Ah, scratch that, I now can't make it as I still have some urgent tasks to do at home that I need to do before a flight this evening. Regardless, I will be keeping an eye on this PR for comments and will likely spend an hour or two on this while on leave in which case I'll comment with any progress.

sadielbartholomew:
All, please hold reviewing on this as it is still WIP & I am in the middle of making some preliminary changes that I may revert & need to tidy up. Sorry for any inconvenience caused.

lib/python/rosie/ws.py
```diff
@@ -19,77 +19,158 @@
...
+import tornado.web
+
+from isodatetime.data import get_timepoint_from_seconds_since_unix_epoch
 from rose.resource import ResourceLocator
 import rosie.db
 from rosie.suite_id import SuiteId
 
 
-class RosieDiscoServiceRoot(object):
+LOG_ROOT_TMPL = "~/.metomi/%(ns)s-%(util)s-%(host)s-%(port)s"
```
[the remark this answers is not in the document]
sadielbartholomew: Good idea! Though my brief was to make minimal changes & this is how it was originally, it's a simple one-line improvement so there shouldn't be any issue in making it.

lib/python/rosie/ws.py
```diff
@@ -19,77 +19,158 @@
...
+                "db_url": db_url,
+                "root_service_url": self.NAMESPACE,
+            })
+            handler = (service_root + key + r"/?$", RosieDiscoService,
+                       prefix_class_args)
+            hello_handler = (service_root + key + r"/hello/?$",
+                             HelloHandler, prefix_class_args)
+            search_handler = (service_root + key + r"/search(.*)",
+                              SearchHandler, prefix_class_args)
+            query_handler = (service_root + key + r"/query(.*)",
```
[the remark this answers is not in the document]
sadielbartholomew: That sounds like a more elegant approach. I'll have a play around to see if I can get it working. Thanks.

lib/python/rosie/ws.py
```diff
@@ -19,77 +19,158 @@
...
+from isodatetime.data import get_timepoint_from_seconds_since_unix_epoch
 from rose.resource import ResourceLocator
 import rosie.db
 from rosie.suite_id import SuiteId
 
 
-class RosieDiscoServiceRoot(object):
+LOG_ROOT_TMPL = "~/.metomi/%(ns)s-%(util)s-%(host)s-%(port)s"
+DEFAULT_PORT = 8080
+INTERVAL_CHECK_FOR_STOP_CMD = 5  # in units of seconds
```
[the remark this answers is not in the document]
sadielbartholomew: I put five in as a placeholder as I wasn't entirely sure what is best; once an ad-hoc server is started up it obviously will only be stopped once (& possibly just by a ``CTRL-C`` signal if the process is foregrounded) so I wouldn't want to call it to catch this *rare* event too often, but then the callback is trivial & will have minimal performance overhead.

Given this balance & in light of your comment I've changed to ``INTERVAL_CHECK_FOR_STOP_CMD = 1`` (& removed the ``"stopping within X seconds"`` STDOUT notice which now seems superfluous along with that stating ``"stopped"``) but don't think we should reduce it further.

lib/python/rosie/ws.py
```diff
@@ -189,33 +234,173 @@ def _render(self, all_revs=0, data=None, filters=None, s=None):
...
+    arg_msg_comp = "ad-hoc web service server"
+    # No value req for Boolean options e.g. --start equivalent to --start=True.
+    define("start", type=bool, default=False,
+           help="start %s (on specified port, else on port %s)" % (
+               arg_msg_comp, DEFAULT_PORT))
+    define("port", type=int, default=DEFAULT_PORT, help="port to listen on")
+    define("stop", type=bool, default=False, help="stop %s" % arg_msg_comp)
+    define("non-interactive", type=bool, default=False,
+           help="to stop %s w/o prompting" % arg_msg_comp)
+    parse_command_line()
```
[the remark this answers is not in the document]
sadielbartholomew: I changed to use the Tornado command-line parsing module thinking it would be cleaner & reduce the line count for management of the ad-hoc server, but considering it again & in light of your reasoning I can see that is not really true. I'll revert the CLI & adapt the PR accordingly. Thanks.

t/rosa-db-create/00-basic-3.dump
```diff
@@ -20,10 +20,10 @@ CREATE TABLE main (
 	from_idx VARCHAR(1024), 
 	PRIMARY KEY (idx, branch, revision)
 );
-INSERT INTO "main" VALUES('foo-aa000','trunk',1,'iris','eye pad','Should have gone to ...','$USER',1234567890,'A ',NULL);
-INSERT INTO "main" VALUES('foo-aa001','trunk',2,'roses','poetry','Roses are Red,...','$USER',1234567891,'A ',NULL);
-INSERT INTO "main" VALUES('foo-aa001','trunk',3,'roses','poetry','Roses are Red, Violets are Blue,...','$USER',1234567892,' M',NULL);
-INSERT INTO "main" VALUES('foo-aa000','trunk',4,'iris','eye pad','Should have gone to ...','$USER',1234567893,'D ',NULL);
+INSERT INTO main VALUES('foo-aa000','trunk',1,'iris','eye pad','Should have gone to ...','$USER',1234567890,'A ',NULL);
```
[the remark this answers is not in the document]
sadielbartholomew: > were they causing problems?

Yes, for reasons not clear to me the test fails with the tables quoted as previously. I tried altering some logic that may have been relevant e.g. within ``create()`` in ``rosie/db_create.py`` where the tables are defined, but after spending (admittedly not too great an amount of) time on it, I could not find a simple way to get a pass without stripping the quotes; it doesn't seem crucial to me & can always be registered as a minor follow-on tweak, unless anyone has the motivation to have a play & see if they can succeed at that?

sphinx/installation.rst
```diff
@@ -356,13 +356,13 @@ Edit the :rose:conf:`rose.conf[rosie-disco]` settings to configure
...
 
-   setsid /path/to/rose/bin/rosie disco start 0</dev/null 1</dev/null 2>&1 &
+   setsid /path/to/rose/bin/rosie disco --start 0</dev/null 1</dev/null 2>&1 &
 
 Check that the server is up and running using ``curl`` or a local
 web browser. E.g. If you have configured the server's port to be 1234,
 you can do::
 
-   curl -I http://localhost:1234/
+   curl -I http://localhost:1234/rosie/
```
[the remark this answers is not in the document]
sadielbartholomew: True, good point. I'll get the behaviour reverted to obey endpoints without this addition.

For context, throughout development I mistakenly assumed the ``rosie/`` URL component would always be part of the root path, forgetting it is just appended in the case of the ``--service-root`` option, or if set-up for a system environment WSGI app.

sadielbartholomew:
All feedback now responded to, including the most recent set from @oliver-sanders, with the exception of https://github.com/metomi/rose/pull/2288#discussion_r262454430, which I have not attempted. I may actually leave that out of this PR at least if there is no challenge, since what I have works, & whilst not as elegant as the suggestion there should only ever be a low number of keys e.g. a few on average, so there should only be a small-ish number of handlers created in all cases even with the ``for`` loop over those keys.

lib/python/rosie/ws.py
matthewrmshin: Can use `HostSelector().get_local_host()` again, now that it has completed 2to3.

lib/python/rosie/ws.py
```diff
@@ -189,33 +239,203 @@ def _render(self, all_revs=0, data=None, filters=None, s=None):
...
+def _log_server_status(application, host, port):
+    """ Log a brief status, including process ID, for an application server."""
+    log_root = os.path.expanduser(LOG_ROOT_TMPL % {
+        "ns": application.NAMESPACE,
+        "util": application.UTIL,
+        "host": host,
+        "port": port})
+    log_status = log_root + ".status"
+    if not os.path.isdir(os.path.dirname(log_root)):
+        os.makedirs(os.path.dirname(log_root))
```
matthewrmshin: Use `os.makedirs(os.path.dirname(log_root), exist_ok=True)` - and remove the `if not ...` test.

sadielbartholomew: Great points. Thanks Matt, I'll get them implemented.

sadielbartholomew:
I have now completed the checklist & I am now happy with the PR, so it is **ready for final review** (though I may have to tweak a few things if the Travis CI report raises anything new). I am cutting the WIP marker in the PR title accordingly.

#### Notes for reviewers

RE the Codacy failure: the three failures are not applicable, as the "overridden 'get' method" takes a variable number of arguments via:

https://github.com/metomi/rose/blob/94811db5b70699dc16256b6dfb04b836aa27fa76/lib/python/rosie/ws.py#L195

There is definitely further work that should be done to improve the code, as listed below, but in interests of getting this port in & moving on with completing the full migration to Python 3, I think this should be registered in an Issue & left to complete in the future.

#### Further work out of scope for this PR

* **Persistent query parameters**: once a search or query is made on a database, the parameters for that request remain in the path, even when other pages are visited intermediately & even on restart after the whole application & its server are stopped. This is obviously not good, but this problem was manifest in the original code too, so cannot be critical.
* **Clunky retrieval of query parameters**: these get picked up via the regular expressions in the Tornado application handler set-up & passed as into the corresponding handler class as arguments to the ``get()`` method, but I am often using ``self.get_query_argument(QUERY_KEY, default=DEFAULT)``. We should make use of the former rather than the latter in most or even all cases, as it is cleaner & does not have to repeatedly check the URL path. This may solve the above.
* **Clunky set-up of request handling**: the regular expressions used to point to specific handlers could also be improved, e.g. as suggested in https://github.com/metomi/rose/pull/2288#discussion_r262454430.
* **Error handling**: I have simply implemented the bare-bones error catching from the original code, but there are standard (certainly not unusual) error cases where traceback is written out, which should be replaced with clean dedicated pages displaying the relevant HTTP error code with a context line.

sadielbartholomew:
> ready for final review

Actually hold that for a moment, I need to check the ``rosie disco`` functionality as some parts appear to have broken as I fixed the tests...

sadielbartholomew:
Okay Travis CI is not happy on a few counts, despite a full local pass on the tests, so I will take a look at that now. Thanks for your patience, all.

sadielbartholomew:
#### For context on commit ``30e4a32``:

 when developing on my local environment, I was seeing for ``t/rosa-db-create/`` that:

> the test fails with the tables quoted as previously

after ``INSERT INTO`` statements (see https://github.com/metomi/rose/pull/2288#discussion_r265691173 for further detail). On the contrary, for the same state of the codebase, on Travis CI the test seems to fail with the quotes removed. This is rather frustrating, as for that reason I can only get either the local or the Travis CI test suite to pass, but not both, without post-processing the diff!

Ultimately the database logic (& functionality) is unaffected, so it is just failing on the quotes making the ``diff`` not match. Therefore I have reverted to the original quoted state, as that seems most common to receive as output (as observed on other environments).

**********

Travis CI is now hanging, so I need to look into that, but I should nearly be there.

sadielbartholomew:
For info, I narrowed the hang down to ``t/rosa-svn-post-commit/...`` ``01-mail-passwd.t`` & ``... 02-mail-owner-passwd.t``, which both hang near the start within a call to ``mock_smtpd_init()``. @matthewrmshin thinks there is an issue with the writing of logs from ``smtpd`` under Python 3, which worked fine in Python 2. @matthewrmshin deduced this was due to a ``-u`` option being required to pass to ``stmpd``, to set ``PYTHONUNBUFFERED`` to be ``True``, though there still appear to be issues even with this change. I am investigating this now.

sadielbartholomew:
#### Updates:

* Travis CI now passes :confetti_ball:;
* Codacy failure is justified in https://github.com/metomi/rose/pull/2288#issuecomment-476096016, immediately under the heading 'Notes for reviewers'.
* The code coverage has actually increased (``+2.99%``), but the general target has not been reached. However this is ultimately a Python 3 migration & not a functionality/feature addition, with no added or intentionally improved tests, so it is natural & acceptable for the coverage to not improve markedly.

Therefore @matthewrmshin, @oliver-sanders & @wxtim please review at your convenience as I believe this is ready for merge.

wxtim:
## Review Progress
- [ ] First pass read
- [ ] Run tests - massive thanks to @sadielbartholomew for helping my get my rose.conf sorted
- [ ] Check files in detail:
    - [ ] rose/reporter.py
    - [ ] rosie/ws.py
    - [ ] rosie/ws_client.py
    - [ ] t/rosie-disco/00-...

lib/python/rosie/suite_id.py
```diff
@@ -143,9 +143,10 @@ def get_latest(cls, prefix=None):
                 if i == 0:
                     return None
                 raise SuiteIdLatestError(prefix)
-            dirs = [line for line in out.splitlines() if line.endswith(b"/")]
```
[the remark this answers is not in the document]
wxtim: Doesn't necessarily mean that they are wrong though. If @sadielbartholomew  has changed them further upstream this feels like a good call.

sadielbartholomew:
Thanks for the feedback @oliver-sanders; sorry, due to priority reviewing & to leave I will not be able to address it until next week.

*Note to self*: I also need to update ``rose check-software`` dependencies to remove CherryPy & add in Tornado, etc.

sadielbartholomew:
(Conflicts resolved.)

lib/python/rosie/suite_id.py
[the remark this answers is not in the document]
sadielbartholomew: In general, I have only made changes to Tim's porting work, in practice almost always relating to the handling of strings, if:

    I have observed any of the new, wider set of ported (non-skipped) tests failing with stdout & stderr suggesting some issue resulting ultimately from trademark Python version related issues, making logical tweaks to fix those tests; or
    The ported code, especially with respect to encoding or decoding, could be simplified &/or made cleaner:
        (as Tim indicated) the de- or en-coding is still done as before, but has been moved further upstream or downstream; or
        (as in this case) by converting certain variables (such as dirs here) that are only used as an intermediate to some other variable, to strings from bytes or vice versa.

Most of these discrete changes should therefore be necessary, but I will go through & do a sanity check on all such changes to make sure I am not changing the string-handling behaviour. Thanks for raising this.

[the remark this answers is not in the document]
sadielbartholomew: In general, I have only made changes to Tim's porting work, in practice almost always relating to the handling of strings, if:

1. I have observed any of the new, wider set of ported (non-skipped) tests failing with stdout & stderr suggesting some issue resulting ultimately from trademark Python version related issues, making logical tweaks to fix those tests; or
2. The ported code, especially with respect to encoding or decoding, could be simplified &/or made cleaner:
       * (as Tim indicated) the de- or en-coding is still done as before, but has been moved further upstream or downstream; or
       * (as in this case) by converting certain variables (such as dirs here) that are only used as an intermediate to some other variable, to strings from bytes or vice versa.

Most of these discrete changes should therefore be necessary, but I will go through & do a sanity check on all such changes to make sure I am not changing the string-handling behaviour. Thanks for raising this.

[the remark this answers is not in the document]
sadielbartholomew: In the cases in this particular file, to provide quick reassurance, I've wrote a small script you could run:

```python
import sys

out = bytes(sys.argv[1].encode())
dir_url= str(sys.argv[2])

###------ Before
old_dirs = [line for line in out.splitlines() if line.endswith(b"/")]
old_dir_url = dir_url + "/" + sorted(old_dirs)[-1].rstrip(b"/").decode()

###------- After
new_dirs = [line for line in out.decode().splitlines() if line.endswith("/")]
new_dir_url = dir_url + "/" + sorted(new_dirs)[-1].rstrip("/")

print("intermediate dirs:", old_dirs, new_dirs)  # might differ but it is irrelevant
print("final dir_url:", old_dir_url, new_dir_url)  # need to be the same; no change
```

so run any (valid) example for the file ``out`` & string ``dir_url``, e.g:

```console
$ python tester_temp.py "orange\n/yellow\n/green/" "blue"
intermediate 'dirs': [b'orange\\n/yellow\\n/green/'] ['orange\\n/yellow\\n/green/']
final 'dir_url': blue/orange\n/yellow\n/green blue/orange\n/yellow\n/green

```

to see the same ``dir_url``, even if the ``dirs`` changes from byte to string (though it is not used elsewhere so that is not an issue).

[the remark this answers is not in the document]
sadielbartholomew: In fact, then the only changes here to Tim's porting work are those in ``rosie/svn_post_commit.py``, which were required to fix the associated tests (``t/rosa-db-create`` & ``t/rosa-svn-post-commit``) which were otherwise failing in several ways.

lib/python/rosie/ws.py
```diff
@@ -189,33 +226,229 @@ def _render(self, all_revs=0, data=None, filters=None, s=None):
...
+
+
+class SearchHandler(RosieDiscoService):
+
+    """Serves a search of the database on the page of a given prefix."""
+
+    def get(self, _):
+        """Search database for rows with data matching the search string."""
+        s_arg = self.get_query_argument("s", default=None)
+        all_revs = self.get_query_argument("all_revs", default=0)
```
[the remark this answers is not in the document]
sadielbartholomew: Yes, I was aware of the clunky nature of the current handling of query parameters, as registered (fair enough if you missed this, as it is hidden away in a substantial comment):

> Clunky retrieval of query parameters: these get picked up via the regular expressions in the Tornado application handler set-up & passed as into the corresponding handler class as arguments to the get() method, but I am often using self.get_query_argument(QUERY_KEY, default=DEFAULT). We should make use of the former rather than the latter in most or even all cases, as it is cleaner & does not have to repeatedly check the URL path.

I seem to recall adding in the ``_`` parameter to ``SearchHandler.get`` simply to appease Codacy somewhat, as, while in the Tornado Handler context I think it is not necessary, it is requesting the same parameter interface for this and the ``RosieDiscoService.get`` it overrides.

Originally I left this as work to complete after this PR, but now the urgency has been downgraded due to reviewer preoccupation, I will go & revise the query parameter handling, as above & to include addressing the points in your comments, as that seems sensible with more time.

sadielbartholomew:
@oliver-sanders, sorry I was not able to get back to this until this afternoon. In response to this:

> Would be nicer if URL params were provided as kwargs, saves all of this get_query_argument stuff and would raise the appropriate error if an invalid parameter if provided. Does Tornado let us do that?

From investigations I have only just observed, though I am finding it surprisingly hard to verify from the Tornado docs or source code or indeed from online Q&A sites etc., that **the regular expressions used in Tornado app ``handlers`` arguments, to match URLs to direct to specific Handler classes, do not capture anything after a ``?`` symbol**, I imagine to keep query parameters distinct.

For example, when specified inside a ``handlers`` argument to a Tornado app ([``tornado.web.Application``](https://www.tornadoweb.org/en/stable/web.html#application-configuration)), the regex I was hoping to use:

```regex
``r"/get_(.+)"(?:&|\?)(format=)?(.+)?``
```
would only find a match for ``get_query_operators&format=json``, not for ``get_query_operators?format=json``, despite both being captured by the standalone regex.

From reading online, it does seem like the suggested way to retrieve query parameters is actually how I have implemented it here, using the ``RequestHandler.get_query_argument``(``s``) methods.

So, in terms of refactoring the query parameter retrieval, the only tidying up that can be done bar sanitising the arguments passed in, as per your comment https://github.com/metomi/rose/pull/2288#discussion_r271374670 & which I have just addressed & pushed up, is to create some method in the inherited class ``RosieDiscoService`` to retrieve any relevant query parameters, such as the three required for the ``SearchHandler``:

https://github.com/metomi/rose/blob/42e0c2e289f16c4e6bd5b87ff320df19a46ae71a/lib/python/rosie/ws.py#L274-L276

Something like:

```python
def retrieve_query_args(*args):
   all_query_args = {}
   for query_tuple in args:
       query_name, query_arg, default_arg = query_tuple
       all_query_args[query_name] = self.get_query_argument(query_arg, default_arg)
    return all_query_args
```

& call this once with the relevant arguments for each Handler sub-class ``get(self)`` method.

But I am not sure if this makes things "nicer" or not, as it reduces duplication somewhat but also adds a layer of abstraction. What do you (& others if they want to chip in) think?

matthewrmshin:
Change looks OK by :eyes: with :eyeglasses:. On testing using the ad-hoc server, there seems to be a complete lack of server log? Can this be added? E.g. an access log to find out what requests were received and their return statuses, etc.

sadielbartholomew:
> Can this be added? E.g. an access log to find out what requests were received and their return statuses, etc.

Sure, @matthewrmshin, that would be very useful & I will get it implemented.

sadielbartholomew:
> Sure, @matthewrmshin, that would be very useful & I will get it implemented.

Sorry for the delay, I was inundated with user support over Easter being the only person not on leave & I put this on hold accordingly, given I also knew it would be a while before reviewers returned from holiday.

I have set-up three informational logs, corresponding to [Tornado's built-in logging streams](https://www.tornadoweb.org/en/stable/log.html#module-tornado.log) (with custom server start & stop messages added for full transparency), to complement the server status log required to parse for cleanly stopping the server, as was already implemented.

matthewrmshin:
Does Codacy have a point about overriding the `get` methods with different signatures?

sadielbartholomew:
> Does Codacy have a point about overriding the get methods with different signatures?

No, as far as I understand it; the method that is being overridden has the signature:  
https://github.com/metomi/rose/blob/2ba14b22d606faa776ec87a79ab8938962f49e94/lib/python/rosie/ws.py#L198
so it should accept any number of positional arguments, required since the sub-classes overriding that method need different numbers by context. I have also manually tested the application quite thoroughly now & have not seen any warnings or errors related to incompatible interfaces for Handler methods.

sadielbartholomew:
For information, I have just made one final (bar any new review feedback) commit to fix the logging of the server stopping, as I have just noticed that with the recently implemented logging code a line to that effect was not being written out due to the blocking nature of the Python logging utility.

matthewrmshin:
I think Codacy/PyLint is telling us that if you override a method with signature `def get(self, *args):`, the sub-class's method should also have the signature `def get(self, *args):`. I suppose, if the methods in the sub-classes have different contexts, you should:
* Either: stick with the same signature. Raise `ValueError` if `args` is expected to be of certain length.
* Or: add some `# pylint: disable=...` to tell PyLint to shut up.

wxtim:
I'm still happy with my approval - but have come up with 2 more questions:
1. If I search for something, click on the link top-left and return to the home screen then return to the mi... page the search is retained. Deliberate? Desirable?
1. There's no terminal output when I start the server. I think it'd be nice to give me a URL?

oliver-sanders:
Have been through and marked a few review comments as resolved.

sadielbartholomew:
Thanks for elaborating @matthewrmshin. In my next commit I will implement one of your suggestions (ideally the first, but it depends if I can do it in practicable time without making the code too artificial & harder to follow).

In response to your comments @wxtim:

> If I search for something, click on the link top-left and return to the home screen then return to the mi... page the search is retained. Deliberate? Desirable?

Desirable, no. It's hard to answer if it is 'deliberate'; I am aware of it but I am not planning to amend it in this PR, as covered this in my comment https://github.com/metomi/rose/pull/2288#issuecomment-476096016 (see the bullet point on 'Persistent query parameters'). The original brief for this PR was to do an absolutely bare-bones CherryPy to Tornado conversion, making no change to behaviour, & the persistence of the URL parameters and their corresponding rendered pages was a problem that was there originally. It is not trivial to fix from some exploratory work I did early on.

> There's no terminal output when I start the server. I think it'd be nice to give me a URL?

There should be? I've put a print statement which should show when you start the server (logically, & it shows for me):

https://github.com/metomi/rose/blob/46abd5684ed43bc2132bbd1e22ddf8e61143a074/lib/python/rosie/ws.py#L465-L468

Though adding a URL is a good idea, thanks. I'll add that in within my next commit. Saying that, can I ask from now on that you are aware that while there is a lot of improvement we can make to this utility, that was not the goal of this PR, so for improvements such as these, please just add them to the list of follow-on work here: https://github.com/metomi/rose/pull/2288#issuecomment-476096016, which will become an Issue once this is merged.

matthewrmshin:
Codacy/PyLint has a final complaint, which I believe can be eliminated.
Travis CI unhappy.

lib/python/rosie/ws.py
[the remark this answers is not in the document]
sadielbartholomew: [See point 4 of https://github.com/metomi/rose/pull/2288#issuecomment-476096016 (which upon merge will be moved to an Issue): this will be follow-on work to this PR.]

lib/python/rosie/ws.py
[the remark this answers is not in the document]
sadielbartholomew: > ...would raise the appropriate error if an invalid parameter if provided. Does Tornado let us do that?
>
> For example would ?formmat=json return an error.

Sorry, I did not address this part in my previous comment, which I deemed & marked as resolving the conversation. In short, given the nature of URL query handling established as standard in Tornado (see https://github.com/metomi/rose/pull/2288#issuecomment-481802596), there is not an easy way to deal with invalid query parameters, so I will leave this as follow-on work, & accordingly I am just about to add it as a new point for the follow-up Issue (text as in https://github.com/metomi/rose/pull/2288#issuecomment-476096016).

oliver-sanders:
Code all reviewed, Rosie Disco tested as working with two small niggles:

1. `rosie disco` doesn't seem to print the status of a running server as documented.
  * I guess this should just print any status files in the terminal so should be pretty quick to fix.
   * Otherwise we could temporally remove or undocument it?
2. If the server crashes or is ctrl-c killed then `rosie disco stop` results in traceback (as the log files get left behind so it tries to kill a process which isn't there).
  * A quick try/except would probably suffice.
  * Bonus points for tidying up the redundant status file but that isn't necessary.

sadielbartholomew:
Thanks, @oliver-sanders, great reviewing!

> rosie disco doesn't seem to print the status of a running server as documented.

Ah, well spotted. I've just fixed that in the latest commit.

> If the server crashes or is ctrl-c killed then rosie disco stop results in traceback (as the log files get left behind so it tries to kill a process which isn't there).

Good spot again. In the latest commit I've also ensured that the traceback is not longer produced in this case.

> Bonus points for tidying up the redundant status file but that isn't necessary.

That will be left as follow-on work, as per bullet point 1 of https://github.com/metomi/rose/pull/2288#issuecomment-476096016.

PR #2289, opened 2019-02-06
config-edit: fix int spin box manual edit

matthewrmshin:
Bug introduced by #2264.

Value change is not being picked up when users input text to the int spin box.
This change should fix.


matthewrmshin:
Travis CI failures unrelated. I'll look into those separately.

matthewrmshin:
We don't test for GUI logic...

sadielbartholomew:
> We don't test for GUI logic...

That'll do as a justification, then :)

PR #2290, opened 2019-02-06
Call opendir before testing contact file existence

matthewrmshin:
Call `os.open(..., os.O_DIRECTORY)` on the suite's service directory
(i.e. the parent directory of the suite's contact file) before testing
for existence of the contact file. This ensures that the file system is
synchronised e.g. for NFS or the file existence test may fail even
though the file is already there.


matthewrmshin:
Travis CI test failures unrelated. I am also unable to reproduce them in my environment...

PR #2294, opened 2019-02-13
Remove unused imports

kinow:
Hi! Re-synced my repository, and the IDE marked these imports as unused :+1: hoping Travis-CI won't complain about it too.

Cheers

### DIFF
--- a/lib/python/rose/apps/comparisons/output_grepper.py
+++ b/lib/python/rose/apps/comparisons/output_grepper.py
@@ -19,8 +19,7 @@
 # -----------------------------------------------------------------------------
 """Return a list of values matching a regular expression."""
 
-from rose.apps.rose_ana_v1 import DataLengthError, data_from_regexp
-import re
+from rose.apps.rose_ana_v1 import data_from_regexp
 
 REGEXPS = {
     'um_wallclock': r"Maximum Elapsed Wallclock Time:\s*(\S+)",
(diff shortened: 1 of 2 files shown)


kinow:
(only risk in changes like this that I can think about, is if we a) have a `import os` for example in `a.py`... b) `a.py` does not specify `__all__ = [...]`, and c) another module is doing `from a import *`.... which may or may not be desired. In a case like that, it could be better to move `import os` to `b.py` IMHO)

lib/python/rosie/browser/__init__.py
```diff
@@ -50,11 +50,7 @@
 
 """
 
-import ast
```
wxtim: You might want to modify the docstring (line 36) to match what is going on.

kinow: Good spot @wxtim . Happy to apply that change, but not too sure what I should change. As I have very little understanding of Rose, I've limited to just what the IDE pointed. Should we delete the whole paragraph, i.e.

```
The values you enter will be cast by Python's ast.literal_eval, so:
foo=100
will be cast to an integer, but:
bar="100"
will be cast to a string.
```

in this pull request too? If so just let me know and I will add an extra commit for that :)

wxtim: propose 
```python
"""
The values in the rose.conf will be treated as string or int depending on
whether you quote them, so:
size_history=100
will be cast to an integer, but:
size_history="100"
will be cast to a string.
This may cause a type error if the value is the wrong type.
"""
```
I'd have created a PR on your branch but Octocat Wan Kenobi said:
https://github.com/404

oliver-sanders:
Conflicts :(

Issue #2278, opened 2019-01-21
testing: register flaky tests for fixing


sadielbartholomew:
Name & shame all known instances of tests with non-deterministic results under either a local test battery or Travis CI run, with the aim of maximising all-round reliability. Sister Issue to cylc/cylc#2894.

Please:

* add to the list as cases emerge;
* annotate the catalogue as appropriate;
* fix (stabilise) them!

#### Tests with random failures:

*Format*: test name (path from ``rose.git/t/``*) | first noticed | notes
*Labels*: LTB == local test battery; TCI == Travis CI.
Tick to mark as amended to be reliable.

- [ ] ``t/docs/02-tutorial-suites.t`` | 21/01/19 | TCI
- [ ] ``t/rose-suite-shutdown/03-normal-name.t`` | 21/01/19 | TCI
- [ ] ``t/rose-app-run/06-namelist.t`` | 21/01/19 | TCI

oliver-sanders:
First two tests are now gone, the third one doesn't seem to be flaky at the moment. Closing this as the tests seem stable, can open issues for individual tests if/when they become a problem.

PR #2295, opened 2019-02-13
Simplify boolean expression

kinow:
Besides searching for `== True` and `== False`, now I'm also searching for `== []` when syncing code :smile: 

I think these two statements are equivalent, but without creating an empty array (though I suspect compiler/interpreter/etc should optimize it)?

Cheers
Bruno

### DIFF
--- a/lib/python/rose/macros/trigger.py
+++ b/lib/python/rose/macros/trigger.py
@@ -61,7 +61,7 @@ class TriggerMacro(rose.macro.MacroBaseRoseEdit):
                 expr = opt_node.value
                 id_value_dict = rose.variable.parse_trigger_expression(expr)
                 for trig_id, values in id_value_dict.items():
-                    if values == []:
+                    if not len(values):
                         id_value_dict.update({trig_id: [None]})
                 self.trigger_family_lookup.update({setting_id: id_value_dict})
         self._trigger_involved_ids = self.get_all_ids()


kinow:
Oh, looks like the code is not automatically optimized by the interpreter.

```bash
Python 3.7.2 (default, Dec 29 2018, 06:19:36) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dis
>>> dis.dis("""
... a = []
... a == []
... """)
  2           0 BUILD_LIST               0
              2 STORE_NAME               0 (a)

  3           4 LOAD_NAME                0 (a)
              6 BUILD_LIST               0
              8 COMPARE_OP               2 (==)
             10 POP_TOP
             12 LOAD_CONST               0 (None)
             14 RETURN_VALUE
>>> dis.dis("""
... a = []
... not a
... """
... )
  2           0 BUILD_LIST               0
              2 STORE_NAME               0 (a)

  3           4 LOAD_NAME                0 (a)
              6 UNARY_NOT
              8 POP_TOP
             10 LOAD_CONST               0 (None)
             12 RETURN_VALUE
>>> 
```

Maybe in runtime or between the interpreter and bytecode execution there's some optimization... but looks like the `not` version avoids a `BUILD_LIST`, and a `COMPARE_OP`. `UNARY_NOT` should be cheaper too? :thinking: anywho, hopefully Travis won't complain about this change.

sadielbartholomew:
... sorry, I suspected the Travis CI build would back up the suspicions from my previous comment, but accidentally pressed 'cancel build' when I went to check. It wasn't intentional :grimacing: I've just restarted it so if it fails consistently in a reliable & justifiably related test (e.g. not one named in #2278) that would suggest the ``trigger.py`` line should also stay as it is.

I realise on the surface the expressions in questions seemed poor & improvable, but we don't want to break anything!

kinow:
Déjà vu indeed @sadielbartholomew ! Sorry about that.

I thought about adding a comment to prevent anyone (especially someone called Bruno with memory as good as Doris the fish) of proposing this change again.

Then decided to give it one last try :slightly_smiling_face:

kinow:
Force-updated the commit, now with `len` :tada: avoids the creation of an empty list, exchanging by an import.

```bash
Python 2.7.15rc1 (default, Nov 12 2018, 14:31:15) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import dis
>>> dis.dis("""
... a = []
... not len(a)
... """
... )
          0 UNARY_POSITIVE 
          1 STORE_GLOBAL    15648 (15648)
          4 SLICE+2        
          5 DELETE_NAME      2653 (2653)
          8 JUMP_FORWARD    29807 (to 29818)
         11 SLICE+2        
         12 IMPORT_NAME     28261 (28261)
         15 STORE_SLICE+0  
         16 STORE_GLOBAL     2601 (2601)
>>> 
```

Which should be OK :grin:

sadielbartholomew:
Really sorry @kinow, I have just realised that my comment (https://github.com/metomi/rose/pull/2228#discussion_r216618807) was actually wrong, as I seemingly mistook the variable ``section`` to be the same as ``sections``. Looking again, ``sections`` cannot be NoneType as it is created from ``get_sections_from_namespace()`` which only returns lists, which may be empty.

So I think these statement changes will be fine as they were, though the changes are they are now do also work & make sense, so I am approving. Feel free to revert the commit to the state it was before & I will approve immediately, though. Sorry about the error!

lib/python/rose/config_editor/nav_panel_menu.py
```diff
@@ -65,7 +65,7 @@ def add_dialog(self, base_ns):
             else:
                 sections = self.data.helper.get_sections_from_namespace(
                     base_ns)
-                if sections == []:
+                if not len(sections):
```
wxtim: PEP8 says no:

> For sequences, (strings, lists, tuples), use the fact that empty sequences are false.
```python
Yes: if not seq:
     if seq:

No:  if len(seq):
     if not len(seq):
```

kinow: See Sadie's previous comment... I believe there's a check elsewhere for `None`. At this step we are validating only list emptiness if I understand it correct.

And I tend to agree with what pep8/pycodestyle/black/etc say most time, but sometimes I have to review and then simply ignore pep8/et al.

e.g. of a warning that may change later (by the maintainer of TornadoI believe: https://github.com/PyCQA/pycodestyle/issues/815).

wxtim: Sorry I missed the bit about the check elsewhere for `None`.
Happy to approve if I know you know about the PEP8 question and have deliberately ignored it, although I'd like to understand why this is more desirable than PEP8's preference, which as far as I can see involves two fewer compiled instructions.

```python
>>> dis("a = []; not a")
  1           0 BUILD_LIST               0
              3 STORE_NAME               0 (a)
              6 LOAD_NAME                0 (a)
              9 UNARY_NOT
             10 POP_TOP
             11 LOAD_CONST               0 (None)
             14 RETURN_VALUE
>>> dis("a = []; not len(a)")
  1           0 BUILD_LIST               0
              3 STORE_NAME               0 (a)
              6 LOAD_NAME                1 (len)
              9 LOAD_NAME                0 (a)
             12 CALL_FUNCTION            1 (1 positional, 0 keyword pair)
             15 UNARY_NOT
             16 POP_TOP
             17 LOAD_CONST               0 (None)
             20 RETURN_VALUE

```

PR #2297, opened 2019-02-19
New "rose suite-run" config: root-dir{log}.

### LINKED ISSUE #2287: New "rose suite-run" config for "root-dir{log}"?
To allow symlinking of suite log directories to another location, distinct from the data dirs.

This would (I think) allow secure job log access via `cylc review` until we have authentication built in in cylc-8.  Here's how:

- configure `rose suite-run` to symlink log dirs to (say) `/etc/cylc-logs/$USER/cylc-run/<SUITE>/log`
- configure the web server running `cylc review` to a) not follow symlinks; and b) resolve `/etc/cylc-logs` insead of the normal user home directory

The current situation is insecure if your home filesystem is too small to hold thousands of job logs, because then you have to allow your web server to follow symlinks (to arbitrary locations, in principle) and anyone who can see the `cylc review` service host:port can see anything that the web server can see.

Presumably this would be a pretty easy change to make?

hjoliver:
Close #2287 

Allow `rose suite-run` to put suite log directories in other locations, symlinked to the standard location.

(I had a crack at this because we need it pretty desperately on the new HPC at NIWA).

Handles archiving of symlinked log dirs by making a local symlink to the external archive.

Example:
```
root-dir{log}=*=/tmp/$USER
```
Result after one run:
```
ls -grt ~/cylc-run/foo                                                                                                 
total 16
lrwxrwxrwx 1 oliverh   46 Feb 20 12:29 log.20190219T232926Z -> /tmp/oliverh/cylc-run/foo/log.20190219T232926Z/
lrwxrwxrwx 1 oliverh   20 Feb 20 12:29 log -> log.20190219T232926Z/
-rw-r--r-- 1 oliverh  607 Feb 20 12:29 suite.rc
...
```

After another run:
```
ls -grt ~/cylc-run/foo                                                                                                 
total 16
-rw-r--r-- 1 oliverh  607 Feb 20 12:29 suite.rc
lrwxrwxrwx 1 oliverh    6 Feb 20 12:29 cylc-suite.db -> log/db
lrwxrwxrwx 1 oliverh   46 Feb 20 12:31 log.20190219T233124Z -> /tmp/oliverh/cylc-run/foo/log.20190219T233124Z/
lrwxrwxrwx 1 oliverh   20 Feb 20 12:31 log -> log.20190219T233124Z/
lrwxrwxrwx 1 oliverh   53 Feb 20 12:31 log.20190219T232926Z.tar.gz -> /tmp/oliverh/cylc-run/foo/log.20190219T232926Z.tar.gz
-rw-r--r-- 1 oliverh  362 Feb 20 12:31 suite.rc.processed
drwxr-xr-x 3 oliverh 4096 Feb 20 12:31 work/
...
```

### DIFF
--- a/etc/rose-meta/rose-suite-conf/rose-meta.conf
+++ b/etc/rose-meta/rose-suite-conf/rose-meta.conf
@@ -21,6 +21,14 @@ help=A new line delimited list of PATTERN=DIR pairs.
     = The DIR should be the root directory where the suite's directory
     = should be created.
 
+[=root-dir{log}]
+description=Specify the suite's log/ directory location on particular hosts.
+help=A new line delimited list of PATTERN=DIR pairs.
+    =
+    = The PATTERN should be a glob-like pattern for matching a host name.
+    = The DIR should be the root directory where the suite's share directory
+    = should be created.
+
 [=root-dir{share}]
 description=Specify the suite's share/ directory location on particular hosts.
 help=A new line delimited list of PATTERN=DIR pairs.
--- a/lib/python/rose/suite_run.py
+++ b/lib/python/rose/suite_run.py
@@ -237,7 +237,8 @@ class SuiteRunner(Runner):
             now_str = None
             if not opts.install_only_mode and not opts.local_install_only_mode:
                 now_str = datetime.utcnow().strftime("%Y%m%dT%H%M%SZ")
-                self._run_init_dir_log(opts, now_str)
+                self._run_init_dir_log(opts, suite_name, conf_tree,
+                                       now_str=now_str, locs_conf=locs_conf)
             self.fs_util.makedirs("log/suite")
 
             # Rose configuration and version logs
@@ -392,6 +393,7 @@ class SuiteRunner(Runner):
                 shcommand += ",now-str=%s" % now_str
             host_confs = [
                 "root-dir",
+                "root-dir{log}",
                 "root-dir{share}",
                 "root-dir{share/cycle}",
                 "root-dir{work}"]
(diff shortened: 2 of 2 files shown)


hjoliver:
Note: the original way to distinguish "named" and old log dirs from new ones, was to distinguish symlinks from directories.  This doesn't work if the log dirs can also be symlinks.  Therefore to identify new logs, I used the known new log names, and to distinguish named logs I changed the named log symlink naming convention from `log.NAME` to `log-NAME`. 

Rose team:
 - will the above named-log change adversely effect anything?
 - anything else I haven't thought of? (I haven't dabbled very often in the Rose codebase!)
 - can you advise what tests might be required for this? (or help out with them)?

Tested with named logs, multi-run log archiving, remote job host with non-shared FS (via other user).

matthewrmshin:
@hjoliver Do you still want to pursue this? If so, it will need re-base.

hjoliver:
> @hjoliver Do you still want to pursue this? If so, it will need re-base.

Yes, we need it at NIWA.

Unfortunately I must have missed the notification of @sadielbartholomew 's review comments earlier...

hjoliver:
(Still haven't found the time to get back this, sorry ... will keep trying)

hjoliver:
(NIWA is surviving with a workaround for this problem, so I think we can close this PR so long as the functionality makes it into Cylc 8 during `rose suite-run` migration (as per the proposal doc on cylc-admin).

hjoliver:
Closing as fixed in Cylc 8.

PR #2299, opened 2019-02-21
docs: fix napoleon URLs

oliver-sanders:


sadielbartholomew:
(Trivial so appropriate to merge on one review only.)

oliver-sanders:
> The links as they are in master redirect

The links are still fine but a 304 redirect will cause the linkcheck in the test battery to fail.

sadielbartholomew:
I see now, thanks for the explanation @oliver-sanders.

PR #2315, opened 2019-03-14
Python3 write safe (Codacy dislikes pytest)

wxtim:
Spun out _write_safe method into a new unicode_utils module


lib/python/rose/unicode_utils.py
```diff
@@ -0,0 +1,53 @@
...
+            Message to be written - type uncertain.
+        handle (file handle):
+            Potentially Obscure file or file-like object. May include a variety of
+            buffer types.
+    """
+    if isinstance(handle, io.BufferedWriter) and isinstance(msg, str):
+        handle.write(msg.encode(ENCODING))
+    elif isinstance(handle, io.StringIO) and isinstance(msg, bytes):
+        handle.write(msg.decode())
+        print('N')
```
matthewrmshin: Debug left over?

lib/python/rose/unicode_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    elif isinstance(handle, io.StringIO) and isinstance(msg, str):
+        handle.write(msg)
+    else:
+        raise TypeError(f'Unsupported types:'
+          f'\t{msg} has type {type(msg)}'
+          f'\t{handle} has type {type(handle)}')
+
+
+if __name__ == "__main__":
+    pass
```
matthewrmshin: No point to have this block?

lib/python/rose/unicode_utils.py
```diff
@@ -0,0 +1,53 @@
...
+        handle.write(msg.decode())
+        print('N')
+    elif isinstance(handle, io.BufferedWriter) and isinstance(msg, bytes):
+        handle.write(msg)
+    elif isinstance(handle, io.StringIO) and isinstance(msg, str):
+        handle.write(msg)
+    else:
+        raise TypeError(f'Unsupported types:'
+          f'\t{msg} has type {type(msg)}'
+          f'\t{handle} has type {type(handle)}')
```
matthewrmshin: Not a good day for polymorphism, but I don't really have any good alternate suggestion.

lib/python/rose/unicode_utils.py
wxtim: Yes.

lib/python/rose/unicode_utils.py
wxtim: None at all.

wxtim:
#### RE: `assert==value` pytests fail in codacy
https://github.com/fossasia/query-server/issues/332 seems informative, if not terribly helpful (all the links to codacy are broken).

.bandit
```diff
@@ -0,0 +1 @@
+skips: ['B101']
```
oliver-sanders: I'm happy with this as we have little to gain from running using Python's [so-called] optimisation.

PR #2316, opened 2019-03-15
travis.yml: cylc no longer requires make version

matthewrmshin:


matthewrmshin:
One review good enough for this.

PR #2323, opened 2019-04-18
metadata: clarify mini-language 'in' operator behaviour

sadielbartholomew:
Several users have been confused about the behaviour of logical tests involving the membership operator ``in``. Specifically, ultimately many have assumed (e.g. ``trigger=env=``...) ``SETTING: this in 'VALUE'`` would evaluate  identically to ``SETTING: 'VALUE'``.

Looking at the documentation, I think we can make the Pythonic nature of ``in`` clearer by segregating it as a distinct test under an explicit "membership" title & providing some examples, as attempted here. Especially since many (most?) of our users are most familiar with FORTRAN, which doesn't seem to have an equivalent/similar operator, so it's behaviour might be non-intuitive to them.

Please let me know if I can make this clearer in any way. I wasn't sure if it was best to use an example with the smallest list allowing for relevant clarification (``this in ['that']``, as I went for) or to include more than one value to test against in the list, which would be the usual usage scenario. Thoughts would be welcome on that.

Feel free to demote the milestone, which I chose since this is documentation-based & trivial.

### DIFF
--- a/sphinx/api/configuration/metadata.rst
+++ b/sphinx/api/configuration/metadata.rst
@@ -1106,6 +1106,14 @@ simple rules for variable values.
 Operators
 ^^^^^^^^^
 
+The following *logical* operators are supported:
+
+.. code-block:: python
+
+   and   # Logical AND
+   or    # Logical OR
+   not   # Logical NOT
+
 The following *numeric* operators are supported:
 
 .. code-block:: python
@@ -1125,8 +1133,6 @@ The following *string* operators are supported:
    +      # concatenate - e.g. "foo" + "bar" implies "foobar"
    *      # self-concatenate some number of times - e.g. "foo" * 2 implies "foofoo"
    %      # formatting - e.g. "foo %s baz" % "bar" implies "foo bar baz"
-   in     # contained in (True/False) - e.g. "oo" in "foobar" implies True
-   not in # opposite sense of in
 
    # Where m, n are integers or expressions that evaluate to integers
    # (negative numbers count from the end of the string):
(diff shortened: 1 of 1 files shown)


sphinx/api/configuration/metadata.rst
```diff
@@ -1155,9 +1153,39 @@ The following *comparison* operators are supported:
...
+   in     # contained in (True/False)
+   # e.g. "oo" in "foobar" implies True; "foo" in ["bar", "baz"] implies False
+
+   # To test for lack of membership, combine with the "not" logical operator:
+   not in # not contained in, i.e. the opposite to that described above
+
+.. tip::
+   The ``in`` operator functions as it does in Python, testing for membership
+   in a collection which can be a sequence or string. This means that you
+   must contain value(s) to test for in a list, like so:
```
wxtim: suggestion (substitute your own preferred nouns, I was just thinking about :cheese: )

Instead of "This means..."

In some cases ``item in string`` will appear to behave the same way as ``item in [list]``:
```python
>>> "cheddar" in ["cheddar", "brie"]
True
>>> "cheddar" in "roquefort cheddar"
True
```
but this is risky, because ``item in "string"`` matches **any** substring.
```python
>>> "fort cheddar" in "roquefort cheddar"
True
```
Safer, is to match a list item:
```python
# whereas searching a list will look to match an entire list item.
>>> "c" in ["roquefort", "cheddar"]
False

#

sadielbartholomew:
Thanks for kindly reviewing @wxtim. With you stepping in as reviewer I have removed Matt.

> Travis will pass after rebasing onto master.

The fail instance you reference here was from a single sub-test which is a known "flaky test" (#2282). Travis has now passed after few restarts of the build. I don't think I need to rebase onto ``master``, since I seem to be up-to-date, as discussed offline. 

Your suggestion about the list is really useful, & looking back I agree that:

> Instead of "This means..."

should change, & in fact it seems presumptuous. I'll make some changes in response to your feedback.

PR #2325, opened 2019-04-24
Rose fileinstall async (failing codecov)

wxtim:
Early efforts to implement rose-fileinstall with asyncio.

## Evidence that this is doing something worthwhile:
I created a wee script
```bash
#!/bin/bash
# slow_svn - a deliberately sabotaged version of SVN
sleep 10
svn "$@"
exit
```
and changed the reference in `loc_handlers/svn.py` to use this script rather than SVN proper. Using a test setup from `t/rose-app-run/11-file-incr-2.t` I ran:
`/usr/bin/time -v rose app-run --config=../config -q`

Which gave the following results:

| Measure | Master | This Branch |
| ---| ---:| ---: |
| User time (seconds) | 12.16 | 12.71 |
| System time (seconds) | 1.72 | 1.90 |
| Percent of CPU this job got | 15% | 58% |
| Elapsed (wall clock) time (h:mm:ss or m:ss) | 1:31.83 | 0:24.85 |

### DIFF
--- a/lib/python/rose/config_processors/fileinstall.py
+++ b/lib/python/rose/config_processors/fileinstall.py
@@ -38,6 +38,7 @@ from io import BytesIO
 import sys
 from tempfile import mkdtemp
 from urllib.parse import urlparse
+import aiofiles
 
 
 class ConfigProcessorForFile(ConfigProcessorBase):
@@ -112,7 +113,6 @@ class ConfigProcessorForFile(ConfigProcessorBase):
                 exc = FileOverwriteError(name)
                 raise ConfigProcessError([key], None, exc)
             self.manager.fs_util.makedirs(self.manager.fs_util.dirname(name))
-
         # Gets a list of sources and targets
         sources = {}
         targets = {}
@@ -352,12 +352,12 @@ class ConfigProcessorForFile(ConfigProcessorBase):
             event = ChecksumEvent(target.name, target.paths[0].checksum)
             self.handle_event(event)
 
-    def process_job(self, job, conf_tree, loc_dao, work_dir):
+    async def process_job(self, job, conf_tree, loc_dao, work_dir):
         """Process a job, helper for "process"."""
         for key, method in [(Loc.A_INSTALL, self._target_install),
                             (Loc.A_SOURCE, self._source_pull)]:
             if job.context.action_key == key:
-                return method(job.context, conf_tree, work_dir)
+                return await method(job.context, conf_tree, work_dir)
 
     @classmethod
     def post_process_job(cls, job, conf_tree, loc_dao, work_dir):
@@ -371,7 +371,7 @@ class ConfigProcessorForFile(ConfigProcessorBase):
         except AttributeError:
             pass
 
-    def _source_pull(self, source, conf_tree, work_dir):
+    async def _source_pull(self, source, conf_tree, work_dir):
         """Pulls a source to its cache in the work directory."""
         source.cache = os.path.join(
             work_dir,
@@ -380,16 +380,17 @@ class ConfigProcessorForFile(ConfigProcessorBase):
             # and filesystem safe identifier for a source name (which could be
             # a url).
             get_checksum_func()(BytesIO(source.name.encode())))
-        return self.loc_handlers_manager.pull(source, conf_tree)
+        return await self.loc_handlers_manager.pull(source, conf_tree)
 
-    def _target_install(self, target, conf_tree, work_dir):
+    async def _target_install(self, target, conf_tree, work_dir):
         """Install target.
 
         Build target using its source(s).
         Calculate the checksum(s) of (paths in) target.
 
         """
-        handle = None
+        CHUNK_SIZE = 1024 * 4
+        tgt_handle = None
         mod_bits = None
         is_first = True
         # Install target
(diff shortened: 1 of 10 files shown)


matthewrmshin:
Well done!

oliver-sanders:
> Elapsed (wall clock) time (h:mm:ss or m:ss) 1:31.83 => 0:24.85

Woohoo!

lib/python/rose/popen.py
```diff
@@ -271,6 +272,57 @@ def which(name):
...
+
+        If kwargs["stdin"] is a str, communicate it to the command via a pipe.
+
+        """
+        proc = await self.run_bg_async(*args, **kwargs)
+        stdin = None
+        if isinstance(kwargs.get("stdin"), str):
+            stdin = kwargs.get("stdin")
+        stdout, stderr = await proc.communicate(stdin)
+        return stdout, stderr
```
oliver-sanders: Should this return `(ret_code, stdout, stderr)` for consistency with `run()`?

lib/python/rose/popen.py
```diff
@@ -271,6 +272,57 @@ def which(name):
...
+
+    async def run_ok_async(self, *args, **kwargs):
+        stdout, stderr = await self.run_async(*args, **kwargs)
```
oliver-sanders: Should this check the `ret_code` to ensure success (or do failures get detected some other way [exception handlers])?

oliver-sanders:
Oh, and could you stick the `aiofiles` dependency into `rose-check-software` (the docs installation page will auto-update to reflect this).

lib/python/rose/popen.py
```diff
@@ -271,6 +272,57 @@ def which(name):
...
+            kwargs["stdin"] = asyncio.subprocess.PIPE
+        elif stdin is None:
+            kwargs["stdin"] = open(os.devnull)
+        self.handle_event(RosePopenEvent(args, stdin))
+        sys.stdout.flush()
+        try:
+            if kwargs.get("shell"):
+                proc = Popen(args[0], **kwargs)
+            else:
+                command = ' '.join(args)
```
oliver-sanders: `asyncio.create_subprocess_shell` appears to be completely independent of the `subprocess` library, it is roughly equivalent to using `Popen` with the `shell=True` argument which has safety concerns which we will have to address.

The best I can come up with is this, it's safe though may lead to some cryptic errors when misused:

```python
command = ' '.join(map(shlex.quote, args))
```

```python
>>> args = ['sleep', '1', ';', 'echo', 'foo']
>>> ' '.join(map(shlex.quote, args))
"sleep 1 ';' echo foo"

>>> args = ['sleep', '1;', 'echo foo']
>>> ' '.join(map(shlex.quote, args))
"sleep '1;' 'echo foo'"
```

lib/python/rose/config_processors/fileinstall.py
```diff
@@ -405,17 +404,15 @@ def _target_install(self, target, conf_tree, work_dir):
...
-                    try:
-                        handle.write(bytes_.encode())
-                    except AttributeError:
-                        handle.write(bytes_)
-                source_handle.close()
+                async with aiofiles.open(
+                    source.cache, mode='rb'
+                ) as source_handle:
+                    bytes_ = await source_handle.read()
+                # Deal with possibility of different types of handle
```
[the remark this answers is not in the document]
wxtim: Fixed in next commit

lib/python/rose/config_processors/fileinstall.py
```diff
@@ -405,17 +404,15 @@ def _target_install(self, target, conf_tree, work_dir):
...
+                try:
+                    handle.write(bytes_.encode())
```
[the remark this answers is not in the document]
wxtim: When you say "Just copy the file" do you mean run a subprocess with cp?

lib/python/rose/popen.py
wxtim: I'm not sure if `ret_code` actually used. If it is, then this code is not tested properly, because putting it in breaks the tests.

lib/python/rose/config_processors/fileinstall.py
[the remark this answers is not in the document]
oliver-sanders: That would be one option, yes.

lib/python/rose/popen.py
oliver-sanders: `ret_code` isn't used in this new code which makes me think that a subprocess failure won't get caught?

I think `ret_code` it should be checked in `run_ok_async` to mirror the synchronous implementation.

wxtim: Yes, I'm a bit slow today. I've now fixed this. :smile: :sleeping:

lib/python/rose/popen.py
wxtim: :worried:

lib/python/rose/popen.py
```diff
@@ -296,7 +296,7 @@ def which(name):
             if kwargs.get("shell"):
                 proc = Popen(args[0], **kwargs)
             else:
-                command = ' '.join(args)
+                command = ' '.join(map(shlex.quote, args))
```
oliver-sanders: @matthewrmshin could you make sure you're OK with this and re-approve.

matthewrmshin: OK with this.

lib/python/rose/config_processors/fileinstall.py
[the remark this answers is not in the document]
oliver-sanders: We still seem to be doing `try: write str; except: write bytes`.

lib/python/rose/config_processors/fileinstall.py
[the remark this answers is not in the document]
oliver-sanders: This seems just to be a Python copy implementation. After talking with @matthewrmshin we think this might pre-date support for copyfiles in the Python standard library.

I'm a bit worried that people will use Rose to copy huge data files. As this implementation loads the whole file into memory if used with whopping files this could cause swapping or crashes.

Sadly neither `aiofiles` or `asyncio` provide a copy utility, however, as luck would have it `aiofiles` supports reading files in chunks so we can keep the original low-footprint approach.

Here is my Python copy utility using `aiofiles`:

```python
import aiofiles
import asyncio


async def async_copy(src, tgt, chunk_size=1024):
    async with aiofiles.open(tgt, 'bw+') as tgt_handle:
        async with aiofiles.open(src, 'br') as src_handle:
            while True:
                bytes_ = await src_handle.read(chunk_size)
                if not bytes_:
                    break
                await tgt_handle.write(bytes_)


if __name__ == '__main__':
    import sys
    asyncio.run(async_copy(sys.argv[1], sys.argv[2]))
```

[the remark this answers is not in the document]
wxtim: Thank you @oliver-sanders for bailing me out on this this one. The lesson is to be very wary of `handle = aiofiles.open()` - if not awaited you don't get the correct object back and attempts to read and write will fail. Miserably. Ensure any `handle.close()` is also prefixed by an await.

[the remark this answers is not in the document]
oliver-sanders: @matthewrmshin one last thing for you to double check...

Two changes:

* files read and written in chunks using asyncio/aiofiles
* files now read in 4098 byte chunks rather than `f_bsize`

wxtim:
#professionalEvidence

PR #2327, opened 2019-04-26
Handle cylc-flow as a module

matthewrmshin:
Now that cylc/cylc-flow#2990 is merged, we should install cylc-flow in a different way for Travis CI.


wxtim:
Happy.

PR #2311, opened 2019-03-14
Implement threading to speed up rose-ana (PR on Master)

stevewardle:
We have a few cases in our test suite of rose-ana taking a long time - typically when doing lots of file comparisons (in the worst case a rose-ana task with ~20 operations had 4 or 5 that took ~10 minutes each, for a total runtime of over an hour)

It seems like an easy win for all rose-ana users to make it able to run operations in parallel.  Given the way the analysis tasks are written (as classes/plugins) it seems it would lend itself most nicely to multi-threading rather than multi-process (despite the process approach generally being easier to manage)

PR #2328, opened 2019-04-30
Implement threading to speed up rose-ana (PR on 2019.01.x)

stevewardle:
Replaces #2311

### DIFF
--- a/lib/python/rose/apps/rose_ana.py
+++ b/lib/python/rose/apps/rose_ana.py
@@ -461,13 +461,18 @@ class RoseAnaApp(BuiltinApp):
             # Start threads within the set number of concurrent threads until
             # all have been started
             itask = 0
+            running = []
             while itask < len(threads):
-                if threading.active_count() < n_threads:
+                if len(running) < n_threads:
                     self.reporter(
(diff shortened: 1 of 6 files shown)


stevewardle:
Good news, between @oliver-sanders and myself we've identified the issue with the testing (although not quite a full trace to the actual cause)... The threaded rose_ana task appears to hang when the `CYLC_CONF_PATH` variable is blanked out.  The cause of the hanging is the usage of `threading.active_count()` to determine how many active threads there are

When running without `CYLC_CONF_PATH` blanked, this returns `0` prior to `rose_ana` creating its threads, but for some reason with the variable blanked it returns `4` and hence enters an infinite loop unable to start any threads.  I guess since it is calling through to the underlying system threading library it is returning the total thread count of it's entire process-chain or something like that (why the Cylc variable would affect this remains a mystery)

But the simple solution is to track the thread usage ourselves

matthewrmshin:
(But that is a real mystery. `CYLC_CONF_PATH` determines where Cylc looks for its `global.rc` files. With it blank out, it does not read anything `global.rc`. That should be it. Should have nothing to do with Rose or `threading`.)

PR #2330, opened 2019-05-01
Refactor loading of define overrides

TomekTrzeciak:
For {suite,app,task}-run commands one can override config setting on the command line with `--define [section]key=value` options. This doesn't work, however, for some special, top-level keys like `opts` or `import` because applying these overrides happens too late in the config loading process. This PR removes this limitation.

### DIFF
--- a/lib/python/rose/config.py
+++ b/lib/python/rose/config.py
@@ -97,6 +97,7 @@ from rose.env import env_var_escape
 import shlex
 import sys
 from tempfile import NamedTemporaryFile, TemporaryFile
+from tempfile import NamedTemporaryFile, SpooledTemporaryFile
 
 
 CHAR_ASSIGN = "="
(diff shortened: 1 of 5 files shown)


matthewrmshin:
(You will need to forward port this to master.)

TomekTrzeciak:
> (You will need to forward port this to master.)

No problem. I'll do it after this one gets green light.

matthewrmshin:
Travis CI doc failure unrelated.

matthewrmshin:
Need some tests.

oliver-sanders:
Code looks good, a really basic use case in test form should suffice.

TomekTrzeciak:
Adding the tests wasn't a problem (done now) but I wasn't quite sure what should be the desired behaviour. After discussing with @matthewrmshin we agreed that what probably makes the most sense is to override `import` and `opts` from the called app only - `opts` and `import`s contained in imported apps should be unaffected.

TomekTrzeciak:
Rebased.

TomekTrzeciak:
@oliver-sanders, doc test still failing, looks like this is broken on 2019.01.x branch.

oliver-sanders:
OK, will fix in separate PR.

PR #2334, opened 2019-05-10
fix rosie REST API docs

oliver-sanders:
Fix a couple of broken examples in the docs.

### DIFF
--- a/sphinx/api/rosie-web.rst
+++ b/sphinx/api/rosie-web.rst
@@ -40,7 +40,7 @@ REST API
    Example Request
       .. code-block:: http
 
-         GET http://host/my_prefix/get_known/keys?format=json HTTP/1.1
+         GET http://host/my_prefix/get_known_keys?format=json HTTP/1.1
 
    Example Response
       .. code-block:: json
@@ -91,7 +91,7 @@ REST API
    Return a list of suites matching all search terms.
 
    :arg str prefix: Repository prefix.
-   :param list query: List of queries.
+   :param list q: List of queries.
    :param string format: Desired return format (``json`` or ``None``).
    :param flag all_revs: Switch on searching older revisions of current suites
       and deleted suites.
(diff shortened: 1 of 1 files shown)


sphinx/api/rosie-web.rst
```diff
@@ -152,7 +152,7 @@ REST API
 
       .. code-block:: http
 
-         GET http://host/my_prefix/search?var+bob+nowcast&format=json HTTP/1.1
+         GET http://host/my_prefix/search/var+bob+nowcast&format=json HTTP/1.1
```
matthewrmshin: Why doesn't the old syntax work any more?

oliver-sanders: I don't think this syntax was ever correct, was it?

matthewrmshin: That's true.

oliver-sanders: It's probably either an error copied forward from the old docs or a translation error by yours truely :frowning:

sadielbartholomew: Isn't this meant to instead be:
```suggestion
         GET http://host/my_prefix/search?s=var+bob+nowcast&format=json HTTP/1.1
```

The original, as with your change suggestion, both don't make sense as URL queries, but this does? To check I am not completely mistaken, I tried these URLs out on both the current ("old") & my new (PR #2288) versions of the rosie discovery utility, & the original & change here both give a 404 error.

oliver-sanders:
> The original & change here both give a 404 error

The change doesn't give a 404, for example:

http://fcm1/rosie/mot/search/frsn

However this isn't the user facing page not the REST API! Whoops! `s` is what we were supposed to document here:

http://fcm1/rosie/mot/search?s=frsn

oliver-sanders:
Yep, that should be the last of them.

PR #2338, opened 2019-05-21
Suite variable dictionary containing all variables inserted in template

TomekTrzeciak:
Implements https://github.com/cylc/cylc-flow/issues/2960#issuecomment-466043979 - a special  `ROSE_SUITE_VARIABLES` Jinja2 (EmPy) dictionary containing all variables inserted in template, so that their names can be resolved in a programmatic way.

### DIFF
--- a/bin/rose-suite-run
+++ b/bin/rose-suite-run
@@ -117,11 +117,14 @@
 #         Each `KEY` in this space delimited list switches on an optional
 #         configuration. The configurations are applied first-to-last.
 #
-# JINJA2 VARIABLES
-#     `rose suite-run` provides the following Jinja2 variables to the suite.
+# SUITE VARIABLES
+#     `rose suite-run` provides the following variables to the suite.
 #
 #     ROSE_ORIG_HOST
 #         The name of the host where the `rose suite-run` command was invoked.
+#     ROSE_SUITE_VARIABLES
+#         Dictionary containing all suite variables inserted via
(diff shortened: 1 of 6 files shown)


TomekTrzeciak:
(There are some failures in the tests, but I don't think they are related to this change, e.g. `t/docs/02-tutorial-suites.t` fails for me on clean checkout of `2019.01.x` branch as well.)

bin/rose-suite-run
```diff
@@ -117,11 +117,14 @@
 #         Each `KEY` in this space delimited list switches on an optional
 #         configuration. The configurations are applied first-to-last.
 #
-# JINJA2 VARIABLES
-#     `rose suite-run` provides the following Jinja2 variables to the suite.
+# SUITE VARIABLES
+#     `rose suite-run` provides the following variables to the suite.
```
oliver-sanders: We should make it clear that these variables are template variables not environment variables.

```suggestion
#     `rose suite-run` provides the following variables to the suite as template variables (i.e. Jinja2 or EmPy).
```

*Why this matters: `$ROSE_ORIG_HOST` is actually an environment variable but isn't only exported locally so we advise people use `{{ ROSE_ORIG_HOST }}`.*

oliver-sanders:
Sorry @TomekTrzeciak can I be a pain and ask for a minor wording tweak, otherwise fine.

TomekTrzeciak:
@oliver-sanders, good spot, I haven't thought of possible confusion with environment variables. Fixed on this and master PRs.

PR #2336, opened 2019-05-20
[7.x] cylc tutorial: don't depend on PYTHONPATH being defined

oliver-sanders:
Due to packaging changes Cylc no longer inserts itself PYTHONPATH meaning that PYTHONPATH could potentially be undefined.

This causes tutorial job scripts to break due to `set -u`.

Reported by @wxtim

**[edit]** on master remove `PYTHONPATH` altogether (in a follow-on PR) after https://github.com/cylc/cylc-flow/pull/3168

PR #2340, opened 2019-05-28
tutorial suite: remove pythonpath

oliver-sanders:
Rose 2.0 equivalent of https://github.com/metomi/rose/pull/2336

Since https://github.com/cylc/cylc-flow/pull/3168 went through (with surprisingly little fuss) we can now get rid of all the horrible `PYTHONPATH` mess we have been expecting students to put up with in the tutorial suite.

Will assign reviewers once tests pass.


wxtim:
Looks like a small enough change that I'd be happy to merge.

PR #2343, opened 2019-06-03
reporter: mojibake

oliver-sanders:
This requirement comes through `rose_ana` where non-unicode characters may appear in file differences. Currently this causes Rose to bail out with a `UnicodeDecodeError`.

This change allows the error to pass and dumps non-unicode to screen if decoding fails.


oliver-sanders:
A different fix will be required on master.

oliver-sanders:
Merging on two approvals.

PR #2346, opened 2019-06-06
mojibake fix [Sphinx Fail == Master Sphinx Fail]

### LINKED ISSUE #2345: python3: mojibake
Attempt after #2317 

Allow [mojibake](https://en.wikipedia.org/wiki/Mojibake) to pass through `rose.reporter`.

This has been addressed on 2019.01.x in #2343

Likely a different fix will be required for Python3.

wxtim:
fix #2345 
I'm not actually sure that there is anything to fix in the python 3 version, but I've created a unit test.


lib/python/rose/tests/test_ancils/unicode.txt
```diff
@@ -0,0 +1,301 @@
+UTF-8 decoder capability and stress test
+----------------------------------------
+
+Markus Kuhn <http://www.cl.cam.ac.uk/~mgk25/> - 2015-08-28 - CC BY 4.0
```
wxtim: I think this is an acceptable licence for use to just bung this file in: https://creativecommons.org/licenses/by/4.0/deed.ast

oliver-sanders: May need to add it to `ATTRIBUTIONS.md` to comply.

wxtim: Done

PR #2347, opened 2019-06-07
docs: fix what-is-cylc page

oliver-sanders:
The introduction page seems to be a bit broken, the slides are fine but the regular HTML version is lacking content.

This PR changes from manually defined slides (`autoslides=False`) to automatically defined ones as used throughout the rest of the documentation. The content remains unchanged.

PR #2348, opened 2019-06-07
2019.01.x docs: fix what-is-cylc page

oliver-sanders:
Back-port of #2347, not too urgent but ideally we would slide this into next-release as it is a pretty straight-forward change.


oliver-sanders:
This appears to be unrelated, I expect this is broken on the `2019.01.x` branch, will investigate tomorrow.

Issue #2211, opened 2018-07-16
cylc tutorial: odds and ends


oliver-sanders:
Docs:

* [ ] ~Extensions for practical exercises~
* [ ] ~[Parametrisation tutorial - use parameter environment templates](https://github.com/metomi/rose/blame/2018.06.0/sphinx/tutorial/cylc/runtime/configuration-consolidation/parameters.rst#L318)~
* [ ] ~[In-situ broadcast example](https://github.com/metomi/rose/blame/2018.06.0/sphinx/tutorial/cylc/furthertopics/broadcast.rst#L96)~
* [ ] ~[Update clock-triggers to the new xtrigger system once it is stabilised](https://github.com/metomi/rose/blame/2018.06.0/sphinx/tutorial/cylc/furthertopics/clock-triggered-tasks.rst#L9)~
* [ ] ~[Cylc message tutorial](https://github.com/metomi/rose/blame/2018.06.0/sphinx/tutorial/cylc/furthertopics/suicide-triggers.rst#L520)~
* [ ] ~Run (Rose/Cylc) code examples against the Parsec upgrader to prevent deprecated syntax creeping into the docs.~
* [ ] ~Update `cylc run` commands according to https://github.com/cylc/cylc/pull/2759~

Tutorial Suite:

* [ ] Provide some sample data an instructions to use it so the tutorial suite can be run of-line.
* [ ] Rainfall data can wrap in the y dimension.
* [ ] Tidy.

sadielbartholomew:
The Cylc tutorial has now been moved to its rightful home under Cylc, in cylc/cylc-doc#38, so I am transferring this Issue to that repo accordingly.

sadielbartholomew:
OK, I was hoping to be able to use the new-ish 'Transfer issue' feature to do this, but it only seems to allow me to transfer to other repositories under this (``metomi``) organisation, not to another one, i.e. ``cylc``. (Though I'm struggling to find online documentation to confirm this or otherwise.) @oliver-sanders shall we therefore do a lift 'n shift?

oliver-sanders:
Don't migrate this issue as it needs to be split up between:

* Rose (for rose tutorial)
* Cylc-Doc (for cylc tutorial)
* Cylc-Sphinx-Extensions (for technical gubbins).
* Cylc Tutorial Suite (tutorial suite)

The Cylc Tutorial Suite is currently homeless, it will probably become (yet) another repository.

sadielbartholomew:
[...]

PR #2351, opened 2019-06-11
docs: tutorial tweaks

oliver-sanders:
Address some of the easier to fix snags in the tutorial material.

Note some of the more involved tasks are covered in #2211 and #2212.

Partly address #2212

### DIFF
--- a/sphinx/index.rst
+++ b/sphinx/index.rst
@@ -69,5 +69,7 @@ Other
 .. toctree::
    :hidden:
    :name: hidden-index
+   :glob:
 
    terms
+   course/*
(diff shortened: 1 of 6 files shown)


sphinx/index.rst
```diff
@@ -69,5 +69,7 @@ Other
 .. toctree::
    :hidden:
    :name: hidden-index
+   :glob:
 
    terms
+   course/*
```
oliver-sanders: This creates the `course` directory in which I've put the UM-UT course outline, it won't be visible from the index page but we can link users to it. Will make a note to pass the link to this page onto the UM-UT course coordinators once it is live.

sphinx/tutorial/cylc/scheduling/integer-cycling.rst
```diff
@@ -239,6 +239,13 @@ Inter-Cycle Dependencies
    ingredients, the bakery wants to purchase ingredients two batches ahead.
    This can be achieved by adding the following dependency:
 
+.. ifslides::
+
+   We need ``purchase_ingredients`` to be dependent on ``sell_bread`` from
+   two cycles before.
```
oliver-sanders: Note, this is a presenter prompt to poke the audience for an answer.

PR #2354, opened 2019-06-12
docs: add debug info in travis builds

oliver-sanders:
* Output Sphinx traceback in the event of error.
* When error reporting don't bail on failure.
* Add the `sphinxcontrib-svg2pdfconverter` dependency.

### DIFF
--- a/.travis/now
+++ b/.travis/now
@@ -92,7 +92,8 @@ _install_sphinx () {
     # sphinx documentation and its extensions
     APT+=(latexmk texlive texlive-generic-extra texlive-latex-extra \
           texlive-fonts-recommended graphviz)
-    PIP+=(sphinx sphinx_rtd_theme sphinxcontrib-httpdomain hieroglyph)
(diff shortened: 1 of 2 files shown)


.travis/now
```diff
@@ -161,12 +162,27 @@ install () {
...
     bash <(curl -s https://codecov.io/bash)
 }
 
+_report_error() {
+    # don't bail out on error
+    set +eu
+
+    printenv PATH PYTHONPATH
+    rose check-software
+    cat /tmp/sphinx-err*  # sphinx traceback
```
matthewrmshin: Doesn't matter here, but should redirect error messages to `>&2` as a good practice.

oliver-sanders:
FYI, the error script diddn't work in `.travis.yml` as `.bashrc` doesn't get sourced, `.travis/now` manually sources `.bashrc` to get around this. Another interesting Travis environment quirk to be aware of...

.travis/now
oliver-sanders: Doesn't make a jot of difference but done!

PR #2355, opened 2019-06-13
rosie: restore rosie ws

oliver-sanders:
NOTE: As `sqlalchemy` was installed to system Python rather than Travis Python some tests wern't being run (due to missing dependencies). These have been fixed.

* Restore files removed in 400494415dcd788f956c991c85e1c003f8c4ff4b.
* Re-jig `.travis/now` with a new `RI_PATH` for setting `PATH` in `rose_site_init` in `env -i` environments.
* Fix tests.


matthewrmshin:
Travis CI build failing due to lack of `request` module?

oliver-sanders:
By adding the Rosie dependencies we are now activating tests which might not have been run since Python2.

Rosie tests are falling over in a peculiar way. It would appear `svn_post_commit` is being run under Python3.5 (OS provided) rather than 3.7 (Travis installed).

oliver-sanders:
In #2288 `sqlalchemy` was [installed to system Python](
https://github.com/metomi/rose/pull/2288/files#diff-354f30a63fb0907d4ad57269548329e3R28) via `apt-get`.

oliver-sanders:
Added a `rose_init_site` script to the Travis environment.

oliver-sanders:
Attempted solutions:

1. Source `.bashrc` in `rose_init_site`.
  * Results in test failures as it modifies `$PATH` and `$PYTHONPATH`
2. Source `.bashrc` only if `${HOME}` is not defined.

oliver-sanders:
The tests should pass this time:

* For the moment I've moved `pygraphviz` imports out the way awaiting complete removal.
* I've skipped a failing test which is punted to https://github.com/metomi/rose/issues/2357

sadielbartholomew:
> I've skipped a failing test which is punted to #2357

I had a look-see at this, & it is failing for quite a significant reason: the ``rosie-disco`` web files are not there!

Because all other ``t/rosie-disco/`` sub-tests were testing on ``?format-json`` endpoints, the only two tests that were checking for rendered pages, which use the HTML templates & CSS & JS files from the ``lib/html/`` directory in ``2019.XX`` & previous versions, were the sub-tests for the root index page! And the second of those:

https://github.com/oliver-sanders/rose/blob/a2309eb25e6768d9c0628d9da75e1f3adb9f91b5/t/rosie-disco/00-basic.t#L65-L70

was just testing for a re-direct to the equivalent with the trailing slash, so it passed. But it redirected from a ``500 Internal Sever Error``, as was being returned for the redirected page causing that one sub-test failure.

Presumably at some point the web files required were removed from ``master``. Though we no longer need the subset of them designed for Rose Bush, we still want them for the updated Python 3 Rosie Disco. I will put in a PR to this branch which re-adds the necessary directories (leaving out obvious Rose Bush files), & removes the comment characters on the currently-failing sub-test which should then pass (Travis, based on the local test battery pass I then see).

oliver-sanders:
Thanks Sadie, hopefully this should now pass.

matthewrmshin:
Travis CI unit test failing - need to filter out `lib/html/` hierarchy?

sadielbartholomew:
Ah, & naturally those web files come a bundle of issues as flagged also by Codacy. Can we get them ignored & (re-)raised as an Issue? I can't recall quite how we dealt with those previously...

matthewrmshin:
On Codacy. We should filter out the external libraries from the analysis. Some of our own stuffs can be fixed. The others should be flagged for ignore.

matthewrmshin:
Branch in conflict.

sadielbartholomew:
@oliver-sanders what is the state of this PR, so I know when to review? Is the plan to address those Codacy points? Also I have kicked Travis CI but across a number of tries it keeps returning an error from a syntax error in ``.travis/now``.

oliver-sanders:
PR was still ready for review, tests broken by most recent conflict resolution.

matthewrmshin:
Unit test broken by style check?

oliver-sanders:
The three codacy issues can be ignored:

* The Jinja2 autoescape comment is just wrong as this is set on the following line
* The `input` comments are not relevant to Python3, Tim has ignored these issues in #2360

matthewrmshin:
@wxtim Please perform a final sanity check. The Codacy/Bandit issue can be unflagged in your branch?

PR #2360, opened 2019-06-25
Another rose package branch

wxtim:
## Aim
Turn Rose into a PyPi Package

## Todo List
- [x] Setup.py written.
    - [x] setup.py rechecked removing working stuff
- [x] Basic functionality ported and a reasonable (working) metomi.rose package being installable using pip.
    - [x] Rose Version
    - [x] rose on the command line
    - [x] Replace $ROSE_HOME
        - [x] Mostly Done
        - [x] Final Checks
- [x] Test battery moved to etc/bin/rose-test-battery and modified to work from there.
    - [x] Rose regression tests
        - [x] Locally
        - [x] Travis
    - [x] Pytest made to work on Travis
    - [x] Shellcheck made to work on Travis
    - [x] Doccheck made to work on Travis
- [x] compacted branch history a little.
- [x] Personally Proofread / Sanity check
- [x] Talk to @oliver-sanders about merge strategy for PR #2355 
- [ ] Create follow on issue to fix Rosie.
- [ ] Create follow on issue RE hard-coded auto-generated content see https://github.com/metomi/rose/pull/2360#discussion_r300668575.

## Notes
* Rosie testing has been disabled in this branch, which almost certainly breaks Rosie. This will need to be fixed after PR #2355 has been finished. 

## Important
`rose-test-battery` and `rose-make-docs` have been moved to `etc/bin/$UTILNAME` on the basis that they are for dev use only.


wxtim:
I tried to rebase and I ended up with a conflict. Will have a more careful think about this tomorrow.

.bandit
```diff
@@ -1 +1 @@
-skips: ['B101']
+skips: ['B101', 'B322']
```
wxtim: B322 Refers to the method `input` which the bandit message asserts is safe in Python3, but complains about anyway. https://github.com/PyCQA/bandit/issues/402

.gitignore
```diff
@@ -5,6 +5,7 @@ etc/opt
 lib/bash/rose_init_site
 doc
 venv
+metomi_rose.egg-info
```
wxtim: @kinow Should I be excluding the egg-info file?

.travis.yml
```diff
@@ -22,20 +22,22 @@ jobs:
...
     - name: "Test Battery"
       before_install:
         - export PATH="$PWD/.travis:$PATH"
-        - now install coverage cylc fcm rose tut_suite
+        - now install coverage fcm tut_suite
+      install:
+        - now install isodatetime cylc rose
       script:
         - now test battery
+
```
wxtim: For tidyness only

.travis/now
```diff
@@ -51,7 +51,7 @@ _join () {
 }
 
 _build_docs () {
-    rose make-docs --strict clean html slides latexpdf
+    etc/bin/rose-make-docs --strict clean html slides latexpdf
```
wxtim: I believe that this is a largely dev-only function?

.travis/now
```diff
@@ -60,8 +60,9 @@ _install_coverage () {
 }
 
 _install_rose () {
+    pip install -e .
```
wxtim: @kinow is this a safe/full test, or should we install fully, since that is what the customer will probably do...

.travis/shellchecker
```diff
@@ -79,12 +79,13 @@ default () {
...
         --exclude etc/rose-bash-completion \
         --exclude t \
-        -- -e SC1090
+        -- -e SC1090 -e SC2119
 
     # run a lenient check on all test scripts
     main t -- -S error -e SC1090
 }
 
+shellcheck -V
```
wxtim: I put this in for debugging, but outputting the version of shellcheck used seems like a good idea. Thoughts?

bin/rose
```diff
@@ -37,7 +37,9 @@
 if ${ROSE_DEBUG:-false}; then
     set -x
 fi
-. "$(dirname "$0")/../lib/bash/rose_init"
+# shellcheck source=lib/bash/rose_init
```
wxtim: Apparently shellcheck and pip don't play nicely.

metomi/rose/apps/rose_ana.py
```diff
@@ -34,16 +34,17 @@
...
+from metomi.rose.reporter import Reporter
+from metomi.rose.resource import ResourceLocator
+from metomi.rose.app_run import BuiltinApp
+from metomi.rose.env import env_var_process
 
 
 class KGODatabase(object):
     """
-    KGO Database object, stores comparison information for rose_ana apps.
+    KGO Database object, stores comparison information for metomi.rose_ana
```
wxtim: Frustratingly a lot of this diff is to prevent to addition of `metomi.` from making `len(lines) > 79`.

.bandit
oliver-sanders: Fine.

.travis/now
oliver-sanders: Fine, we will probably change this in-line with https://github.com/cylc/cylc-doc/pull/26 anyway.

.travis/now
oliver-sanders: This is what we do for Cylc so should be fine?

.travis/shellchecker
oliver-sanders: Whatever.

If desired you could put it in the Travis "after failure" section (ps: Matt will ask you to send this output to stderr).

bin/rose
oliver-sanders: > Apparently shellcheck and pip don't play nicely.

???

.gitignore
oliver-sanders: Yep.

.travis/now
kinow: Editable install is more likely to get broken. That's why I cjose to build like that in travis. So if not concerned about editable installs failing, you can use the normal approach instead

t/rosa-db-create/00-basic.t
```diff
@@ -22,6 +22,8 @@
 # svn-post-commit", which is tested quite thoroughly in its own test suite.
 #-------------------------------------------------------------------------------
 . $(dirname $0)/test_header
+skip_all # Skipped pending fix
```
sadielbartholomew: Why is ``skip_all`` called twice for many of the skipped tests?

t/rose.config_tree/00-self.t
```diff
@@ -17,4 +17,5 @@
 # You should have received a copy of the GNU General Public License
 # along with Rose. If not, see <http://www.gnu.org/licenses/>.
 #-------------------------------------------------------------------------------
-exec python3 $ROSE_HOME/lib/python/rose/config_tree.py "$@"
+from metomi.rose.config_tree import _Test
+_Test().run()
```
sadielbartholomew: Pop a newline at the end of this file, to avoid nastiness that can sometimes occur without.

metomi/rose/__init__.py
```diff
@@ -127,3 +127,5 @@
 
 # Paths in the Rose distribution.
 FILEPATH_README = "README.md"
+
+__version__ = "2.0.0alpha"
```
sadielbartholomew: :smile: Rose 2 here we come!

sphinx/api/command-reference.rst
```diff
@@ -9,6 +9,59 @@ Rose Commands
 
 .. auto-cli-doc:: rose rose
 
+----
+
+.. _command-rose-test-battery:
```
sadielbartholomew: The command reference is auto-generated & should now be excluded from the content under version control. If you have added new description for a command, it needs to be put with that command's module file so it gets printed to STDOUT with the ``--help`` option, & then the script to parse will pick it up & add it to this rendered file in the HTML docs.

oliver-sanders: @sadielbartholomew this is here because the command has been moved to `etc/` so the `.. auto-cli-doc:: rose rose` directive is no longer able to document it so Tim has added in the derived content for just the test battery command for now (as there are references to it elsewhere in the docs).

sadielbartholomew: Thanks for the explanation. I'll add a tick-box note to the main PR comment as a reminder to create a follow-on Issue for this. (@wxtim in future could important background like this be popped onto the main PR comment to aide with reviewing?)

bin/rose
wxtim: Shellcheck objects to the script being sourced because PIP puts it somewhere shellcheck doesn't think to look.

metomi/rose/apps/rose_ana.py
wxtim: @kinow  - is that a confused face or a sad face?
if confused:
```python
# If a line is > 72 chars long...
from rose.verylongname import AReallyQuiteExtraordinarilyLongClassName

# ... metomi. will push it over 79 chars...
from metomi.rose.verylongname import AReallyQuiteExtraordinarilyLongClassName

# ... so will need a bit of refactoring/jiggery-pokery
from metomi.rose.verylongname import (
    AReallyQuiteExtraordinarilyLongClassName
)
```

t/rosa-db-create/00-basic.t
wxtim: Probably a `Tim: Cannot use SED competently error`

sphinx/api/command-reference.rst
wxtim: Yes. sorry.

.travis/now
wxtim: stet

wxtim:
> Some comments from a quick preliminary check. The PyPI installation seems to work grinning: 
> `Successfully installed metomi-rose-0.0.1`!

**Hooray!** :fireworks: 

---

> Some commands e.g. `rose suite-run` & `rose check-software` seem to work, 
> but others e.g. `rose make-docs` & `rose test-battery` are not recognised, giving 
> this error format:

Have been replaced by `etc/bin/rose-app-name`. My fault for not putting a note in the initial PR comment explaining what has happened to these items. 

---

>     * It seems we're going from `lib/python/` to `metomi/` as the sub-directories 
> to hold the `rose` (etc.) Python modules (Why have we gone for that naming, 
> out of interest?) 

Rose is taken. We wanted an umbrella name-space. I don't think the reasoning was any more than that.

---

> From a `grep` there still are a lot of references to `lib/python/`, many of which
> need amending (e.g. paths written into the docs).

Well spotted, I'll have a look at that!

---

>     * The PyPI listing for `metomi-rose` has a few broken links ('How to Contribute'
> , 'Acknowledgement for Non-Rose Work')

This branch has not been loaded onto PyPi. Can you tell me what you did to test this?

---

> There are some files under `HEAD` & `vnX` directories, which doesn't 
> seem right (unless it is some test battery facet I have not seen before)?

These belong to test battery (mainly `t/rose-macro/17-import.t` and rose-task-run for rose arch and prune). It looks like it's for testing macros dealing with multiple versions. These exist in master: I've only moved them.
It appears to be a rose macro for the creation of baked Alaska. :cake: :ice_cream:

metomi/rose/apps/rose_ana.py
oliver-sanders: `AReallyQuiteExtraordinarilyLongClassName`, rename the class!

oliver-sanders: Also it isn't `metomi.rose_ana` it's just `rose_ana` or "Rose Ana".

wxtim: > `AReallyQuiteExtraordinarilyLongClassName`, rename the class!

This was a made-up example to illustrate the problem. Your solution is of course correct and the class should be re-named. The point still stands though that the addition of metomi. requires a change where the magnitude of the diff is very much greater than the significance of the change.

wxtim:
## Remaining instances of `lib/python`
* In the test folder: All instances of `lib/python` are inside the test dir, indicating that these are not references to $ROSE_HOME/lib/python but to $TEST_DIR/suite/or/test/lib/python. These should not be changed.

* In sphinx folder - 

* In `metomi/rose/resource.py`. This only appears in a comment, but should be changed

* In docs: Should follow sphinx fix.

Working on this, will update comment when done.

wxtim:
@sadielbartholomew @oliver-sanders  & @kinow - I have got the branch of this branch with the Rosie and Rosa tests in it working PR #2366 so depending on how you feel you may wish to deprecate this PR and have a look at both together. If you choose not to the second PR will have a much smaller diff. :grinning:

wxtim:
This PR deprecated in favour of #2366

sadielbartholomew:
Thanks for your detailed response to my comments & questions, @wxtim. Sorry for the delay in responding (delivering User Training is quite intensive). Some responses & further detail:

> Rose is taken. We wanted an umbrella name-space. I don't think the reasoning was any more than that.

For the record of anyone following, I didn't know this but as Tim & Matt explained offline, for PyPI namespacing, there are requirements on the directory names at the top level under the entire repository directory. Plain ``Rose`` is taken under PyPI so we are going for ``metomi.rose`` as the Rose module & hence we need to use the sub-directory name ``metomi`` to contain the Rose modules.

> This branch has not been loaded onto PyPi. Can you tell me what you did to test this?
>
> I had a look on Pypi and I've made these link to the corresponding file on Github because
> ...

Ah, yes, my apologies since that point wasn't strictly related to this PR, but when I went to see if we had reserved the namespace up on PyPI, I did notice that, & I am not down as a maintainer for the package so couldn't go in & fix it myself. I've re-tested the links after your changes & they now work, thanks.

> I think we might prefer to have a separate PYPIREADME.md

Eventually, perhaps, but for now the current set up is more than sufficient, I think.

PR #2361, opened 2019-07-09
doc: tutorial tweaks

oliver-sanders:
Address some snag points in the tutorial material.

### DIFF
--- a/sphinx/tutorial/rose/metadata.rst
+++ b/sphinx/tutorial/rose/metadata.rst
@@ -66,7 +66,8 @@ The Metadata Format
 
    [env=WORLD]
    description=The name of the world to say hello to.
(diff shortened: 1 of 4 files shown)


sphinx/tutorial/rose/metadata.rst
```diff
@@ -66,7 +66,8 @@ The Metadata Format
 
    [env=WORLD]
    description=The name of the world to say hello to.
-   values=Mercury, Venus, Earth, Mars, Jupiter, Saturn, Uranus, Neptune
+   values=Mercury, Venus, Earth, Mars, Jupiter,
+         =Saturn, Uranus, Neptune
```
sadielbartholomew: To stop the text getting cut off by the box end, I am sure, but this will also be a nice reminder on multi-line value usage, which is mentioned for the Rose config format but then not used in any of the examples.

PR #2364, opened 2019-07-09
doc: fix app poll delays syntax

matthewrmshin:
Correct syntax is PT0S not 0.


oliver-sanders:
1 review sufficient for this small doc change.

PR #2391, opened 2019-10-30
rosie: only raise SuiteIdTextError if there is actually a suite id error

oliver-sanders:
Rosie unhelpfully raises a `SuiteIdTextError` upon any error in `svn info <location>`.

Whilst error could mean an invalid suite name it could also mean a connection problem e.g. authentication.

I've seen this recently with the `rosie copy` command.

TODO:

* [ ] Forward port to Rose 2
* [ ] Test?

### DIFF
--- a/lib/python/rosie/suite_id.py
+++ b/lib/python/rosie/suite_id.py
@@ -515,8 +515,12 @@ class SuiteId(object):
             try:
                 info_entry = info_parser.parse(
                     self.svn("info", "--xml", location))
-            except RosePopenError:
-                raise SuiteIdTextError(location)
+            except RosePopenError as exc:
+                if 'E200009' in exc.stderr:
+                    # E200009: Could not display info for all targets
+                    #          because some targets don't exist
+                    raise SuiteIdTextError(location)
+                raise
             else:
                 if "commit:revision" in info_entry:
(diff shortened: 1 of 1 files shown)


dpmatthews:
Just ran a test with GNOME keyring locked so Subversion couldn't access it's cached password.

Using Rose 2019.01.2:
```
$ rosie id --next u
<snip>
rose.popen.RosePopenError: svn ls https://code.metoffice.gov.uk/svn/roses-u # return-code=1, stderr=
svn: E170013: Unable to connect to a repository at URL 'https://code.metoffice.gov.uk/svn/roses-u'
svn: E215000: GNOME Keyring is locked and we are non-interactive
```
That's OK - we can see the error reported by Subversion.

```
$ rosie copy u-xx999
<snip>
rosie.suite_id.SuiteIdTextError: https://code.metoffice.gov.uk/svn/roses-u/x/x/9/9/9/trunk@HEAD: invalid suite ID
```
Not OK - no indication there is a problem accessing the repository.

With the proposed change:
```
$ rosie copy u-xx999
<snip>
rosie.suite_id.SuiteIdPrefixError: u: cannot determine prefix location
```
Different but no better :-(

PR #2395, opened 2020-01-14
2019.1.x steve ws pr

wxtim:
This is the same change as merged to master (PR2394)[for https://github.com/metomi/rose/pull/2394]


wxtim:
Looks like it's broken in the same ways as rose master. :(

PR #2398, opened 2020-03-17
Fix tutorial

MetRonnie:
I had a couple of problems with the cylc & rose tutorials, specifically with the [cylc runtime configuration tutorial](https://metomi.github.io/rose/2019.01.2/html/tutorial/cylc/runtime/runtime-configuration.html#admonition-0) and the [rose suite configuration tutorial](https://metomi.github.io/rose/2019.01.2/html/tutorial/rose/suites.html#admonition-5).

This PR fixes:
- HTTP 403 error in the weather data API request
- KeyError that sometimes happens when no wind direction/speed data is served (thanks to Joe)

### DIFF
--- a/etc/tutorial/cylc-forecasting-suite/bin/get-observations
+++ b/etc/tutorial/cylc-forecasting-suite/bin/get-observations
@@ -56,9 +56,20 @@ def get_datapoint_data(site_id, time, api_key):
     time = datetime.strptime(time, '%Y%m%dT%H%MZ').strftime('%Y-%m-%dT%H:%MZ')
     url = BASE_URL.format(time=time, site_id=site_id,
                           api_key=api_key)
+    # HTTP header to prevent 403 error:
+    hdr = {'User-Agent': ('Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.11 '
+                          '(KHTML, like Gecko) Chrome/23.0.1271.64 '
+                          'Safari/537.11'),
+           'Accept': ('text/html,application/xhtml+xml,application/xml;'
+                      'q=0.9,*/*;q=0.8'),
+           'Accept-Charset': 'ISO-8859-1,utf-8;q=0.7,*;q=0.3',
+           'Accept-Encoding': 'none',
+           'Accept-Language': 'en-US,en;q=0.8',
+           'Connection': 'keep-alive'}
+    req = urllib2.Request(url, headers=hdr)
     # Get the data and parse it as JSON.
     print 'Opening URL: %s' % url
-    return json.loads(urllib2.urlopen(url).read())['SiteRep']['DV']['Location']
+    return json.loads(urllib2.urlopen(req).read())['SiteRep']['DV']['Location']
 
 
 def get_archived_data(site_id, time):
--- a/etc/tutorial/cylc-forecasting-suite/bin/get-observations
+++ b/etc/tutorial/cylc-forecasting-suite/bin/get-observations
@@ -83,10 +83,22 @@ def get_archived_data(site_id, time):
 
 def extract_observations(data):
     """Extract the relevant observations from the weather data."""
+    try:
+        direction = ORDINATES[data['Period']['Rep']['D']]
+    except KeyError:
+        direction = '0'
+        print 'No wind direction in file so using 0'
+
+    try:
+        speed = data['Period']['Rep']['S']
+    except KeyError:
+        speed = '0'
+        print 'No wind speed in file so using 0'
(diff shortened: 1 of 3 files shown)


MetRonnie:
> The tutorial docs have now been migrated to cylc/cylc-doc so you could also raise a PR there to get two for the price of 1 :)

@datamel Aren't the docs are unaffected?

etc/tutorial/cylc-forecasting-suite/bin/get-observations
```diff
@@ -56,9 +56,20 @@ def get_datapoint_data(site_id, time, api_key):
...
+    hdr = {'User-Agent': ('Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.11 '
+                          '(KHTML, like Gecko) Chrome/23.0.1271.64 '
+                          'Safari/537.11'),
+           'Accept': ('text/html,application/xhtml+xml,application/xml;'
+                      'q=0.9,*/*;q=0.8'),
+           'Accept-Charset': 'ISO-8859-1,utf-8;q=0.7,*;q=0.3',
+           'Accept-Encoding': 'none',
+           'Accept-Language': 'en-US,en;q=0.8',
+           'Connection': 'keep-alive'}
+    req = urllib2.Request(url, headers=hdr)
```
oliver-sanders: Dammit, we are going to have to move away from DataPoint soon.

etc/tutorial/cylc-forecasting-suite/bin/get-observations
```diff
@@ -72,10 +83,22 @@ def get_archived_data(site_id, time):
...
+    except KeyError:
+        direction = '0'
+        print 'No wind direction in file so using 0'
+
+    try:
+        speed = data['Period']['Rep']['S']
+    except KeyError:
+        speed = '0'
+        print 'No wind speed in file so using 0'
+
```
oliver-sanders: I'm not so sure about this solution as it will cause the resultant wind field to be zero which will make it look like the forecast task is not doing anything.

Then again this isn't the expected case and it's better to have no wind than a broken workflow.

etc/tutorial/cylc-forecasting-suite/bin/get-observations
```diff
@@ -72,10 +83,22 @@ def get_archived_data(site_id, time):
...
+    try:
+        direction = ORDINATES[data['Period']['Rep']['D']]
+    except KeyError:
+        direction = '0'
+        print 'No wind direction in file so using 0'
+
+    try:
+        speed = data['Period']['Rep']['S']
+    except KeyError:
+        speed = '0'
```
oliver-sanders: I can't remember where type handling gets handled in these scripts, should this be `0` rather than `'0'`?

oliver-sanders:
Poking Travis to see if the test battery will pass second time.

The test battery actually runs the tutorial suite so it passing should test your code.

Because of the `[skip travis]` commit the test battery run can be found on [the previous commit](https://travis-ci.org/github/metomi/rose/jobs/663541403?utm_medium=notification&utm_source=github_status)

etc/tutorial/cylc-forecasting-suite/bin/get-observations
MetRonnie: This fix was provided by Joe Hickson, he said
> I know setting speed / direction to a string of ‘0’ (string? Really? Eurgh) is complete lies and not how to handle an absent data point but I am assuming that it will suffice for the tutorial!

datamel:
> The tutorial docs have now been migrated to cylc/cylc-doc so you could also raise a PR there to get two for the price of 1 :)
> 
> @datamel Aren't the docs are unaffected?

Unless I am being really daft (which is possible and probable) the tutorials were moved quite recently from rose to cylc-docs. 
https://github.com/cylc/cylc-doc/blob/master/cylc/doc/etc/tutorial/cylc-forecasting-suite/bin/get-observations

https://github.com/cylc/cylc-doc/blob/master/cylc/doc/etc/tutorial/cylc-forecasting-suite/bin/get-rainfall

These are the same as the ones you have changed, no?

MetRonnie:
I see, you're right. I only tried searched for `urllib2` in `cylc-doc/` but nothing came up because it uses `urllib` and Python 3 instead

oliver-sanders:
`t/docs/02-tutorial-suites.t` passes in my environment

oliver-sanders:
For now let's not worry about re-raising this against the Cylc copy as that will likely change entirely for the Cylc8 release.

PR #2405, opened 2020-06-22
docs: document rosa command

oliver-sanders:


wxtim:
I'm afraid that I'm struggling to run this locally, although it looks reasonable enough - I imagine that it's some sort of environment issue.

oliver-sanders:
Ah yes, we finally have the long-awaited hieroglyph 2! Unfortunately the metadata may be wrong as it is trying to install Python3 code in a Python2 env, will take a look, in the mean time here's a patch.

bin/rose-make-docs
```diff
@@ -180,7 +180,7 @@ venv-install () {
     pip install 'sphinx'
     pip install 'sphinx_rtd_theme'
     pip install 'sphinxcontrib-httpdomain'
-    pip install 'hieroglyph'
+    pip install 'hieroglyph=1.0.0'
```
wxtim, line 183:
```suggestion
    pip install 'hieroglyph==1.0.0'

```

PR #2406, opened 2020-07-02
pinned hieroglyph to version 1.x as V2 doesn't have python2 support

wxtim:
Hieroglyph Version 2.0.0 removes Python 2 compatibility. This changes pins the Python 2 Rose branch to Version 1.


oliver-sanders:
Have you got the same commit here three times?

wxtim:
> Have you got the same commit here three times?
Same name - intention is to squash them...

PR #2407, opened 2020-07-31
DOCS: Fine tune rosie-create docstring

wxtim:
After  a conversation with Adam I attempted to clarify this documentation.


oliver-sanders:
Seems legit, note that these descriptions aren't actually written in markdown, but we explicitly parse back-ticks (and `NOTE` I think) but any other markdown syntax is ignored in the docs builder.

PR #2409, opened 2020-09-18
tutorial-bug: Clarify that suite won't run _yet_

wxtim:
I don't think it's clear that this suite won't run yet, and that we are only testing the installation and validation at this point.

### DIFF
--- a/sphinx/tutorial/rose/suites.rst
+++ b/sphinx/tutorial/rose/suites.rst
@@ -413,8 +413,6 @@ See the :ref:`Cheat Sheet` for more information.
 
    #. **Install The Suite.**
 
-      Running :ref:`command-rose-suite-run` will cause the suite to be
-      installed, validated and run.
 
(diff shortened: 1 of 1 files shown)


sphinx/tutorial/rose/suites.rst
```diff
@@ -416,8 +416,9 @@ See the :ref:`Cheat Sheet` for more information.
       Running :ref:`command-rose-suite-run` will cause the suite to be
       installed, validated and run.
```
oliver-sanders, line 417:
```suggestion
```

wxtim: I hadn't been going to be that brutal, but yes, that works!

oliver-sanders (approved):

oliver-sanders:
No need to copy these changes to cylc-doc as this section will be re-written.

oliver-sanders:
Going to ignore test failures as Rose2 needs a bit of tlc.

PR #2413, opened 2020-11-03
Fix setup.py for use with Cylc8

wxtim:
## Fix setup.py for use with Cylc8
I can't see why there is an upper limit to jinja2 - it was introduced when I created the package a year ago.

## Fix broken macro test
This test was broken on master.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,6 @@ with open("README.md", "r") as fh:
 
 INSTALL_REQUIRES = [
     "jinja2>=2.10.1",
-    "cylc-flow==8.*",
     "aiofiles",
     "tornado",
     "sqlalchemy",
--- a/setup.py
+++ b/setup.py
@@ -49,8 +49,8 @@ with open("README.md", "r") as fh:
 
 
 INSTALL_REQUIRES = [
-    "jinja2>=2.10.1, <2.11.0",
-    "cylc-flow==8.0a1",
+    "jinja2>=2.10.1",
+    "cylc-flow==8.*",
     "aiofiles",
     "tornado",
     "sqlalchemy",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -49,8 +49,8 @@ def find_version(*file_paths):
 
 
 INSTALL_REQUIRES = [
-    "jinja2>=2.10.1, <2.11.0",
-    "cylc-flow==8.0a1",
+    "jinja2>=2.10.1",
+    "cylc-flow==8.*",
```
oliver-sanders, line 53: Do we need the cylc-flow dependency?

wxtim: Probs not as discussed.

setup.py
```diff
@@ -49,8 +49,7 @@ def find_version(*file_paths):
 
 
 INSTALL_REQUIRES = [
-    "jinja2>=2.10.1, <2.11.0",
```
oliver-sanders, line 52: Not sure why this version range crept in, I think we should be good to up the requirement.

t/rose-macro/00-null.t
```diff
@@ -52,7 +52,7 @@ setup
 run_fail "$TEST_KEY" rose macro --unknown-option
 file_cmp "$TEST_KEY.out" "$TEST_KEY.out" </dev/null
 file_cmp "$TEST_KEY.err" "$TEST_KEY.err" <<'__CONTENT__'
-Usage: rose macro [OPTIONS] [MACRO_NAME ...]
+Usage: rose macro [options]
```
oliver-sanders, line 55: ?

wxtim: This is probably wrong. Ignore for now.

oliver-sanders (approved):

oliver-sanders:
Need to strip out rose suite-run and the related tests sometime.

wxtim:
> Need to strip out rose suite-run and the related tests sometime.

Let us get it running on Travis, move to Actions and then make changes when it does not take 15-20 minutes to run the tests. My branch for the "running on Travis bit is nearly ready"

PR #2414, opened 2020-11-04
Try to get the testing working again.

wxtim:
Replace the unit testing and linting tests from Travis with GH actions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,13 +18,6 @@ after_failure:
 
 jobs:
   include:
-    - name: "Unit Tests"
-      install:
-        - now install coverage linters pytest cylc rose
-      script:
-        - now test style
-        - now test units
-
     - name: "Test Battery"
       before_install:
         - export PATH="$PWD/.travis:$PATH"
--- a/.travis/now
+++ b/.travis/now
@@ -75,6 +75,7 @@ _gh_extract () {  # extract project from GitHub to $HOME
 }
 
 _install_coverage () {
+    APT+=(nodejs)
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -18,13 +18,6 @@ after_failure:
 
 jobs:
   include:
-    - name: "Unit Tests"
-      install:
-        - now install coverage linters pytest cylc rose
-      script:
-        - now test style
-        - now test units
```
oliver-sanders, line 26: Why not just copy the `now` commands over to GH actions to save re-implementing everything?

wxtim: Will that work?

.travis/now
```diff
@@ -75,6 +75,7 @@ _gh_extract () {  # extract project from GitHub to $HOME
 }
 
 _install_coverage () {
+    APT+=(nodejs)
```
oliver-sanders, line 78: Python-Coverage doesn't use nodejs.

.travis.yml
oliver-sanders: Yes that will work.

wxtim: I'll give it a go - probably will end up creating a new PR and closing this one. Ah well.

oliver-sanders: What's the motive for this PR?

oliver-sanders: Happy to move over to GH actions, however, it wont help the tests to pass.

wxtim: Tests take a long time to pass or fail. Perhaps your are right - i should get it working on Travis.

PR #2415, opened 2020-11-05
Fix for linting, unit and functional tests

wxtim:
#### Summary

A bunch of changes to try and get more of the tests working prior to moving tests to GH actions, since Travis seems painfully slow. Some of these are crude and the rose testing scheme needs a bit more of an overhaul, but that will be easier on Actions.

1. Remove references to `$TRAVIS_BUILD_PATH` which will break.
2. Remove `apt install heirloom-mailx` which is no longer available. I'm not clear which tests it's used for because it breaks nothing downstream.
3. Add a shellcheck exclusion for error [SC2001](https://github.com/koalaman/shellcheck/wiki/SC2001): I'm pretty sure this should not be a fatal exception anyway as there are cases where it should be over-ridden - I think this test is being over-strict.
4. Pinned eslint to version 6 to prevent it from complaining about syntax introduced since Node 8. This should not be allowed to stand long term, but I wanted a working version before porting to actions.
5. Fixed links which had become dead. I think that we should consider using wayback machine or similar when linking to sites that change periodically - this keeps breaking each time the Met Office re-organizes it's website, whilst the definition of nowcasting doesn't.

#### Expected test outcome
1. Unit tests: Pass
2. Test battery: Pass
3. Documentation: Fail


bin/rose-check-software
```diff
@@ -317,8 +317,8 @@ def tutorial(check=check):
 def docs(check=check):
     """Check software dependencies for the documentation builder."""
     ret = check_all('Documentation Builder', [
-        check('graphviz-dot <dot>', command_template='-V',
-               version_template='graphviz version ([^\s]+)', outfile=2),
+        # check('graphviz-dot <dot>', command_template='-V',
+        #        version_template='graphviz version ([^\s]+)', outfile=2),
```
oliver-sanders, line 321:
```suggestion
```

oliver-sanders:
`heirloom-mailx` is used to send mail to the user. I think the only things that use it are:

* `cylc` (not relevant)
* `rose suite-hook` - obsolete
* `svn_post_commit` - looks like we aren't testing email support?

So should be good to remove.

oliver-sanders (approved):

PR #2417, opened 2020-11-11
Address pre-/post-commit hook issues

MetRonnie:
Fixes several bugs.

Pre-commit hook:
- Check validity of rose-suite.info in branches, not just trunk, for each transaction (this includes whether the file is present, whether the owner, title, project are specified)
- Ensure sections are not permitted in the info file

Post-commit hook:
- If the info file fail is not encoded in UTF-8, try latin-1 aka ISO 8859-1 for decoding (this decoding is necessary to store the info files values in the database)
- If owner, project or title missing in a historical commit, insert the string `'null'` into the database instead of no entry at all

`db_create.py`:
- Instead of returning if the post-commit hook fails, skip the revision and continue

### DIFF
--- a/lib/python/rose/config.py
+++ b/lib/python/rose/config.py
@@ -1171,10 +1171,11 @@ class ConfigLoader(object):
         """Initialise the configuration utility.
 
         Arguments:
-        char_comment -- the character to indicate the start of a
-        comment.
-        char_assign -- the character to use to delimit a key=value
-        assignment.
+            char_assign (str): the character to use to delimit a key=value
+                assignment.
+            char_comment (str): the character to indicate the start of a
+                comment.
+            allow_sections (bool): whether to permit sections in the config.
 
         """
         self.char_assign = char_assign
--- a/lib/python/rose/config.py
+++ b/lib/python/rose/config.py
@@ -1166,7 +1166,8 @@ class ConfigLoader(object):
     TYPE_OPTION = "TYPE_OPTION"
     UNKNOWN_NAME = "<???>"
 
-    def __init__(self, char_assign=CHAR_ASSIGN, char_comment=CHAR_COMMENT):
+    def __init__(self, char_assign=CHAR_ASSIGN, char_comment=CHAR_COMMENT,
+                 allow_sections=True):
         """Initialise the configuration utility.
 
         Arguments:
@@ -1178,6 +1179,7 @@ class ConfigLoader(object):
         """
         self.char_assign = char_assign
         self.char_comment = char_comment
+        self.allow_sections = allow_sections
         self.re_option = re.compile(
             r"^(?P<state>!?!?)(?P<option>[^\s" +
             char_assign + r"]+)\s*" +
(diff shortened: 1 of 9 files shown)


lib/python/rosie/db_create.py
```diff
@@ -185,9 +201,21 @@ def load(self, repos_path):
...
                 sys.stdout.flush()
-            self.post_commit_hook.run(
-                repos_path, str(revision), no_notification=True)
-            event = RosieDatabaseLoadEvent(repos_path, revision, youngest)
+            try:
+                self.post_commit_hook.run(
+                    repos_path, str(revision), no_notification=True)
+            except (ConfigSyntaxError, DatabaseError, InfoFileError) as err:
+                if sys.stdout.isatty():
+                    sys.stdout.write("\r")
```
oliver-sanders, line 209: `\r`?

Is the intention to rewrite the current line?

lib/python/rosie/db_create.py
```diff
@@ -185,9 +201,21 @@ def load(self, repos_path):
...
+                    sys.stdout.flush()
+                err_msg = "Exception occurred: {0} - {1}".format(
+                    type(err).__name__, str(err))
+                message = ("Could not load revision {0} of {1} as the post-"
+                           "commit hook failed:\r{2}\r".format(
```
oliver-sanders, line 214: `\r`?

Did you mean `\n`?

`\r` is a carriage return without the resulting newline, it overwrites what comes before it:

```console
$ python2 -c 'print "foo\rbar\r"'
bar
```

oliver-sanders (commented):
I need to go through the larger diffs when I've got a bit more energy, looks good though.

> If owner, project or title missing in a historical commit, insert the string 'null' into the database instead of no entry at all

Sqlite3 supports NULL values which is what it was doing before? Why swap to `'null'`?

lib/python/rose/config.py
```diff
@@ -1533,10 +1540,12 @@ class ConfigSyntaxError(Exception):
 
     BAD_CHAR = "BAD_CHAR"
     BAD_SYNTAX = "BAD_SYNTAX"
+    BAD_SYNTAX_NO_SECTIONS = "BAD_SYNTAX_NO_SECTIONS"
 
     MESSAGES = {
         BAD_CHAR: """unexpected character or end of value""",
         BAD_SYNTAX: '''expecting "[SECTION]" or "KEY=VALUE"''',
+        BAD_SYNTAX_NO_SECTIONS: '''expecting "KEY=VALUE"'''
```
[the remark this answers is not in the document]
dpmatthews: I'm not sure an unexpected section is the only thing that could trigger this warning. The proposed warning is consistent with the BAD_SYNTAX warning.

lib/python/rosie/svn_hook.py
```diff
@@ -0,0 +1,155 @@
...
+                transaction.
+            transaction (str): A commit transaction identifer. Cannot be used
+                with revision.
+            allow_popen_err (bool): If True, return None if a RosePopenError
+                occurs during svnlook command.
+        """
+        if not branch:
+            branch = self.TRUNK
+        commit_opts = []
+        # TODO: warn or raise if both supplied?
```
dpmatthews, line 138: To do?

MetRonnie:
> Sqlite3 supports NULL values which is what it was doing before? Why swap to `'null'`?

Because the problem was `NULL` is not allowed. We reasoned this doesn't matter too much as it should only apply to historical commits that were missing an owner/title/project in `rose-suite.info`; the pre-commit hook should prevent this going forward. At least this way the historical commit can have an entry in the DB

lib/python/rosie/svn_pre_commit.py
```diff
@@ -207,39 +134,64 @@ def run(self, repos, txn):
...
+
+                    # Suite must have an owner
+                    txn_owner, txn_access_list = self._get_access_info(
+                        txn_info_map[sid])
+                    if not txn_owner:
+                        bad_changes.append(
+                            InfoFileError(InfoFileError.NO_OWNER))
+                        continue
+
+            # No need to check other non-trunk changes (?)
```
dpmatthews, line 192: Hopefully no need for the "(?)"

lib/python/rosie/db_create.py
MetRonnie: I think that's for when it's writing to the terminal, so it clears the previous loading message or something like that

MetRonnie:
All the points should be addressed by https://github.com/metomi/rose/pull/2417/commits/04e1b6a4400687d8effdbff9c9ba17b6e9c761a7.

On another note, should the syntax error message
> `expecting "[SECTION]" or "KEY=VALUE"`

mention continuation lines?
```ini
foo=bar
baz  # Syntax error
```
and
```ini
foo=bar
 baz  # Continuation line
```

lib/python/rose/config.py
```diff
@@ -1378,38 +1381,47 @@ def load(self, source, node=None, default_comments=None):
...
             # Match the start of an option setting?
             match = self.re_option.match(line)
             if not match:
-                raise ConfigSyntaxError(
-                    ConfigSyntaxError.BAD_SYNTAX, file_name, line_num, 0, line)
+                if self.allow_sections:
+                    err = ConfigSyntaxError.BAD_SYNTAX
+                else:
+                    err = ConfigSyntaxError.BAD_SYNTAX_NO_SECTIONS
+                raise ConfigSyntaxError(err, file_name, line_num, 0, line)
```
oliver-sanders, line 1424: Now that sections are handled separately this can go back the way it was?

```suggestion
                raise ConfigSyntaxError(
                      ConfigSyntaxError.BAD_SYNTAX, file_name, line_num, 0, line)
```

lib/python/rosie/svn_pre_commit.py
```diff
@@ -207,39 +134,64 @@ def run(self, repos, txn):
...
+                         path_trunk_info_file) not in changes):
                     bad_changes.append(
                         BadChange(status, path, BadChange.NO_INFO))
                 continue
 
-            # The rest are trunk changes in a suite
-            path_head = "/".join(names[0:self.LEN_ID]) + "/"
+            sid = "".join(names[0:self.LEN_ID])
+            branch = names[self.LEN_ID] if len(names) > self.LEN_ID else None
+            path_head = "/".join(sid) + "/"
```
oliver-sanders, line 164: Is this definitely equivalent to the original logic?

```python
# before
path_head = "/".join(names[0:self.LEN_ID]) + "/"
```

```python
# after
sid = "".join(names[0:self.LEN_ID])
path_head = "/".join(sid) + "/"
# expanded
path_head = "/".join("".join(names[0:self.LEN_ID])) + "/"
```

oliver-sanders (approved):
This looks ok to me, the logic seems to check out. I've flagged a couple of potential diferences, once they are reviewed should be good from the code perspective.

lib/python/rosie/svn_pre_commit.py
```diff
@@ -207,39 +134,64 @@ def run(self, repos, txn):
...
                 continue
 
-            # No need to check non-trunk changes
-            if len(names) > self.LEN_ID and names[self.LEN_ID] != "trunk":
-                continue
-
             # New suite should have an info file
-            if status[0] == self.ST_ADD and len(names) == self.LEN_ID:
-                if (self.ST_ADD, path + "trunk/") not in changes:
+            if len(names) == self.LEN_ID and status == self.ST_ADDED:
```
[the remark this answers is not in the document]
MetRonnie: According to http://svnbook.red-bean.com/en/1.7/svn.ref.svnlook.c.changed.html, for `svnlook changed`, if the change was an addition, the status will only ever be `"A"`

lib/python/rosie/svn_pre_commit.py
MetRonnie: Yes,
```python
"/".join("ab123")
```
is the same as
```python
"/".join(["a", "b", "1", "2", "3"])
```
if that's what you're asking

lib/python/rosie/svn_pre_commit.py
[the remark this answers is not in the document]
MetRonnie: And `status` is
```python
status, path = line.split(None, 1)
```
so it won't contain any whitespace after the `A`

lib/python/rosie/svn_hook.py
dpmatthews: Still need to remove the comment?

dpmatthews (approved):
I've tested this against the existing Rosie repository and it fixes all the issues identified.
Just the "TODO" comment in the code to fix before merging.

lib/python/rosie/svn_hook.py
MetRonnie: D'oh

PR #2418, opened 2020-11-12
Fix broken checkbox in rosie suites discovery

MetRonnie:
In the query page of Rosie Suites Discovery, the "all revisions" checkbox wasn't working


dpmatthews (approved):
Fixes the problem in my tests

PR #2421, opened 2020-11-27
Pre-commit hook: Add messages to certain bad changes

MetRonnie:
Split off the main hook-overhaul PR #2417. Gives more useful error messages

### DIFF
--- a/lib/python/rosie/svn_pre_commit.py
+++ b/lib/python/rosie/svn_pre_commit.py
@@ -131,18 +131,22 @@ class RosieSvnPreCommitHook(RosieSvnHook):
                     is_bad = True
                     break
             if is_bad:
-                bad_changes.append(BadChange(status, path))
+                msg = "Directories above the suites must match the ID patterns"
+                bad_changes.append(BadChange(status, path, content=msg))
                 continue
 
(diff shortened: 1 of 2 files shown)


lib/python/rosie/svn_pre_commit.py
```diff
@@ -260,7 +264,8 @@ def run(self, repos, txn):
 
             # Only admin users can remove the suite
             if author not in admin_users and not path_tail:
-                bad_changes.append(BadChange(status, path))
+                msg = "Only admin users can remove the suite"
```
dpmatthews, line 267: ```admin_users``` includes the suite owner who is the only person allowed to remove a suite under normal circumstances. Therefore, I think the message should be:
```"Only the suite owner can remove the suite"```

dpmatthews (changes requested):
Looks good apart from the one message

lib/python/rosie/svn_pre_commit.py
MetRonnie: Done

dpmatthews (approved):

wxtim (approved):

PR #2422, opened 2020-12-03
Fix db-create, pre-/post-commit hooks

MetRonnie:
#2417 ported to master

> Fixes several bugs.
>
> Pre-commit hook:
> - Check validity of rose-suite.info in branches, not just trunk, for each transaction (this includes whether the file is present, whether the owner, title, project are specified)
> - Ensure sections are not permitted in the info file
>
> Post-commit hook:
> - If the info file fail is not encoded in UTF-8, <del>try latin-1 aka ISO 8859-1 for decoding</del> <ins>raise a useful error saying configuration files must be in UTF-8 (this applies to pre-commit hook too)</ins> (this decoding is necessary to store the info files values in the database)
> - If owner, project or title missing in a historical commit, insert the string `'null'` into the database instead of no entry at all
>
> `db_create.py`:
> - Instead of returning if the post-commit hook fails, skip the revision and continue

### DIFF
--- a/metomi/rose/config.py
+++ b/metomi/rose/config.py
@@ -96,9 +96,16 @@ import re
 from metomi.rose.env import env_var_escape
 import shlex
 import sys
+import codecs
 from tempfile import NamedTemporaryFile, SpooledTemporaryFile
 from functools import cmp_to_key
 from metomi.rose.unicode_utils import write_safely
+
+
+codecs.register_error(
+    'handle_decode_err',
+    lambda err: (_ for _ in ()).throw(ConfigDecodeError(err))
+)  # The above is just a hack to raise an exception inside a lambda
 
 
 CHAR_ASSIGN = "="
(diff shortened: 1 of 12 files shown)


oliver-sanders (approved):

oliver-sanders:
Latest commits look good.

oliver-sanders:
@dpmatthews I think we can merge if you are happy.

dpmatthews:
Testing with the latest version showed up an unexpected difference in the output.
2019.01.x:
```
 [FAIL] BAD VALUE IN INFO FILE: <fdopen>(6): expecting "KEY=VALUE"
```
This branch:
```
> [FAIL] BAD VALUE IN INFO FILE: 5(6): expecting "KEY=VALUE"
```

dpmatthews (approved):
Everything now working fine in my tests

PR #2430, opened 2020-12-30
env_var_process: add opt to provide environ dict

oliver-sanders:
Allows usage like this:

```python
my_environ = {'foo': '42'}
env_var_process('$foo', my_environ)
```

### DIFF
--- a/metomi/rose/env.py
+++ b/metomi/rose/env.py
@@ -136,7 +136,7 @@ def env_var_escape(text, match_mode=None):
     return ret
 
 
-def env_var_process(text, unbound=None, match_mode=None):
+def env_var_process(text, unbound=None, match_mode=None, environ=os.environ):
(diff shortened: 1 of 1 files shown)


wxtim (approved):
- [x] Sanity Check.
- [x] Looked at failing tests - seems to be failing in the same way as master. Pretty nasty looking though. 😢 
- [x] Appears to allow tests on (cylc-rose:PR16)[https://github.com/cylc/cylc-rose/pull/16/files] to pass successfully.
- [ ] Utterly convinced that I have not missed small breakages in my comparison of test failures against master.

@oliver-sanders  - I'd like you to assure me that you think the current crop of test failures are reasonable before I merge.

oliver-sanders:
Currently working on an overhaul branch that will get the tests passing (down to the last 13 fails!).

PR #2431, opened 2020-12-30
Auto update copyright year

MetRonnie:
Sorry, this is getting repetitive!

The middle commit contains the 400-odd removals of years from copyright notices in the code

### DIFF
--- a/t/rose-config-diff/00-basic.t
+++ b/t/rose-config-diff/00-basic.t
@@ -1,6 +1,6 @@
 #!/bin/bash
 #-------------------------------------------------------------------------------
-# Copyright (C) 2012-2019 British Crown (Met Office) & Contributors.
+# Copyright (C) British Crown (Met Office) & Contributors.
 #
 # This file is part of Rose, a framework for meteorological suites.
 #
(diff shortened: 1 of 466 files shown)


t/rose-config-diff/00-basic.t
```diff
@@ -305,7 +305,7 @@ file_cmp "$TEST_KEY.out" "$TEST_KEY.out" <<'__DIFF__'
     13	# description=Environment variable configuration
     14	[env]
     15	GEARSTICK_DECORATION=golfball
-    16	
+    16
```
oliver-sanders, line 308: Will need to check these file comparisons still pass with whitespace stripped.

oliver-sanders: Nope, the whitespace is functional.

MetRonnie: D'oh

MetRonnie: I've gone through every file where more than 1 line was changed and undone the whitespace changes in heredocs

oliver-sanders: Thanks.

oliver-sanders (approved):

Issue #2438, opened 2021-01-29
Incremental file install not working if previous source re-used


dpmatthews:
If I run the following
```
echo 1 > test1
echo 2 > test2
echo 3 > test3
mkdir app-run
cd app-run
printf "[command]\ndefault=\n[file:test]\nsource=$OLDPWD/test1" > rose-app.conf 
rose app-run -q
echo  "Installed contents: $(cat test)"
printf "[command]\ndefault=\n[file:test]\nsource=$OLDPWD/test2" > rose-app.conf 
rose app-run -q
echo  "Installed contents: $(cat test)"
printf "[command]\ndefault=\n[file:test]\nsource=$OLDPWD/test3" > rose-app.conf 
rose app-run -q
echo  "Installed contents: $(cat test)"
printf "[command]\ndefault=\n[file:test]\nsource=$OLDPWD/test1" > rose-app.conf 
rose app-run -q
echo  "Installed contents: $(cat test)"
```
I get
```
Installed contents: 1
Installed contents: 2
Installed contents: 3
Installed contents: 3
```
Changing the source back to a previous version is just ignored.

wxtim:
I believe that this is not catered for in line 213 of `lib/python/rose/config_processors/fileinstall.py`.
```python
            source.is_out_of_date = (
                not prev_source or
                (not source.key and not source.paths) or
                prev_source.scheme != source.scheme or
                prev_source.loc_type != source.loc_type or
                prev_source.key != source.key or
                sorted(prev_source.paths) != sorted(source.paths))
```
I'm still investigating what `scheme`, `loc_type` and `key` are, but it doesn't look like a file with the same path as one in the DB is recognised as different. I presume that we'd like to store a checksum in the DB and check this?

wxtim:
I've found and fixed this bug in Rose 1.x. I will look at whether this bug occurs in `cylc install` once https://github.com/cylc/cylc-flow/pull/4023 has been merged. The test created in #2440  should probably be ported to `cylc/cylc-rose`.

oliver-sanders:
closed by #2440

PR #2440, opened 2021-02-03
[2019.x] File not installed if source used previously.

wxtim:
Address #2438 

- [x] New test created, which fails when put on 7.8.x

Response to bug reported by @jarich.


lib/python/rose/config_processors/fileinstall.py
```diff
@@ -250,6 +250,19 @@ def _process(self, conf_tree, nodes, loc_dao, **kwargs):
...
+                if not target.is_out_of_date:
+                    conn = loc_dao.get_conn()
+                    prev_dep_locs = conn.execute(
+                        "SELECT * FROM dep_names WHERE name=?", [target.name]
+                    ).fetchall()
+                    prev_dep_locs = [i[1] for i in prev_dep_locs]
+                    prev_dep_locs = [loc_dao.select(i) for i in prev_dep_locs]
+                    if (
+                        [i.name for i in prev_dep_locs] !=
+                        [i.name for i in target.dep_locs]
```
oliver-sanders, line 263: This ensures strict order is preserved, which may or may not be a good thing, will have to do some digging, my questions:

* Are `prev_dep_locs` always going to be stored in the same order as `target.dep_locs`?
  * I.E. Even if you don't change the `source` list could this equality resolve to `False` anyway.
* Does the order matter?

oliver-sanders: Answer, yes the order does matter:

```
[file:multiple]
source=a b c
```

Would result in a file that looks like this:

```
cat a
cat b
cat c
```

wxtim: I meant to sort, and forgot. Oops.

oliver-sanders: No oops at all, I think the order can matter.

oliver-sanders: After a bit of a play I'm happy that dep_names are stored in the right order too.

oliver-sanders (approved):

wxtim:
Request @oliver-sanders  review change to testing this change.
Request @dpmatthews checks functionality.

lib/python/rose/config_processors/fileinstall.py
wxtim:
```suggestion
                        sorted([i.name for i in prev_dep_locs]) !=
                        sorted([i.name for i in target.dep_locs])
```

lib/python/rose/config_processors/fileinstall.py
```diff
@@ -250,6 +250,19 @@ def _process(self, conf_tree, nodes, loc_dao, **kwargs):
                         if prev_path != path:
                             target.is_out_of_date = True
                             break
+                # See if any sources have changed names.
+                if not target.is_out_of_date:
+                    conn = loc_dao.get_conn()
```
oliver-sanders, line 255: I think we should close this in a try/finally block.

oliver-sanders (approved):

lib/python/rose/config_processors/fileinstall.py
oliver-sanders: Agh, no, my comments here were backing up your previous decision *not* to sort.

The order of sources can effect the resulting output so exact (unsorted) equality must be tested.

lib/python/rose/config_processors/fileinstall.py
```diff
@@ -250,6 +250,25 @@ def _process(self, conf_tree, nodes, loc_dao, **kwargs):
...
+                            "SELECT * FROM dep_names WHERE name=?",
+                            [target.name]
+                        ).fetchall()
+                        prev_dep_locs = [i[1] for i in prev_dep_locs]
+                        prev_dep_locs = [
+                            loc_dao.select(i) for i in prev_dep_locs
+                        ]
+                        if (
+                            sorted([i.name for i in prev_dep_locs]) !=
+                            sorted([i.name for i in target.dep_locs])
```
wxtim, line 267:
```suggestion
                            [i.name for i in prev_dep_locs] !=
                            [i.name for i in target.dep_locs]
```

kinow (approved):

kinow:
Not closing as `[2.0]` has a review pending (not sure if they need to go together, or if this one can be merged first). Thanks Tim!

Issue #2369, opened 2019-07-15
Remove GUI part of Rosie Graph and Rose Metadata Graph


wxtim:
Rosie Graph & Rose Metadata Graph have escaped the cull of GUI programs, but need to have their current GUI aspects removed. Text based testing however should be kept intact wherever possible to provide continuity.

Issue #2445, opened 2021-02-24
platform/host logic and Cylc integration


oliver-sanders:
Two bits of functionality depend on extracting hosts from Cylc workflows:

* `rose_prune` must get the host a task ran on.
* `fcm_make` must get the host a task will run on.

Broken interfaces awaiting re-implementation:

* `get_suite_jobs_auths`
* `get_task_auth`

Both these will be broken by #2446 

#### some more info on the fcm_make bit

The mirroring logic used by the `fcm_make` app relies on `SuiteEngineProc.get_task_auth` which has become broken by Cylc platforms.

https://github.com/metomi/rose/blob/4f6aaa5d2ffe6d327f3034a5ceef227b005cb4af/metomi/rose/suite_engine_procs/cylc.py#L207-L228

Note the `fcm_make` logic has a "back compat" notice in it:

https://github.com/metomi/rose/blob/4f6aaa5d2ffe6d327f3034a5ceef227b005cb4af/metomi/rose/apps/fcm_make.py#L197-L198

Investigate this logic, try and find a solution which allows us to remove `SuiteEngineProc.get_task_auth`.

dpmatthews:
The functionality is (mostly) described in the documentation: https://metomi.github.io/rose/doc/html/api/built-in/fcm_make.html
In particular look at the description of `mirror-step`.

The functionality is tested here: https://github.com/metomi/rose/tree/master/t/rose-task-run/08-app-fcm-make
`fcm_make_t5 => fcm_make2_t5`

Note that the `fcm_make.cfg` file sets `mirror.target=$ROSE_TASK_MIRROR_TARGET`.
https://github.com/metomi/rose/blob/master/t/rose-task-run/08-app-fcm-make/app/fcm_make_t5/file/fcm-make.cfg#L4
This is a deprecated setting which I don't think is required.
(We should remove this deprecated feature if possible)

We need to determine the remote host used by the `fcm_make2*` task in order to set the mirror target. In Cylc 8 I think that means:
1. Check to see if a remote host is set and use it if it is (note that this may be a command but I assume Rose already handles this).
2. Otherwise, check for a platform setting (again, this may be a command). Then lookup the hosts for that platform and choose one. I think selecting one at random should be good enough given that this feature is no longer widely used. For extra points we could check the host is accessible via ssh (which is what would happen with remote hosts specified via `rose host-select` at Cylc 7).

oliver-sanders:
[...]

Issue #2444, opened 2021-02-24
cli: convert from bash to python


oliver-sanders:
Currently the CLI is implemented as a collection of bash commands:

* `rose`
* `rose-config`
* `rose-macro`
* ...

The `rose` command exec's the `rose-<util>`.

This is problematic because:

* each command is unique (i.e. you get `rose-config` not `rose config`)
  * this makes command whitelisting hard
  * this is also just unnecessary and awkward
* pip will install each of the Rose commands individually
  * pip doesn't just install the one `rose` command, it installs all of them
* pip is bad at cleaning these bash commands up with upgrades or uninstalls
  * this means you can end up with deprecated and broken commands left behind
  * recommend you do something like `rm $(which rose)-*` to remove these before testing
* conda doesn't modify the shebang of these scripts as it does with Python scripts
  * this makes life harder when it comes to managing environments

Suggest:

* Re-implementing the root commands (`rose`, `rosie`) in Python.
* Using the same, import and call mechanism as Cylc to invoke these commands without calling them as `rose-<util>`.
* Support a legacy `os.execvp` implementation for any remaining bash commands (e.g. `rose-mpi-run`)
  * These will continue to be invoked as `rose-<util>` rather than `rose <util>` but they are few so damage is limited.
  * Example implementation https://github.com/cylc/cylc-flow/blob/accae0c9463e0ce91b22b7b750f9c14d26d297bb/cylc/flow/scripts/cylc.py#L50-L59

oliver-sanders:
> **Note:** From #2309, this should move Python scripts into the `metomi.rose` namespace.

oliver-sanders:
Closed by #2494

PR #2446, opened 2021-02-24
beta

oliver-sanders:
closes https://github.com/metomi/rose/issues/2369

Get Rose more-or-less beta ready.

It's code-dump time (sorry), wasn't easy to break this down into smaller chunks without a working test battery. At least the diff reduced to just 4000 lines 🤦.

Suggest chunking up the review to avoid unnecessary code-staring [suggested reviewers, feel free to edit]:

* [x] `.github/workflows/test.yml` [@MetRonnie]
* [x] [Only] The expected commands have been removed and "dead ends" inserted in their place. [@wxtim] [@datamel also checked this]
* [ ] [Only] The expected configurations have been removed (rose has no deprecation mechanism). [@wxtim]
* [ ] Rose & Cylc still play together nicely via `cylc install` and the `cylc-rose` plugin. [@wxtim]
* [ ] The host-select logic will need a thorough review and some heavy testing - 3155b673 [@datamel performed manual testing of host-select command]
* [x] The new `rose resource` command. [@datamel checked this]
* [x] The docs install and build nicely.[@datamel checked this]
* [x] The tests all pass locally (excluding remote host tests which are configured in `~/.metomi/rose.conf:[t]` [all]. [@datamel checked this]
* [x] The actions test failures are just `$PATH` related and not more serious - https://github.com/metomi/rose/pull/2446#issuecomment-800198318 [@datamel checked this (similar $PATH issues encountered as mentioned below)]
* [ ] `test_header` changes.

If there are larger batches of edits feel free to counter-PR, raise inline suggestions, whatever you feel like.

**Highlights:**

* Remove legacy Rose2019 stuff (the GUI, `rose suite-run`, etc).
* Upgrade the remaining workflow logic to Cylc8.
* Upgrade `rose host-select` to use an adapted form of the Cylc scheduler host select logic - 3155b673
* Removed the `rose-init` script logic.
* Moved the test battery under a registration hierarchy (similar to Cylc).
* GitHub Actions CI.
* Fix the tests (upgrade to Cylc8).
* Fix the docs (upgrade to Sphinx3).
* Flake8, eslint, shellcheck, etc.

**Fixes:**

* Issue causing rose_bunch to erroneously rerun succeeded sub jobs - 3155b673
* Bug causing fcm_make to fail with hierarchical workflow registrations - 9324db7c
* Re-instate the lost `rose version --long` option - b878e3ab
* Import issue with libsecret
* Fix non-editible `pip install`s - 87b14c1b
* popen.__str__ bugs - 132ec967
  * Infinite loop reading from stdin (not reproducible on MO Linux)
  * Bytes/string conversions
* Some minor portability fixes (mostly to the tests) for better *nix support (incl MacOS except `fcm` functionality)

**Caveats:**

* ~Small gaggle of `$PATH` related test failures on GH actions only - https://github.com/metomi/rose/pull/2446#issuecomment-800198318~ fixed by https://github.com/cylc/cylc-flow/pull/3440
* The following two commands must return the same result (when called from the appropriate environment):
   * `bash -c which python`
   * `bash -l which python`
   * https://github.com/metomi/rose/issues/2444
* I've disabled the MacOS build due to the issue above, this should pass once that has been sorted.
   * https://github.com/metomi/rose/issues/2447
* The mirroring logic in fcm_make has been jiggered by platforms work, tests skipped.
   * https://github.com/metomi/rose/issues/2445
* `rose stem` has been removed awaiting resurrection (possibly in another repo)
    * https://github.com/metomi/rose/issues/2448
* Docs build fixed but content not updated.
* PDF build removed from docs, can re-implement later if we actually want it.

**TODO:**

* [x] host-select mount https://github.com/metomi/rose/pull/2446#pullrequestreview-597506031

**Before Reviewing:**

Clean out any `rose` commands from previous installations:

```
$ which rose
~/miniconda3/envs/cylc-8.0a3.dev/bin/rose
$ rm ~/miniconda3/envs/cylc-8.0a3.dev/bin/rose-*
```

Then re-pip-install.


metomi/rose/host_select.py
```diff
@@ -484,59 +547,39 @@ class LoadScorer(RandomScorer):
...
-        return float(mem)
+        return out[0].available
 
 
 class FileSystemScorer(RandomScorer):
 
     """Score host by average file system percentage usage."""
 
-    ARG = "~"
+    ARG = "/"  # TODO?
```
oliver-sanders, line 575: Note to self; check whether this is testing the correct mount.

metomi/rose/popen.py
```diff
@@ -71,19 +70,23 @@ def __str__(self):
             ret = RosePopener.list_to_shell_str(self.command)
         if isinstance(self.stdin, str):
             ret += " <<'__STDIN__'\n" + self.stdin + "\n'__STDIN__'"
+        elif isinstance(self.stdin, bytes):
+            ret += " <<'__STDIN__'\n" + self.stdin.decode() + "\n'__STDIN__'"
         elif isinstance(self.stdin, io.IOBase):
-            try:
-                # FIXME: Is this safe?
```
oliver-sanders, line 76: No, not it's not!

oliver-sanders:
Ok, we are back to all local tests passing, woohoo.

However, a small gaggle of tests are now failing on GH Actions. These tests are ensuring that `$PATH` as seen by a Rose app is the same as `$PATH` as seen by the script calling the Rose app. After a bit of debug I can confirm that invoking bash non-interactively *without* the `-l` option results in a bash shell with a modified `$PATH` (which it shouldn't).

Evidence:

```bash
echo $PATH
bash -c 'echo $PATH'
bash -l -c 'echo $PATH'
```

Results in:

```
 /home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:/opt/pipx_bin:/usr/share/rust/.cargo/bin:/usr/local/.ghcup/bin:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/runner/.dotnet/tools:/home/runner/.config/composer/vendor/bin:/home/runner/.local/bin
/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:/opt/pipx_bin:/usr/share/rust/.cargo/bin:/usr/local/.ghcup/bin:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/runner/.dotnet/tools:/home/runner/.config/composer/vendor/bin:/home/runner/.local/bin:/home/runner/.dotnet/tools:/home/runner/.config/composer/vendor/bin:/home/runner/.local/bin
/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:/opt/pipx_bin:/usr/share/rust/.cargo/bin:/usr/local/.ghcup/bin:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/runner/.dotnet/tools:/home/runner/.config/composer/vendor/bin:/home/runner/.local/bin:/home/runner/.dotnet/tools:/home/runner/.dotnet/tools:/home/runner/.config/composer/vendor/bin:/home/runner/.local/bin:/home/runner/.dotnet/tools:/home/runner/.config/composer/vendor/bin:/home/runner/.local/bin
```

https://github.com/metomi/rose/pull/2446/checks?check_run_id=2121121227#step:3:1

Normally you would expect the first two to match and the third one to differ depending on profile files (e.g. `.bashrc`), however, the first two do *not* match which is breaking test logic. I think this is something GitHub has shimmed onto the shell in their runner. These tests passed on actions a while back so this must be a recent-ish change.

The difference between the first and second is some repeated stuff that has been re-appended onto the end of the path so not really a functional change, it's only a problem because it is tripping up test logic. It's for reasons like this that we should probably just run our tests in Docker - #2452.

oliver-sanders:
> **Rebase Lock**: Linear commits only from here on to give reviewers a chance, can rebase/rebase-squash pre-merge.

.github/workflows/test.yml
```diff
@@ -0,0 +1,227 @@
...
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Cylc
+        env:
+          cylc_repo: ${{ github.event.inputs.cylc_repo || 'cylc/cylc-flow' }}
+          cylc_branch: ${{ github.event.inputs.cylc_ref || 'master' }}
+        run: |
+          pip install "git+https://github.com/$cylc_repo@$cylc_branch"
```
MetRonnie, line 63: I think shellchecker would complain about this
```suggestion
          pip install "git+https://github.com/${cylc_repo}@${cylc_branch}"
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,227 @@
...
+
+      - name: Re-Run Fails
+        if: failure() && steps.functest.outcome == 'failure'
+        working-directory: rose
+        env:
+          OBJC_DISABLE_INITIALIZE_FORK_SAFETY: YES
+        run: |
+          # so we only re-run for debug purposes
+          cylc scan --state=all --color=never
+          etc/bin/rose-test-battery -j 1 -v --state=save,failed
```
MetRonnie, line 179: So is the idea that if the tests fail the 1st time, the whole actions run should fail even if the tests somehow pass on retry?

.github/workflows/test.yml
```diff
@@ -0,0 +1,227 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.rose_ref || github.sha }}
+
+      - name: Configure Python
+        uses: actions/setup-python@v1
```
MetRonnie, line 198:
```suggestion
        uses: actions/setup-python@v2
```

MetRonnie (changes requested):
Have reviewed the GH Actions changes only at this stage

.github/workflows/test.yml
oliver-sanders: Shellcheck is fine with this because `@` is not a valid char for environment variables so `$cylc_repo` cannot be mistaken for `$cylc_repo@`, that said I like braces.

.github/workflows/test.yml
oliver-sanders: Yes.

Tests should be deterministic, testing shouldn't be a polling exercise of running and re running tests until they pass.

Unfortunately with Cylc we just don't have the time at our disposal to re-write the test battery so have had to live with the idea of accepting tests that pass on the second run but thankfully the Rose test battery is much more stable so doesn't require this treatment.

datamel:
First spot, I'm starting with looking at the deprecated help commands:
- `rose suite-run help` (and commands like it) correctly are pointing users to the correct new command
- the general `rose help` command is still returning commands that are deprecated, without warning.
I'm happy to look into this further if you are snowed under @oliver-sanders.

oliver-sanders:
> the general rose help command is still returning commands that are deprecated

Aah, sorry all I forgot something:

```console
$ which rose
~/miniconda3/envs/cylc-8.0a3.dev/bin/rose
$ rm ~/miniconda3/envs/cylc-8.0a3.dev/bin/rose-*
```

You will need to manually remove the old rose scripts if you have previously `pip installed` Rose2 into your environment as pip will not remove them for you.

datamel:
> > the general rose help command is still returning commands that are deprecated
> 
> Aah, sorry all I forgot something:
> 
> ```
> $ which rose
> ~/miniconda3/envs/cylc-8.0a3.dev/bin/rose
> $ rm ~/miniconda3/envs/cylc-8.0a3.dev/bin/rose-*
> ```
> 
> You will need to manually remove the old rose scripts if you have previously `pip installed` Rose2 into your environment as pip will not remove them for you.

Thanks!

.github/workflows/test.yml
```diff
@@ -0,0 +1,227 @@
...
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 45
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest']
+        python-version: ['3.7']
```
wxtim, line 35: Not 3.8 & 3.9 yet?

oliver-sanders: Was just keeping the test matrix to a minimum whilst I went through the CI test failure loop.

metomi/rosie/suite_id.py
```diff
@@ -365,9 +364,31 @@ def _from_id_text(self, id_text):
...
+                    # (most likely a Cylc8 run directory)
+
+                    # TODO: extract version control information written by
+                    # Cylc install, see:
+                    # https://github.com/metomi/rose/issues/2432
+                    # https://github.com/cylc/cylc-flow/issues/3849
+                    raise SuiteIdLocationError(location)
+        except ValueError:
+            # Not an installed Cylc8 workflow run directory
+            pass
```
MetRonnie, line 388: I think `if Path('')` evaluates as `if True`, so better to remove the `if` completely to avoid confusion
```suggestion
        try:
            loc.relative_to(sdrr)
        except ValueError:
            # Not an installed Cylc8 workflow run directory
            pass
        else:
            if (loc / 'rose-suite.info').is_file():
                # This is an installed workflow with a rose-suite.info file
                # (most likely a Cylc8 run directory)

                # TODO: extract version control information written by
                # Cylc install, see:
                # https://github.com/metomi/rose/issues/2432
                # https://github.com/cylc/cylc-flow/issues/3849
                raise SuiteIdLocationError(location)
```

MetRonnie (commented):
Progress: Finished https://github.com/metomi/rose/pull/2446/commits/7757cf02df04bee1152b9b8737cef4f06eac1571

metomi/rose/apps/rose_bunch.py
```diff
@@ -554,3 +559,36 @@ def same_prev_config(self, current):
...
+    """
+    path = path.split(':')
+    counter = Counter(path)
+    for item, count in counter.items():
+        ptr = len(path) - 1
+        while count > 1:
+            if path[ptr] == item:
+                path.pop(ptr)
+                count -= 1
+            ptr -= 1
```
MetRonnie, line 593: Simpler might be
```suggestion
    path = list(dict.fromkeys(path))
```

> As of Python 3.7 (and CPython 3.6), standard dict is [guaranteed to preserve order](https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6)

https://stackoverflow.com/a/53657523/3217306

metomi/rosie/suite_id.py
oliver-sanders: Good spot.

Ah, yes, I was using it like `is_relative_to` (which we can't use because it was introduced in Python3.9).

MetRonnie (commented):
Progress: finished [a65141cf](https://github.com/metomi/rose/pull/2446/commits/a65141cf25a2c79ae2dc5a24848dc1c65cf9f318)

metomi/rose/apps/rose_bunch.py
oliver-sanders: Much nicer, can be a one-liner.

wxtim:
I think that ` t/rose-suite-run/24-host-log-timestamp/rose-suite.conf ` should have been deleted, but the diff says it's merely empty.

wxtim:
### Read:
- [x] Test battery section
- [ ] Other bits
### Tried:
- [ ] Running tests locally
- [ ] Playing with sample suites.

sphinx/tutorial/cylc/scheduling/integer-cycling.rst
```diff
@@ -1,612 +0,0 @@
-.. _tutorial-integer-cycling:
```
wxtim, line 1: Moved to Cylc, I presume?

oliver-sanders: Yes.

metomi/rose/tests/conftest.py
```diff
@@ -16,15 +13,27 @@
...
 
-import coverage
-coverage.process_startup()
+@pytest.fixture(scope='module')
+def mod_tmp_path():
+    """A tmp_path fixture with module-level scope."""
+    path = Path(TemporaryDirectory().name)
+    path.mkdir()
+    yield path
+    rmtree(path)
```
wxtim, line 39: possible alternative
```suggestion
@pytest.fixture(scope='module')
def mod_tmp_path(tmp_path_factory):
    path = tmp_path_factory.mktemp('foo')
    yield path
```
- Uses pytests inbuilt cleanup
- Rescoped by scope of inheriting fixture - I tried it with tests in different modules and I got test files `basetemp/foo1` and `basetemp/foo2`.

metomi/rose/suite_run.py
```diff
@@ -1,694 +0,0 @@
-# -*- coding: utf-8 -*-
```
wxtim, line 1: Satisfying.

bin/rose
```diff
@@ -176,6 +239,14 @@ doc)
     :;;
 esac
 
+CMD="${ROSE_NS}-${UTIL}"
```
wxtim, line 242: This works fine for me if I do `rose suite-run`, but if I do `rose`/`rose --help` the help message contains the old info: i.e.
```
    suite-run
        Install and run a Cylc suite.
```

oliver-sanders: I think you have probably got some old Rose files kicking about in your environment, steps to remove them here: https://github.com/metomi/rose/pull/2446#issuecomment-800342335

bin/rose-tutorial
```diff
@@ -30,8 +30,7 @@
 #     Make a copy of one of the tutorial SUITE in the cylc-run directory.
 #
 #-------------------------------------------------------------------------------
-# shellcheck source=lib/bash/rose_init
-. rose_init
+set -eu
```
wxtim, line 33: I think that this may be broken - 
```bash
$ rose tutorial
rose tutorial SUITE [DIR]

Available tutorial suites:
find: './etc/tutorial/': No such file or directory
```

oliver-sanders: Currently works for me in an editable installation. probably more by chance than anything though.

I would like to punt this to an issue as the tutorials themselves are currently broken anyway, this would be better tackled as part of documentation work (will require a lot of docs updates etc):

https://github.com/metomi/rose/issues/2454

wxtim: I fixed it with
```diff
--- a/bin/rose-tutorial
+++ b/bin/rose-tutorial
@@ -37,7 +37,7 @@ usage () {
     echo 'rose tutorial SUITE [DIR]'
     echo
     echo 'Available tutorial suites:'
-    find "./etc/tutorial/" -maxdepth 1 -exec basename {} \; | sed 's/^/    /'
+    find "${ROSE_HOME}/etc/tutorial/" -maxdepth 1 -exec basename {} \; | sed 's/^/    /'
 }
 
 api-key () {
@@ -45,6 +45,8 @@ api-key () {
     sort -R "${ROSE_HOME}/etc/tutorial/api-keys" | head -n 1
 }
 
+ROSE_HOME=$(python3 -c "from pathlib import Path; import metomi.rose; print(Path(metomi.rose.__file__).parent.parent.parent)")
+
 if [[ $# == 1 ]]; then
     DIRECTORY="$HOME/cylc-run/$1"
 elif [[ $# == 2 ]]; then
@@ -87,7 +89,7 @@ fi
 echo -n "Copying tutorial files to ${DIRECTORY} ... "
 
 mkdir -p "$(dirname "${DIRECTORY}")"
-if run rsync -rLptgoD "${ROSE_HOME}/etc/tutorial/$1/" "${DIRECTORY}" \
+if rsync -rLptgoD "${ROSE_HOME}/etc/tutorial/$1/" "${DIRECTORY}" \
     --exclude='.validate'; then
     find "${DIRECTORY}" -type f \
         -exec sed -i "s/DATAPOINT_API_KEY/$(api-key)/" {} \;
```

wxtim (changes requested):
Generally happy, but I have 3 tests I can't pass locally. 

### Read:
- [x] Test battery files
- [x] Sphinx subfolder files
- [x] Top Level Files
- [x] `metomi/rosie` files
- [x] `metomi/rose/tests` files
- [x] `metomi/rose/` files
### Tried:
- [x] Running Pytest tests locally. All good.
- [ ] Running tests locally (3 Fail, looking to see if I can see obvs fix)
    - [x] FAILED: `t/rose-task-run/07-app-arch.t`.
    - [x] FAILED: `t/rose-task-run/24-app-fcm-make-fast.t`
    - [x] FAILED: `t/rose-task-run/31-app-bunch.t`
- [x] Playing with sample suites.
- [x] Playing with Tutorial - broken - happy to leave this until later, on your say-so. The scope of this PR is already somewhat large.
- [x] `make clean html linkcheck` 
### Checked for:
- [x] `rose` where `metomi.rose` required.
- [x] `suite.rc`.

bin/rose
wxtim: No, that doesn't seem to fix it. It works for you?

oliver-sanders: It is definitely not coming from this branch because:

```console
$ git grep 'Install and run a Cylc suite'
```

Could your wrapper script be picking up Rose2019?

wxtim: OK now - I copied your rather more sophisticated wrapper script.

oliver-sanders:
I can't replicate those failures, here's my test setup:

cylc-flow@master 793d251cf85ad090bf964155aa96048972a79a30
cylc-rose@master dc7e4f6701cc727d4e663f36de3f07c17a3a0dbb
metomi-rose@oliver-sanders/beta 2dd8ee3595f368c5c3286e136a09d77fb773a495

There has just been a change to cylc-flow which will require a `pip install` which could be tripping things up.

wxtim:
> I can't replicate those failures, here's my test setup:

My bad, my updated wrapper script looks like it's sorted it out.

wxtim (approved):

metomi/rose/macro.py
```diff
@@ -766,7 +765,7 @@ def get_macro_class_methods(macro_modules):
         macro_name = macro_module.__name__
         contents = inspect.getmembers(macro_module)
         for obj_name, obj in contents:
-            if not inspect.isclass(obj):
+            if not inspect.isclass(object):
```
MetRonnie, line 768: Possible mistake?

MetRonnie (commented):
Progress: finished [91e647f](https://github.com/metomi/rose/pull/2446/commits/91e647f5571ec970bb6178a582230dde01b35fc2)

metomi/rose/host_select_client.py
```diff
@@ -17,19 +15,38 @@
...
+    # read metrics from stdin
+    line = True
+    metrics = ''
+    while True:
+        line = sys.stdin.readline().strip()
+        if '**start**' in line:
+            continue
+        if '**end**' in line:
+            break
+        metrics += f'\n{line}'
```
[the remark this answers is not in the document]
oliver-sanders: Made it so any junk before the start marker is ignored, added a test.

On the infinite loop thing, yes you could end up with an infinite loop here, however, we can't fix this because we are reading from stdin.

`while line` would just break on the first blank line and would still get into the infinite loop because `sys.stdin.readline` will hang whilst waiting for something to read.

metomi/rose/macro.py
oliver-sanders: Definite mistake, great spot.

t/rosa-svn-pre-commit/04-unicode.t
```diff
@@ -47,16 +47,13 @@ chmod +x repos/foo/hooks/pre-commit
 export LANG=C
 
 TEST_KEY="${TEST_KEY_BASE}-western"
-cat > rose-suite.info << '__INFO__'
-owner=ivy
-project=euro
-title=We should not éñçödê config files in latin-1/western
-__INFO__
-iconv -f UTF-8 -t LATIN1 rose-suite.info -o rose-suite.info
-# -----------------------------------------------------------------------------
```
MetRonnie, line 56: Has this been moved to a file due to portability issues of `iconv`?

metomi/rose/host_select_client.py
```diff
@@ -1,6 +1,4 @@
-# -*- coding: utf-8 -*-
-# -----------------------------------------------------------------------------
-# Copyright (C) British Crown (Met Office) & Contributors.
+# Copyright (C) 2012-2019 British Crown (Met Office) & Contributors.
```
MetRonnie, line 1:
```suggestion
# Copyright (C) British Crown (Met Office) & Contributors.
```

metomi/rose/tests/test_host_select_client.py
```diff
@@ -0,0 +1,80 @@
+# Copyright (C) 2012-2019 British Crown (Met Office) & Contributors.
```
MetRonnie, line 1:
```suggestion
# Copyright (C) British Crown (Met Office) & Contributors.
```

bin/rose
```diff
@@ -37,10 +38,55 @@
...
-. rose_init
-rose_init
+set -eu
+
+ROSE_NS=$(basename "$0")
+ROSE_HOME_BIN="$(dirname "$0")"
+ROSE_VERSION="$(python3 -c "import metomi.rose; print(metomi.rose.__version__)")"
+export ROSE_NS ROSE_HOME_BIN ROSE_VERSION
+
+# NOTE: cannot use assiciative array due to bash version requirement
```
datamel, line 48:
```suggestion
# NOTE: cannot use associative array due to bash version requirement
```

MetRonnie (commented):
Progress: finished [dcbb73e](https://github.com/metomi/rose/pull/2446/commits/dcbb73e8a3222335acac4a1e1f1b051b744c98ef)

sphinx/Makefile
```diff
@@ -1,186 +1,25 @@
-# Makefile for Sphinx documentation
-#
+# Minimal makefile for Sphinx documentation
```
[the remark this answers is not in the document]
oliver-sanders: Yeah that was purposeful (you don't execute the `Makefile` directly you run `make` which loads the `Makefile` for you).

t/rosa-svn-pre-commit/04-unicode.t
oliver-sanders: Can't remember TBH, would only have changed if the test failed.

metomi/rose/config_processors/fileinstall.py
```diff
@@ -260,23 +260,20 @@ def _process(self, conf_tree, nodes, loc_dao, **kwargs):
...
-                            loc_dao.select(i)
-                            for i in prev_dep_locs
-                        ]
-                        if (
-                            [i.name for i in prev_dep_locs] !=
-                            [i.name for i in target.dep_locs]
-                        ):
-                            target.is_out_of_date = True
-                    finally:
-                        conn.close()
```
MetRonnie, line 271: When does it get closed, then?

MetRonnie (commented):
Progress: finished! (but have yet to test anything)

sphinx/tutorial/rose/furthertopics/rose-stem-tutorial.rst
```diff
@@ -129,12 +129,12 @@ The ``$SOURCE_SPACESHIP`` environment variable will be set using the
 Jinja2 variable of the same name which is provided by Rose Stem.
```
MetRonnie, line 131: Is the rose-stem tutorial being deliberately kept?

etc/conf/macos-patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/cylc/flow/hostuserutil.py b/cylc/flow/hostuserutil.py
+index 21e51735e..17917b8fc 100644
+--- a/cylc/flow/hostuserutil.py
++++ b/cylc/flow/hostuserutil.py
+@@ -113,7 +113,7 @@ class HostUtil:
+         """Return the extended info of the current host."""
+         if target not in self._host_exs:
+             if target is None:
+-                target = socket.getfqdn()
++                target = socket.gethostname()
```
datamel, line 10:
```suggestion
                 target = socket.getfqdn()
                 target = socket.gethostname()
```

etc/tutorial/rose-stem/.validate
```diff
@@ -1,4 +1,4 @@
 mkdir "${CYLC_RUN_DIR}/${REG}"
-echo -e '#!Jinja2\n{% set RUN_NAMES=["command_spaceship"] %}' > "${CYLC_RUN_DIR}/${REG}/suite.rc"
-cat "$TUT_DIR/rose-stem/suite.rc" >> "${CYLC_RUN_DIR}/${REG}/suite.rc"
+echo -e '#!Jinja2\n{% set RUN_NAMES=["command_spaceship"] %}' > "${CYLC_RUN_DIR}/${REG}/flow.cylc"
+cat "$TUT_DIR/rose-stem/flow.cylc" >> "${CYLC_RUN_DIR}/${REG}/suite.rc"
```
datamel, line 3: Should that be...
```suggestion
cat "$TUT_DIR/rose-stem/flow.cylc" >> "${CYLC_RUN_DIR}/${REG}/flow.cylc"
```

metomi/rose/apps/fcm_make.py
```diff
@@ -173,6 +176,7 @@ def _run_orig(self, app_runner, conf_tree, opts, args, uuid, task,
         # Determine the name of the continuation task
         task_name_cont = task.task_name.replace(
             orig_cont_map[ORIG], orig_cont_map[CONT])
+        # TODO: get_task_auth currently does nothing
```
datamel, line 179: Is this TODO important before beta? Just wondering if it was intended to be in this pr?

datamel (commented):
I'm 250 files into my review. In terms of running rose: I have run the helps and checked the dead ends. 
Will come back to this review in a little while.

metomi/rose/config_processors/fileinstall.py
oliver-sanders: This got put here because of a review comment by me on another branch but it turned out to be completely wrong. I don't know where or if the connection is closed (other than by an implicit try/finally in sys.exit or something like that) would have to go digging.

sphinx/tutorial/rose/furthertopics/rose-stem-tutorial.rst
oliver-sanders: Yep, `rose stem` isn't dead, it's just awaiting its new life in cylc-rose - https://github.com/cylc/cylc-rose/pull/45

I guess makes sense if the docs remain here since it's a rose cmd.

etc/conf/macos-patch
oliver-sanders: It's a patch file so these +/- symbols tell the `patch` command which lines to add/remove.

metomi/rose/apps/fcm_make.py
oliver-sanders: Kinda but we can limp on without it for a short time, here's the issue - https://github.com/metomi/rose/issues/2445

etc/conf/macos-patch
datamel: doh, my brain!

metomi/rose/suite_engine_procs/cylc.py
```diff
@@ -180,81 +67,19 @@ def get_suite_dir_rel(cls, suite_name, *paths):
...
-        for row in self._db_exec(suite_name, stmt, stmt_args):
-            if row and row[0]:
-                auth = self._parse_user_host(auth=row[0])
-                if auth:
-                    auths.append(auth)
-        self._db_close(suite_name)
-        return auths
+        # TODO: reimplement for Cylc8?
+        # https://github.com/metomi/rose/issues/2445
+        return []
```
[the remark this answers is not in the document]
oliver-sanders: Yeah, why not, I'll add something. I didn't because in the event that this causes problems the log message would end up in the job.err of a different task to the one that actually fails but a log message in the wrong place is better than no log message.

MetRonnie:
- [x] All tests pass locally

datamel:
After setting up Rose on remote platform, and with a similar `rose.conf` to the one you posted Oliver, I have quite a few functional tests are failing locally for me.
I have re-run these and they look consistent fails. I will look a little closer at this tomorrow.
image

datamel:
> After setting up Rose on remote platform, and with a similar `rose.conf` to the one you posted Oliver, I have quite a few functional tests are failing locally for me.
> I have re-run these and they look consistent fails. I will look a little closer at this tomorrow.

Update: the failures were related to the wrapper script. Disabling this results in test passes.
Whilst the wrapper script is needed, the failures are only related to tests, not functionality/user experience.

datamel (approved):
I am comfortable with the changes made.
Thanks @oliver-sanders , mammoth change *clap, clap*, and thanks for the help with the testing!

MetRonnie:
Exactly 21,000 lines deleted 😮

oliver-sanders:
Added a changelog (totally not to make it exactly 4,500 lined added)

MetRonnie:
I looked into the latest test failure as it seemed related to my recent PR in cylc-flow. You need this patch:
```diff
diff --git a/t/rose-task-run/38-app-bunch-counts.t b/t/rose-task-run/38-app-bunch-counts.t
index 67452522..4e7fd6b6 100755
--- a/t/rose-task-run/38-app-bunch-counts.t
+++ b/t/rose-task-run/38-app-bunch-counts.t
@@ -33,7 +33,7 @@ run_pass "$TEST_KEY" \
     cylc install \
         -C "$TEST_SOURCE_DIR/$TEST_KEY_BASE" \
         --flow-name="$FLOW" \
-        --no-run-name \
+        --no-run-name
 TEST_KEY="${TEST_KEY_BASE}-play"
 run_pass "$TEST_KEY" \
     cylc play \
```

oliver-sanders:
3b834a8 closes the outstanding TODO.

cylc/cylc-flow#3440 fixes caveat point (1).

sphinx/hyperlinks.rst
```diff
@@ -13,7 +13,8 @@
 .. _Cylc Flow: https://github.com/cylc/cylc-flow
 .. _Cylc Rose: https://github.com/cylc/cylc-rose
 .. _Cylc User Guide: https://cylc.github.io/documentation/#the-cylc-user-guide
-.. _Cylc Suite Design Guide: https://cylc.github.io/cylc-doc/current/html/suite-design-guide/suite-design-guide-master.html
+.. _Cylc Suite Design Guide: https://cylc.github.io/cylc-doc/stable/html/suite-design-guide/suite-design-guide-master.html
```
MetRonnie, line 16: At the moment, stable points to 7.x, should this be latest instead for 8.x? Or is the idea to just leave this alone so when Cylc 8 is out proper, it doesn't have to be updated?

oliver-sanders: I couldn't decide which to use so went with stable because at least then we can't forget to update it later.

MetRonnie (approved):

sphinx/hyperlinks.rst
oliver-sanders: Created an issue - https://github.com/metomi/rose/issues/2457

PR #2449, opened 2021-02-25
Revert "templatevars: forewarn of cylc8 changes"

oliver-sanders:
Reverts metomi/rose#2425


oliver-sanders:
Sadly it turned out that the scale of incompatibility in existing workflows is too great to make this change so it was necessary to provide support for Jinja2 literal definitions in cylc-rose.

The new parser will reject Jinja2 expressions (e.g. `1 + 1`), however, these are few and far between so we do not need to invest much energy into pre-warning about these usages. It would be difficult to back-port the new parser due to changes in Python and Jinja2 version.

wxtim (approved):
😢

PR #2453, opened 2021-03-15
fileinstall: fix

oliver-sanders:
This test is failing on 2019.01.x:
```
t/rose-app-run/11-file-incr-2.t
```

### DIFF
--- a/lib/python/rose/config_processors/fileinstall.py
+++ b/lib/python/rose/config_processors/fileinstall.py
@@ -253,22 +253,24 @@ class ConfigProcessorForFile(ConfigProcessorBase):
                 # See if any sources have changed names.
                 if not target.is_out_of_date:
                     conn = loc_dao.get_conn()
-                    try:
-                        prev_dep_locs = conn.execute(
-                            "SELECT * FROM dep_names WHERE name=?",
-                            [target.name]
-                        ).fetchall()
-                        prev_dep_locs = [i[1] for i in prev_dep_locs]
-                        prev_dep_locs = [
-                            loc_dao.select(i) for i in prev_dep_locs
-                        ]
-                        if (
-                            [i.name for i in prev_dep_locs] !=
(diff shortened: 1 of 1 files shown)


lib/python/rose/config_processors/fileinstall.py
```diff
@@ -253,22 +253,24 @@ def _process(self, conf_tree, nodes, loc_dao, **kwargs):
...
-                        ):
-                            target.is_out_of_date = True
-                    finally:
-                        conn.close()
+                    prev_dep_locs = conn.execute(
+                        """
+                            SELECT *
+                            FROM dep_names
+                            WHERE name=?
+                            ORDER BY ROWID
```
oliver-sanders, line 261: This is needed to get the rows to come out in the right order.

lib/python/rose/config_processors/fileinstall.py
```diff
@@ -253,22 +253,24 @@ def _process(self, conf_tree, nodes, loc_dao, **kwargs):
...
-                        prev_dep_locs = [
-                            loc_dao.select(i) for i in prev_dep_locs
-                        ]
-                        if (
-                            [i.name for i in prev_dep_locs] !=
-                            [i.name for i in target.dep_locs]
-                        ):
-                            target.is_out_of_date = True
-                    finally:
-                        conn.close()
```
oliver-sanders, line 271: This is my bad, it turns out that this method can be called multiple times on the same DB (which I should have expected) so closing it here will cause subsequent errors. TBH I'm not sure if/how the DB is closed.

oliver-sanders:
The `t/rose-app-run/` tests hit this logic pretty hard and are a good benchmark of functionality.

wxtim (approved):
- [x] Read and sanity checked.
- [x] Run tests locally.
- [x] Checked that test I recently wrote is still passing.

PR #2461, opened 2021-03-31
Fix: FCM Make and Rose Prune

wxtim:
Response to #2445: Required [cylc-rose PR 53](https://github.com/cylc/cylc-rose/pull/53) as pre-requisite.
Reimpliment methods using platforms allowing fcm_make and rose_prune to work

- [x] `get_task_auth` reimplimented.
  - [x] test_skips with #2445 references removed.
- [x] `get_suite_jobs_auths` reimplimented.
  - [x] Test skips with #2445 references removed.
- [x] Re-instated methods unit tested

PR #2470, opened 2021-05-14
mypy: configure rose to use mypy

oliver-sanders:
built on: https://github.com/metomi/rose/pull/2461

Done whilst validating the type hints added in #2461

Minimal changes required to allow mypy to pass, allowing future development of type checking.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,6 +153,11 @@ jobs:
           etc/bin/shellchecker
           yarn run lint
 
+      - name: MyPy
+        working-directory: rose
+        run: |
+          mypy --explicit-package-bases
+
       - name: Unit Tests
         working-directory: rose
         run: |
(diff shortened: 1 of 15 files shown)


wxtim:
Were you waiting for me to comment, or is this still WIP @oliver-sanders ?
P.S. I'm struggling to run mypy locally - it fails with errors - I don't think that --exclude and it's ini file equiv are included at version 8.00. I tried cloning into master and installing that. I think that we should be prepped to fix breakages for the next few mypy updates, or to pin the mypy version for our projects more firmly until mypy settles a little.

oliver-sanders:
./t/rosie-lookup/00-basic.t  failed in the 3.7 test (but not in the others) it is slightly flaky and takes much longer than the other tests, should look into it one day.

wxtim (approved):
Now passes locally. Re-running the tests. Will merge once done.

PR #2472, opened 2021-07-12
mypy: fix config

oliver-sanders:
Copy cylc config into rose.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,5 +7,8 @@ exclude = metomi/rose/etc/
 # Enable PEP 420 style namespace packages, which we use.
 # Needed for associating "import foo.bar" with foo/bar.py
 namespace_packages = True
+# Need this because we don't have __init__.py in top-level dir:
+explicit_package_bases = True
 
 strict_equality = True
+show_error_codes = True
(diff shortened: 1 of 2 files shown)


MetRonnie (changes requested):
It needs `types-aiofiles` in the `setup.py` dependencies or `mypy --install-types` twice in the GH Actions workflow. We went for the former in cylc-flow

MetRonnie (approved):
Leaving it to you to merge in case you want to add `allow_redefinition = True` as I've proposed in cylc-flow: https://github.com/cylc/cylc-flow/pull/4304

mypy.ini
```diff
@@ -7,5 +7,8 @@ exclude = metomi/rose/etc/
 # Enable PEP 420 style namespace packages, which we use.
 # Needed for associating "import foo.bar" with foo/bar.py
 namespace_packages = True
+# Need this because we don't have __init__.py in top-level dir:
+explicit_package_bases = True
```
oliver-sanders, line 11:
```suggestion
explicit_package_bases = True
allow_redefinition = True
```

oliver-sanders:
Added

MetRonnie (approved):

wxtim (approved):

PR #2473, opened 2021-07-15
get tutorial workflow up and running again

wxtim:
Fixes Tutorial Workflow

#### Problem
- The wind instruments at Heathrow seem to be flaky.
- Datapoint is deprecated

#### Done
- [x] Removed dependency on Datapoint (I'm not sure that the replacement is automatically better, by Ogimet seems fairly reliable).
- [x] Added a (very crude) routine for guessing the next nearest site for getting wind data.

#### Not Done in this PR
- [ ] Removed the datapoint dependency for the rainfall data.

To test

```bash
rose tutorial rose-weather-forecasting-suite ~/rose-tut
cd ~/rose-tut
sed -i '@# TODO - set the initial cycle point.@initial cycle point = DATETIME@'  suite.rc
rose suite-run
```


etc/tutorial/cylc-forecasting-suite/bin/get-observations
```diff
@@ -1,10 +1,24 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
 # -----------------------------------------------------------------------------
+# This is illustrative code developed for tutorial purposes, it is not
+# intended for scientific use and is not guarantied to be accurate or correct.s
```
kinow, line 21: s/guarantied/guaranteed, and an extra `s` after period.

kinow (approved):
One typo, but otherwise looks good! Thanks for fixing it @wxtim !

etc/tutorial/cylc-forecasting-suite/bin/get-observations
wxtim:
```suggestion
# intended for scientific use and is not guaranteed to be accurate or correct.
```

etc/tutorial/cylc-forecasting-suite/bin/get-observations
```diff
@@ -1,10 +1,24 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
MetRonnie, line 18: Doesn't the Rose licence differ from Cylc?

wxtim: Yep, althoug I think that this is superficial. Will fix.

wxtim: I think I've fixed this, although I notice that a number of other copyright notices have leaked from Cylc into Rose they are not related to this change, and I'd prefer to address them separately.

PR #2506, opened 2021-11-26
app-run: undo path changes made by the wrapper script

oliver-sanders:
See https://github.com/cylc/cylc-flow/pull/4425

Will need to be niceified.


dpmatthews:
I think we only need to set PATH in the wrapper when running `cylc hub`.
If so we won't need this change.

oliver-sanders:
No longer needed :+1:

PR #2508, opened 2021-11-29
Fix problems with rose stem

wxtim:
- Remove duplicate `.. _command-rose-stem`.
- Allow Rose Stem's CLI Help parsing to use Options from Cylc Option parser.

### DIFF
--- a/sphinx/api/command-reference.rst
+++ b/sphinx/api/command-reference.rst
@@ -41,8 +41,6 @@ This command has been replaced by ``cylc play``.
 
 .. TODO: This is here to allow the documentation tests to pass
 
-.. _command-rose-stem:
-
 rose stem
 ^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


sphinx/api/command-reference.rst
```diff
@@ -41,8 +41,6 @@ This command has been replaced by ``cylc play``.
 
 .. TODO: This is here to allow the documentation tests to pass
 
-.. _command-rose-stem:
-
 rose stem
```
oliver-sanders, line 44: Need to remove this too?

wxtim: Yes.

MetRonnie (commented):
How can I test this? (If a 2nd reviewer other than Oliver is needed)

wxtim:
> How can I test this? (If a 2nd reviewer other than Oliver is needed)

```
cd ~/rose/sphinx
make html
```

[It only needs one reviewer, but if you can do it sooner it would be nice to have it in]

sphinx/api/command-reference.rst
oliver-sanders: Still need to remove this.

wxtim:
Test broken due to SRS outage breaking linkcheck.

oliver-sanders (approved):

PR #2509, opened 2021-11-29
Popen always decodes

### LINKED ISSUE #2320: Rose popener should return consistent types
Had I been starting my Python 3 conversion from scratch I'd have added some code to rose.popener.run thus:
```python
        if isinstance(stderr, bytes):
            stderr = stderr.decode()
        if isinstance(stdout, bytes):
            stdout = stdout.decode()
```
This however breaks decode/encode statements used in multiple other places.

wxtim:
Closes #2320 

Make the rose Popener run object always return stdout, stderr and retcode as str and not bytes. This allows removed of a number of checks elsewhere.

This is an oversight from my original conversion of Rose to Python 3.

### DIFF
--- a/metomi/rose/popen.py
+++ b/metomi/rose/popen.py
@@ -257,10 +257,6 @@ class RosePopener:
         """
         ret_code, stdout, stderr = self.run(*args, **kwargs)
         if ret_code:
-            if stderr:
-                stderr = stderr
-            else:
-                stderr = ''
             raise RosePopenError(
                 args, ret_code, stdout, stderr, kwargs.get("stdin")
             )
--- a/metomi/rose/popen.py
+++ b/metomi/rose/popen.py
@@ -189,16 +189,7 @@ class RosePopener:
         stdout, stderr = proc.communicate(stdin)
         retcode = proc.wait()
 
-        def _decode_if_bytes(obj):
-            if isinstance(obj, bytes):
-                return obj.decode()
-            return obj
-
-        return (
-            _decode_if_bytes(retcode),
-            _decode_if_bytes(stdout),
-            _decode_if_bytes(stderr)
-        )
+        return retcode, stdout, stderr
 
     def run_bg(self, *args, **kwargs):
         """Provide a Rose-friendly interface to subprocess.Popen.
@@ -222,9 +213,9 @@ class RosePopener:
         sys.stdout.flush()
         try:
             if kwargs.get("shell"):
-                proc = Popen(args[0], **kwargs)
+                proc = Popen(args[0], text=True, **kwargs)
             else:
-                proc = Popen(args, **kwargs)
+                proc = Popen(args, text=True, **kwargs)
         except OSError as exc:
             if exc.filename is None and args:
                 exc.filename = args[0]
--- a/metomi/rose/popen.py
+++ b/metomi/rose/popen.py
@@ -188,9 +188,17 @@ class RosePopener:
             stdin = kwargs.get("stdin")
         stdout, stderr = proc.communicate(stdin)
         retcode = proc.wait()
-        if isinstance(retcode, bytes):
-            retcode = retcode.decode()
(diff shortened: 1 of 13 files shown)


metomi/rose/popen.py
```diff
@@ -188,9 +188,17 @@ def run(self, *args, **kwargs):
             stdin = kwargs.get("stdin")
         stdout, stderr = proc.communicate(stdin)
         retcode = proc.wait()
-        if isinstance(retcode, bytes):
-            retcode = retcode.decode()
-        return retcode, stdout, stderr
+
```
wxtim, line 191: This changes removes the need for a goodish number of decodes elsewhere.

metomi/rose/popen.py
```diff
@@ -188,9 +188,17 @@ def run(self, *args, **kwargs):
...
+        def _decode_if_bytes(obj):
+            if isinstance(obj, bytes):
+                return obj.decode()
+            return obj
+
+        return (
+            _decode_if_bytes(retcode),
+            _decode_if_bytes(stdout),
+            _decode_if_bytes(stderr)
+        )
```
MetRonnie, line 201: Instead of doing this, why not use `Popen(..., text=True)` below in `def run_bg()`?

https://docs.python.org/3.7/library/subprocess.html#subprocess.run

MetRonnie:
Hmm, I think the tests are failing because using `text=True` also means it expects the command/stdin to be `str` but it's being encoded to `bytes` here
```
File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/metomi/rose/host_select.py", line 427, in select
    proc.stdin.write(stdin.encode('UTF-8'))
```

wxtim:
> Hmm, I think the tests are failing because using `text=True` also means it expects the command/stdin to be `str` but it's being encoded to `bytes` here
> 
> ```
> File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/metomi/rose/host_select.py", line 427, in select
>     proc.stdin.write(stdin.encode('UTF-8'))
> ```

Thank you Ronnie!

MetRonnie (commented):
Possible missed cases? (I am not sure if these `.decode()`s are now broken, the lack of docstrings/type annotations makes it difficult to tell)
- https://github.com/metomi/rose/blob/05805e48f93ccde05ce588fbf87e459be60e1aec/metomi/rose/loc_handlers/rsync.py#L66-L67
- https://github.com/metomi/rose/blob/05805e48f93ccde05ce588fbf87e459be60e1aec/metomi/rose/suite_engine_procs/cylc.py#L329-L330
- https://github.com/metomi/rose/blob/05805e48f93ccde05ce588fbf87e459be60e1aec/metomi/rose/suite_engine_procs/cylc.py#L435

wxtim:
@MetRonnie  - Well spotted: I think that only the middle one is technically directly part of this PR, but I don't see any reason not to fix the lot. Will do.

datamel (approved):
I have read and checked out the code and run the tests locally.
The doc failure is unrelated. No problems found from me. Thanks @wxtim.

metomi/rose/popen.py
```diff
@@ -259,7 +258,7 @@ def run_ok(self, *args, **kwargs):
         ret_code, stdout, stderr = self.run(*args, **kwargs)
         if ret_code:
             if stderr:
-                stderr = stderr.decode()
+                stderr = stderr
             else:
                 stderr = ''
```
MetRonnie, line 263: These lines are redundant now, I believe

MetRonnie (approved):
One optional comment, but otherwise 👍

metomi/rose/popen.py
wxtim, line 263:
```suggestion
```

wxtim: I think that you are right.

PR #2510, opened 2021-11-30
Re-enable Rosie ID.

### LINKED ISSUE #2432: rosie: Cylc8 support for rosie id
See TODO in `_from_location` method.

Extract version control information from installed Cylc8 workflows for the `rosie id` command.

Depends on: https://github.com/cylc/cylc-flow/issues/3849

wxtim:
Closes #2432 

> Extract version control information from installed Cylc8 workflows for the rosie id command.
>
> Depends on: cylc/cylc-flow#3849

Requires #2509  merger to pass

- [x] Re-written basic functionality.
- [x] Re-enables t/rosie-id/00-basic.t tests 31-33 which tests functionality.

### DIFF
--- a/metomi/rosie/suite_id.py
+++ b/metomi/rosie/suite_id.py
@@ -421,17 +421,17 @@ class SuiteId:
         loc = Path(location).expanduser().resolve()
         if suite_dir_rel_root:
             sdrr = Path('~', suite_dir_rel_root).expanduser().resolve()
-        try:
-            loc.relative_to(sdrr)
-        except ValueError:
-            # Not an installed Cylc8 workflow run directory
-            pass
-        else:
-            # Slightly odd construction = loc + parents
-            for loc in (loc.relative_to(sdrr) / '_').parents:
-                vcfilepath = sdrr / loc / SuiteId.VC_FILENAME
-                if os.access(vcfilepath, os.F_OK | os.R_OK):
-                    return vcfilepath
+            try:
+                loc.relative_to(sdrr)
+            except ValueError:
+                # Not an installed Cylc8 workflow run directory
+                pass
+            else:
+                # Slightly odd construction = loc + parents
+                for loc in (loc.relative_to(sdrr) / '_').parents:
+                    vcfilepath = sdrr / loc / SuiteId.VC_FILENAME
+                    if os.access(vcfilepath, os.F_OK | os.R_OK):
+                        return vcfilepath
         return None
 
     @staticmethod
--- a/metomi/rosie/suite_id.py
+++ b/metomi/rosie/suite_id.py
@@ -24,7 +24,7 @@ Functions:
     main - CLI interface function.
 
 """
-
+from ast import literal_eval
 import os
 from pathlib import Path
 import re
@@ -118,6 +118,7 @@ class SuiteId:
     REC_IDX = re.compile(r"\A(?:(\w+)-)?(\w+)(?:/([^\@/]+))?(?:@([^\@/]+))?\Z")
     BRANCH_TRUNK = "trunk"
     REV_HEAD = "HEAD"
+    VC_FILENAME = "log/version/vcs.conf"
     svn = SvnCaller()
 
     STATUS_CR = "X"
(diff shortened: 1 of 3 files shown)


metomi/rosie/suite_id.py
```diff
@@ -456,6 +404,77 @@ def _from_location(self, location):
...
+
+        Returns:
+            If a file is found, a path, else None.
+        """
+        suite_engine_proc = SuiteEngineProcessor.get_processor()
+        suite_dir_rel_root = getattr(
+            suite_engine_proc, "SUITE_DIR_REL_ROOT", None
+        )
+        loc = Path(location).expanduser().resolve()
+        sdrr = Path('~', suite_dir_rel_root).expanduser().resolve()
```
datamel, line 422: I think this line might break with traceback if you get the default None back from `suite_dir_rel_root`. Maybe put in try/except?

wxtim:
```suggestion
        if suite_dir_rel_root:
            sdrr = Path('~', suite_dir_rel_root).expanduser().resolve()
        else:
            return None
```

metomi/rosie/suite_id.py
```diff
@@ -456,6 +404,77 @@ def _from_location(self, location):
...
+        except ValueError:
+            # Not an installed Cylc8 workflow run directory
+            pass
+        else:
+            # Slightly odd construction = loc + parents
+            for loc in (loc.relative_to(sdrr) / '_').parents:
+                vcfilepath = sdrr / loc / SuiteId.VC_FILENAME
+                if os.access(vcfilepath, os.F_OK | os.R_OK):
+                    return vcfilepath
+            return None
```
wxtim, line 434:
```suggestion
        return None
```

metomi/rosie/suite_id.py
```diff
@@ -456,6 +404,78 @@ def _from_location(self, location):
...
+        """
+        suite_engine_proc = SuiteEngineProcessor.get_processor()
+        suite_dir_rel_root = getattr(
+            suite_engine_proc, "SUITE_DIR_REL_ROOT", None
+        )
+        loc = Path(location).expanduser().resolve()
+        if suite_dir_rel_root:
+            sdrr = Path('~', suite_dir_rel_root).expanduser().resolve()
+        try:
+            loc.relative_to(sdrr)
```
datamel, line 425: Think you will get an sdrr is not defined if `suite_dir_..` is None here. Could indent block into the `if` to fix.

wxtim: Yes.

datamel (commented):
The tests are failing for me locally with: `location not associated with a suite ID`
Are they passing for you @wxtim?

datamel:
> The tests are failing for me locally with: `location not associated with a suite ID` 

Update: Tried again this morning with a fresh environment and this has fixed my problems @wxtim.  I think pip installing (including cache clearing) everything is not enough sometimes.

t/rosie-id/00-basic.t
```diff
@@ -117,28 +117,26 @@ file_cmp "$TEST_KEY.err" "$TEST_KEY.err" </dev/null
 # TODO: Cylc8 support for rosie id
```
datamel, line 117: You mentioned this TODO can go now.

datamel (approved):
Thanks @wxtim. Working as expected now for me.

t/rosie-id/00-basic.t
```diff
@@ -117,28 +117,26 @@ file_cmp "$TEST_KEY.err" "$TEST_KEY.err" </dev/null
...
 # https://github.com/metomi/rose/issues/2432
```
wxtim, line 119:
```suggestion
```

metomi/rosie/suite_id.py
```diff
@@ -456,6 +404,78 @@ def _from_location(self, location):
...
+            ...         )
+            >>> mypath = MockFpath()
+            >>> SuiteId.parse_cylc_vc_file(mypath)
+            '/a/b/c@4242'
+        """
+        location = None
+        vcsystem = None
+        url = None
+        rev = None
+        for line in fpath.read_text().split('\n'):
```
MetRonnie, line 462: Is it not possible to parse this using Rose's parser? I guess not because Rose doesn't allow multiline strings like
```
status = """
 M flow.cylc
 M whatever.txt
"""
```
?

Then perhaps it would make sense to save VCS info as JSON instead, as you suggested. But I think that would have to be a Cylc 8.0rc1 PR

MetRonnie (approved):
1 question - not a blocker

metomi/rosie/suite_id.py
wxtim: It's an internal which users shouldn't see: so IMO we should create an issue to change this beyond rc1. I don't think that it needs to be a blocker. 
If we want to be failsafe I can't think that it'd be massively expensive to write this + json both.

Issue #2504, opened 2021-11-22
Python 3/Rose 2 General Checks on Documentation


wxtim:
Update the documentation - including the following items:

- [x] Look for refs to ``rose`` not ``metomi.rose`` in code examples.
- [x] Look for Python2 Shebang Lines in examples.

PR #2518, opened 2021-12-07
Removed obvious references to Cylc from glossary.

### LINKED ISSUE #2372: Local vs remote env vars in root-dir config.
NIWA is moving to use of communal (project member) project areas, including for suite output, so we need to configure `rose suite-run` to symlink suite run directories to the right project directories.

Some users have more than one project on the go at once though, so hardwiring the project in login scripts isn't practical.  Instead, I figured we could use a shell variable `$PROJECT` like this:
```
# rose.conf
[rose suite-run]
root-dir=*=/project/${PROJECT}/${USER}
```
...and advise users to set `$PROJECT` to the appropriate value in their local terminal session before executing `rose suite-run`.

However, this fails if the suite has a remote host (as most do) because `rose suite-run` re-invokes itself on the remote host to install files, and `$PROJECT` (for `root-dir`) is not defined there: 
```
[oliverh@w-nwp01 ~/suites/suitex]$ export PROJECT=niwa00004
[oliverh@w-nwp01 ~/suites/suitex]$ rose suite-run -i
[INFO] export CYLC_VERSION=7.8.3
[INFO] export ROSE_ORIG_HOST=w-nwp01.maui.niwa.co.nz
[INFO] export ROSE_SITE=
[INFO] export ROSE_VERSION=2019.01.2
[INFO] create: /project/niwa00004/oliverh/cylc-run/suitex
[INFO] symlink: /project/niwa00004/oliverh/cylc-run/suitex <= /home/oliverh/cylc-run/suitex
...
[FAIL] ssh -oBatchMode=yes -n remote.maui.niwa.co.nz env\ ROSE_VERSION=2019.01.2\ 
   CYLC_VERSION=7.8.3\ bash\ -l\ -c\ \'\"$0\"\ \"$@\"\'\ rose\ suite-run\ -vv\ -n\ suitex\ --install-only\ 
   --run=run\ --remote=uuid=8fb5aaf8-9692-447a-b0c4-92a4726c3302,
   root-dir=\'/nesi/nobackup/${PROJECT}/${USER}\'
   # return-code=1, stderr=
[FAIL] [FAIL] 2019-07-22T04:16:01+0000 [UNDEFINED ENVIRONMENT VARIABLE] PROJECT
```
As a possibly-temporary measure, the following patch works with recent releases, by getting `rose suite-run` to interpolate local environment variables into "host config" values before re-invoking itself on remote hosts:
```diff
--- a/lib/python/rose/suite_run.py
+++ b/lib/python/rose/suite_run.py
@@ -400,6 +400,9 @@ class SuiteRunner(Runner):
                 value = self._run_conf(key, host=host, conf_tree=conf_tree)
                 if value is not None:
                     val = self.popen.list_to_shell_str([str(value)])
+                    # NIWA: interpolate local env vars into host_confs values.
+                    # (We use root-dir=/projects/$PROJECT/$USER, for example.)
+                    val = env_var_process(val)
                     shcommand += ",%s=%s" % (key, pipes.quote(val))
                     locs_conf.set([auth, key], value)
             command.append(shcommand)
```

But ... is this the appropriate permanent fix? Mabye others need different `root-dir` (etc.) on different hosts configured according to different login environments on those hosts (as per the status quo)?

wxtim:
### Closes #2372 
Made build work with Cylc Glossary references.
Not yet a full re-write (#2504 ), will embark on that, based on this branch, but if you are happy you can merge so that the issue gets closed.


oliver-sanders (approved):

oliver-sanders:
I think #2372 was the wrong issue?

PR #2522, opened 2021-12-10
removed the rsync from prune-remote-logs-at

### LINKED ISSUE #2469: rose prune: remove rysnc of remote logs
The rose prune built-in app currently performs an rsync of remote job log files when using the `prune-remote-logs-at=cycle` option.
https://github.com/metomi/rose/blob/2.0b1/sphinx/api/built-in/rose_prune.rst#configuration
We propose removing this functionality.

I think this option dates back to the early days of Cylc when job log retrieval was done by Rose via hook scripts. Performing the log retrieval immediately after task completion sometimes resulted in incomplete log files so the extra rsync prior to pruning remote log files was useful. However, remote log retrieval is now done by Cylc and a configurable delay can be used which should ensure incomplete log files don't occur (or, at least, are very rare). Therefore, the rsync done by rose prune is not necessary and results in additional load on the server running the scheduler (which is where rose prune tasks typically run). Furthermore, Cylc allows you to configure the max size of the job log files to retrieve but this is not used by rose prune which can result in large log files being transferred unexpectedly. Cylc also allows you to not perform remote log retrieval but, again, rose prune will currently perform the rsync regardless.

wxtim:
Closes #2469 

AFAI Can Tell there is not an existing test for this functionality, so I've added a test for it's removal, although this seems ugly.


dpmatthews:
Looks good in principle but I'll leave it to others to review properly

wxtim:
Not at all sure that this works as advertised. :(

datamel (commented):
May benefit from @oliver-sanders taking a look (if/when time permits). @wxtim gave me a great demo although we are unsure it is behaving as intended.

metomi/rose/suite_engine_procs/cylc.py
```diff
@@ -302,20 +303,21 @@ def job_logs_pull_remote(
...
+                            cmd_list.append("--exclude=" + exclude)
+                        cmd_list.append("%(auth)s:%(log_dir_rel)s/" % data)
+                        cmd_list.append(log_dir)
+                        try:
+                            cmd = self.popen.get_cmd(*cmd_list)
+                            self.popen(*cmd)
+                        except RosePopenError as exc:
+                            self.handle_event(exc, level=Reporter.WARN)
+                        if not prune_remote_mode:
+                            continue
```
datamel, line 320: I think these two lines need de-denting?

wxtim:
```suggestion
                    if not prune_remote_mode:
                        continue
```

datamel (commented):
Thanks Tim, unfortunately test isn't passing for me locally. Checking GH Actions it is skipping : `./t/rose-task-run/42-app-prune-no-rsync.t .......................... skipped: This test requires a remote host to be setup in your rose.conf file.`. 
Is it possible to get this running on GH Actions? I shall work on getting it passing locally, as I suspect that is environment related (as so many of my problems are 😞 )

wxtim:
> Thanks Tim, unfortunately test isn't passing for me locally. Checking GH Actions it is skipping : `./t/rose-task-run/42-app-prune-no-rsync.t .......................... skipped: This test requires a remote host to be setup in your rose.conf file.`. Is it possible to get this running on GH Actions? I shall work on getting it passing locally, as I suspect that is environment related (as so many of my problems are disappointed )

I think that getting it running on GH Actions is well and truly out of scope, although it would be nice to do: I'll get back to you on teams about the setup.

datamel (approved):
Great, thanks @wxtim. All working correctly, note `retrieve job logs = true` will cause the test to fail.

metomi/rose/suite_engine_procs/cylc.py
```diff
@@ -208,7 +208,8 @@ def job_logs_archive(self, suite_name, items):
                 pass
 
     def job_logs_pull_remote(
-        self, suite_name, items, prune_remote_mode=False, force_mode=False
+        self, suite_name, items, prune_remote_mode=False, force_mode=False,
+        rsync=False
```
oliver-sanders, line 212: I think the default for `rsync` should be `True` if we are only disabling it for this one case?

wxtim: Fixed, will push later.

wxtim:
n.b. This is now broken with Cylc master if you don't have the branch from https://github.com/cylc/cylc-rose/pull/108 checked out.

oliver-sanders (approved):

oliver-sanders:
(rerunning the tests just to be safe)

PR #2519, opened 2021-12-07
Refurbish Rose Docs

wxtim:
Closes #2504 
Closes #2503 

Read through entire user guide section, except "further topics".

- [x] Removed cheat sheet because it is entirely about things which aren't in the Cylc 8 Ecosystem.
- [x] Removed references to `rose suite-run`, replaced with notes on Cylc scripts where possible.
- [x] Removed explanation of Suite Dir vs. Rundir as this is now `cylc install`'s problem.
- [x] Re-written tutorial about Rose Suites. Mostly passed over much more quickly with links to in-depth docs for Cylc Validate/Install/Play, leaving material about Rose configs.
- [x] Removed references to `rose host-select`. It still exists, but we want people to use platform selection.
- [x] Referred to `[template variables]` section rather than `[jinja2:suite.rc]` wherever possible.
- [x] Checked the further topics tutorials.
- [x] Fixed a [small Python 3](https://stackoverflow.com/questions/11941817/how-to-avoid-runtimeerror-dictionary-changed-size-during-iteration-error)  bug in validation script.
- [x] Grepped for dead-ends from `metomi/rose/rose.py` in the docs*

\* See
```
cd ~/rose/sphinx
for deadend in $(python -c "from metomi.rose.rose import DEAD_ENDS; x = ['.'.join(end) for end in DEAD_ENDS]; [print(i) for i in x]"); do 
  echo $deadend
  grep -rin "$deadend" *
done
``` 

--------------

### Additional changes after review 1 from @oliver-sanders 
- [x] Searched for instances of `cylc gui` and changed if req'd.
- [x]  `cylc get-config` → `cylc config`
- [x] `[cylc]` → `[scheduler]`
- [x] Ran a crude check for the validity of Cylc code snippets.

PR #2526, opened 2021-12-13
Docs.revamp.rose macro tutorial

wxtim:
Update the basic rose macro tutorial to work with packaged rose.


wxtim:
Merged into #2519

PR #2528, opened 2021-12-14
Rose Date only warns if run in an interactive shell

### LINKED ISSUE #2523: `pygraphviz` dependency
Currently there is a dependency on `pygraphviz` which is not listed in `setup.py`.
https://github.com/metomi/rose/blob/19e6c87bf6045950d42e684219ec22ee6656e880/metomi/rosie/graph.py#L22

Digging around I found this discussion: https://github.com/metomi/rose/pull/2366#discussion_r303263303

### LINKED ISSUE #2524: Rose Date: Remove deprecation warning
Rose date is used inside other scripts and the deprecation warning (use isodatetime) that it emits to stderr may break many of these causing customer pain.

Should we:
a) Remove this warning
b) Remove this warning in non-interactive mode?

n.b. Currently we don't have a wrapper script to allow users to access isodatetime in the way they can access Rose and Cylc.

wxtim:
Closes #2524 

Easy test of whether this is working 
```
# Control - should have the warning
rose date
# Changed from master
cd ~/rose/sphinx
make clean html
```

The second case should produce the warning on master, but not on this branch.

### DIFF
--- a/metomi/rose/date_cli.py
+++ b/metomi/rose/date_cli.py
@@ -180,10 +180,11 @@ from metomi.isodatetime.main import main as iso_main
 
 def main():
     """Implement rose date."""
-    print(
-        'WARNING: "rose date" is depreacted, use the "isodatetime" command.',
-        file=sys.stderr
-    )
+    if sys.stdin.isatty():
+        print(
+            'WARNING: "rose date" is depreacted, use the "isodatetime" command.',
+            file=sys.stderr
+        )
 
     if '--help' in sys.argv:
         print('\n' + __doc__)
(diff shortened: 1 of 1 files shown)


metomi/rose/date_cli.py
```diff
@@ -182,7 +182,8 @@ def main():
     """Implement rose date."""
     if sys.stdin.isatty():
         print(
-            'WARNING: "rose date" is depreacted, use the "isodatetime" command.',
+            'WARNING: "rose date" is depreacted, use the "isodatetime" '
```
datamel, line 185:
```suggestion
            'WARNING: "rose date" is deprecated, use the "isodatetime" '
```

datamel (approved):
Thanks @wxtim, tested as working.

oliver-sanders (approved):

oliver-sanders:
Squashing.

PR #2540, opened 2022-02-02
pygraphviz < 1.8

wxtim:
Pygraphviz 1.8 dropped support for Python 3.7

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -199,11 +199,10 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y graphviz pkg-config libgraphviz-dev
-          pip install pygraphviz
 
       - name: install
         run: |
-          pip install -e .[docs]
+          pip install -e .[docs,graph]
           pip install git+https://github.com/cylc/cylc-flow.git
           pip install git+https://github.com/cylc/cylc-rose.git
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -74,7 +74,7 @@ docs =
     sphinxcontrib-httpdomain
     sphinxcontrib-svg2pdfconverter
 graph =
-    pygraphviz
+    pygraphviz>1.0,!=1.8
 rosa =
 tests =
     flake8>=3.0.0


oliver-sanders (changes requested):
There should be no need to specify a version.

oliver-sanders:
Looks like this was caused by an issue with the pygraphviz packaging that is triggered by `pip >= 22`:

https://github.com/pygraphviz/pygraphviz/issues/395

oliver-sanders:
There is a workaround we can use for the time being, we could include the `pygraphviz` dependency twice, once for Python 3.7 and once for 3.8+ using environment markers to select appropriately:

https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies

.github/workflows/test.yml
```diff
@@ -199,7 +199,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y graphviz pkg-config libgraphviz-dev
-          pip install pygraphviz
+          pip install pygraphviz==1.7.*
```
MetRonnie, line 202: Instead of doing this, why don't we do
```
pip install -e .[docs,graph]
```
on L206?

oliver-sanders: No reason, probably an omission of mine.

setup.cfg
```diff
@@ -74,7 +74,7 @@ docs =
     sphinxcontrib-httpdomain
     sphinxcontrib-svg2pdfconverter
 graph =
-    pygraphviz
+    pygraphviz<1.8
```
matthewfeickert, line 77: Hi. :wave: Coming here from `pygraphviz` Issue 395, and I'd like to suggest that you just exclude the broken version rather than capping `pygraphviz` below `1.8` artificially. 

```suggestion
    pygraphviz>1.0,!=1.8
```

`pygraphviz` has the correct `python_requires` (c.f. https://github.com/pygraphviz/pygraphviz/pull/394) so nothing will break in the future once they fix the `pip` `v22.X` issue they have now.

For more on this please see https://iscinumpy.dev/post/bound-version-constraints/

:+1:

.github/workflows/test.yml
oliver-sanders: @wxtim could you remove this line and add the optional output `graph` to the `pip install -e .[docs]` line below. That way it will use the pygraphviz version range specified in the `setup.cfg` file.

wxtim:
@MetRonnie @oliver-sanders - Hopefully now the diff we want. Commits rationalized.

oliver-sanders (approved):

MetRonnie (approved):

PR #2556, opened 2022-03-16
get_suite_job_auths should use install targets not host from platform.

### LINKED ISSUE #2549: rose-task-run file deletion
Tests `t/rose-task-run/{34,35}` are failing.

It looks like an additional delete operation is being performed where it isn't required.

wxtim:
Closes #2549 

### Summary
Tests `t/rose-task-run/{34,35}` are failing.
It looks like an additional delete operation is being performed where it isn't required.

`get_suite_jobs_auths` in `metomi/rose/suite_engine_procs/cylc` should be returning install targets rather than hosts.

### How to test

These tests need to be run from a host with no shared file system.

### Checklist

- [x] Fixed issue in #2549 
- [x] Already tested.
- [ ] Change log updated.

### DIFF
--- a/metomi/rose/suite_engine_procs/cylc.py
+++ b/metomi/rose/suite_engine_procs/cylc.py
@@ -67,34 +67,45 @@ class CylcProcessor(SuiteEngineProcessor):
         """
         return os.path.join(cls.SUITE_DIR_REL_ROOT, suite_name, *paths)
 
+    @staticmethod
     def get_suite_jobs_auths(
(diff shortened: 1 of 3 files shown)


wxtim:
### Investigation into how this works at Cylc 7

At Cylc 7 `get_suite_jobs_auths` returns `['host1', 'host2']`.

wxtim:
Discovered that tests are failing because they are broken as documented on the associated issue.

PR #2563, opened 2022-03-23
docs: add reference labels for builtin apps

oliver-sanders:
Sneaking some reference labels into the docs for intersphinx links.


datamel (approved):
Checked out and built. Does `rose-arch` need one too? 
Approving anyway as can be added at a later stage if necessary.

oliver-sanders:
Rose Arch already has one, just not with the builtin prefix.

datamel:
One review will do.

PR #2570, opened 2022-03-29
Yarn

oliver-sanders:
Address - https://github.com/metomi/rose/security/dependabot/1

By rebuilding the yarn.lock file (to pull in a later version of this transient dep)


MetRonnie (approved):
Have manually tested linting still works after doing `yarn install`

PR #2578, opened 2022-04-09
Rose env cat fix

wxtim:
Fixes two bugs found by tester.

#### Missing input file produces traceback
Handle this more elegantly.

#### Write to output file doesn't work
Trying to write a string to a file handle opened in bytes mode.

#### Testing/Refactor
The first commit is a refactor without functional changes to allow for testing, and tests for the issues described. Checking f8126199 out should cause two of the three tests (`pytest metomi/rose/tests/test_env_cat.py   -vv`) to fail.

The second commit 1abaca8fe3ae7ef655baf1cffab30a60a532704b fixes both problems.

#### Additional check
By inserting print statements into the code I was able to ascertain that there coverage is now ~100%, though this is only a measure of quantity, not quality.

Tagged @oliver-sanders as a single Python test :. this is a smallish change with @dpmatthews as a functional tester.

### DIFF
--- a/metomi/rose/env_cat.py
+++ b/metomi/rose/env_cat.py
@@ -30,12 +30,18 @@ def rose_env_cat(args, opts):
     if not opts.output_file or opts.output_file == "-":
         out_handle = sys.stdout
     else:
-        out_handle = open(opts.output_file, "wb")
+        out_handle = open(opts.output_file, "w")
     for arg in args:
         if arg == "-":
             in_handle = sys.stdin
         else:
-            in_handle = open(arg)
+            try:
+                in_handle = open(arg)
+            except FileNotFoundError as exc:
+                Reporter().report(exc)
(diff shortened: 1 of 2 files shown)


metomi/rose/env_cat.py
```diff
@@ -19,10 +19,50 @@
...
+                        line, opts.unbound, opts.match_mode
+                    )
+                )
+            except UnboundEnvironmentVariableError as exc:
+                name = arg
+                if arg == "-":
+                    name = "<STDIN>"
+                sys.exit("%s:%s: %s" % (name, line_num, str(exc)))
+        in_handle.close()
+    out_handle.close()
```
MetRonnie, line 63: These should probably come in a `try... finally` block, or use `with open()` instead, but understandable to leave as is

MetRonnie (approved):

PR #2579, opened 2022-04-10
update set_time_zone_to_... methods

wxtim:
Use new names of Isodatetime interfaces.

### DIFF
--- a/metomi/rose/date.py
+++ b/metomi/rose/date.py
@@ -176,7 +176,7 @@ class RoseDateTimeOperator:
                 )
                 parse_format = time_point.dump_format
         if self.utc_mode:
-            time_point.to_utc()
+            time_point = time_point.to_utc()
         return time_point, parse_format
 
     def date_shift(self, time_point=None, offset=None):
--- a/metomi/rose/date.py
+++ b/metomi/rose/date.py
@@ -301,8 +301,7 @@ class RoseDateTimeOperator:
     @classmethod
     def get_datetime_strftime(cls, time_point, print_format):
         """Use the datetime library's strftime as a fallback."""
-        calendar_date = time_point.copy().to_calendar_date()
-        year, month, day = calendar_date.get_calendar_date()
+        year, month, day = time_point.get_calendar_date()
         hour, minute, second = time_point.get_hour_minute_second()
         microsecond = int(1.0e6 * (second - int(second)))
         hour = int(hour)
--- a/metomi/rose/date.py
+++ b/metomi/rose/date.py
@@ -146,7 +146,7 @@ class RoseDateTimeOperator:
             time_point_str = self.ref_point_str
         if time_point_str is None or time_point_str == self.STR_NOW:
             time_point = get_timepoint_for_now()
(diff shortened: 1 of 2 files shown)


metomi/rose/date.py
```diff
@@ -146,7 +146,7 @@ def date_parse(self, time_point_str=None):
             time_point_str = self.ref_point_str
         if time_point_str is None or time_point_str == self.STR_NOW:
             time_point = get_timepoint_for_now()
-            time_point.set_time_zone_to_local()
+            time_point.to_local_time_zone()
```
MetRonnie, line 149:
```suggestion
```
This line isn't doing anything for 2 reasons:
- `TimePoint`s are now immutable, so this returns a new `TimePoint` without affecting the variable `time_point`
- `get_timepoint_for_now()` already is in the local time zone by default, unless you specify the arg `utc=True`

metomi/rose/date.py
```diff
@@ -177,7 +177,7 @@ def date_parse(self, time_point_str=None):
                 )
                 parse_format = time_point.dump_format
         if self.utc_mode:
-            time_point.set_time_zone_to_utc()
+            time_point.to_utc()
```
MetRonnie, line 180:
```suggestion
            time_point = time_point.to_utc()
```

MetRonnie (changes requested):
I spotted another needed fix:
```diff
diff --git a/metomi/rose/date.py b/metomi/rose/date.py
index dd4511c3..bfcea15c 100644
--- a/metomi/rose/date.py
+++ b/metomi/rose/date.py
@@ -302,7 +302,7 @@ class RoseDateTimeOperator:
     @classmethod
     def get_datetime_strftime(cls, time_point, print_format):
         """Use the datetime library's strftime as a fallback."""
-        calendar_date = time_point.copy().to_calendar_date()
+        calendar_date = time_point.to_calendar_date()
         year, month, day = calendar_date.get_calendar_date()
         hour, minute, second = time_point.get_hour_minute_second()
         microsecond = int(1.0e6 * (second - int(second)))
```

Also, I think we should specify this version range, because we rely on the 3.x API:
```diff
diff --git a/setup.cfg b/setup.cfg
index d89a5b80..40e9fe36 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,7 +53,7 @@ install_requires =
     aiofiles
     jinja2>=2.10.1
     ldap3
-    metomi-isodatetime
+    metomi-isodatetime==3.*
     psutil>=5.6.0
     requests
     sqlalchemy
```
Possibly even `3.0.*`? cc @oliver-sanders

setup.cfg
```diff
@@ -53,7 +53,7 @@ install_requires =
     aiofiles
     jinja2>=2.10.1
     ldap3
-    metomi-isodatetime
+    metomi-isodatetime==1!3.*
```
MetRonnie, line 56: Good spot about the epoch

MetRonnie (approved):

oliver-sanders (approved):

PR #2581, opened 2022-04-11
re-enable rose date --offset 1w1d1h1m1s

### LINKED ISSUE #2577: rose date: elegantly deprecate \d+[wdhms] format
Rose 2019 supported a deprecated offset format:

https://github.com/metomi/rose/blob/8d1f144b3724734ccad1a347a8b0cf106393f3c1/lib/python/rose/date.py#L206-L214

E.G:

```console
$ rose date now --offset 1h
```

This was listed in the CLI help as deprecated, however, no deprecation warning was produced at runtime.

This back support was dropped at Rose 2, however, is still in use. Due to the prevalent use of `rose date` in bash scripts there is a high likelihood of hard-to-debug errors in Rose 2 resulting from this change.

**Options:**

1) Add a deprecation warning in Rose 2019.
   * Easy enough, users would need to be encouraged to grep through logs for this warning.
2) Continue to support this, but only in `rose date` (and not the newer `isodatetime` command).
   * Less easy as the CLI has not been parsed at this stage so we would need to crudely parse it in Rose.
   * We would also need to apply this fix here https://github.com/metomi/isodatetime/pull/204

Tagging this against 2.0rc3 as a reminder for us to solve this one on that timeframe rather than as a leaning to option (2).

wxtim:
Closes #2577 

#### Summary

Re-enable deprecated syntax for rose-date where `--offset \d[wdhms]`

### DIFF
--- a/metomi/rose/date.py
+++ b/metomi/rose/date.py
@@ -443,19 +443,16 @@ def upgrade_offset(offset: str) -> str:
 
     Examples:
         >>> upgrade_offset('1w')
-        [WARN] This offset syntax is deprecated.
+        [WARN] This offset syntax 1w is deprecated: Using P7DT0H0M0S
         'P7DT0H0M0S'
         >>> upgrade_offset('1w1d1h')
-        [WARN] This offset syntax is deprecated.
+        [WARN] This offset syntax 1w1d1h is deprecated: Using P8DT1H0M0S
         'P8DT1H0M0S'
         >>> upgrade_offset('1h1d')
-        [WARN] This offset syntax is deprecated.
+        [WARN] This offset syntax 1h1d is deprecated: Using P1DT1H0M0S
         'P1DT1H0M0S'
     """
-    Reporter().report(
-        'This offset syntax is deprecated.',
-        prefix=Reporter.PREFIX_WARN, level=Reporter.WARN
-    )
+
     sign = '-' if offset[0] == '-' else ''
     offsets = LEGACY_OFFSET.findall(offset)
     offsets = {i.upper(): j for j, i in offsets}
@@ -478,7 +475,14 @@ def upgrade_offset(offset: str) -> str:
     # Week is not a built-in type:
     days = days + weeks * 7
 
-    return f'{sign}P{days}DT{hours}H{minutes}M{seconds}S'
+    result = f'{sign}P{days}DT{hours}H{minutes}M{seconds}S'
+
+    Reporter().report(
+        f'This offset syntax {offset} is deprecated: Using {result}',
+        prefix=Reporter.PREFIX_WARN, level=Reporter.WARN
+    )
+
+    return result
 
 
 if __name__ == "__main__":
--- a/metomi/rose/date.py
+++ b/metomi/rose/date.py
@@ -455,7 +455,7 @@ def upgrade_offset(offset: str) -> str:
 
     sign = '-' if offset[0] == '-' else ''
     offsets = LEGACY_OFFSET.findall(offset)
-    offsets = {i.upper(): j for j, i in offsets}
+    offsets = {unit.upper(): number for number, unit in offsets}
 
     # Rose 2019 did not make any distinction between 1s1m and 1m1s,
     # so we do an implicit sort here:
--- a/metomi/rose/date.py
+++ b/metomi/rose/date.py
@@ -472,7 +472,6 @@ def upgrade_offset(offset: str) -> str:
         if unit == 'S':
             seconds = int(value)
 
-    # Week is not a built-in type:
     days = days + weeks * 7
 
     result = f'{sign}P{days}DT{hours}H{minutes}M{seconds}S'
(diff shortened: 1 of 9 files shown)


metomi/rose/date_cli.py
```diff
@@ -177,6 +177,40 @@
...
+    >>> _handle_old_offsets(['rose-date', '-s', '1d1s'])
+    [WARN] This offset syntax is deprecated.
+    ['rose-date', '-s', 'P1DT0H0M1S']
+    """
+    for index in range(len(args)):
+        arg = args[index]
+        if '--offset' in arg or '-s' in arg:
+            # Case: --offset=<offset> is a single item in args list:
+            if (
+                len(arg.split('=')) > 1
```
oliver-sanders, line 202: Absolutely fine but can be done more simply:

```suggestion
                '=' in arg
```

metomi/rose/date.py
```diff
@@ -29,6 +29,9 @@
 from metomi.rose.reporter import Reporter
 
 
+LEGACY_OFFSET = re.compile(r'-?(?P<value>\d+)(?P<unit>[wdhms])')
```
MetRonnie, line 32: The `-` doesn't do anything with the way this is used
```suggestion
LEGACY_OFFSET = re.compile(r'(?P<value>\d+)(?P<unit>[wdhms])')
```

metomi/rose/date_cli.py
```diff
@@ -177,6 +177,40 @@
...
+                and LEGACY_OFFSET.match(arg.split('=')[1])
+            ):
+                offset = upgrade_offset(arg.split("=")[1])
+                args[index] = f'{arg.split("=")[0]}={offset}'
+            # Case: --offset <offset> is two items in args list:
+            elif (
+                LEGACY_OFFSET.match(args[index + 1])
+            ):
+                args[index + 1] = upgrade_offset(args[index + 1])
```
oliver-sanders, line 211: I found a couple of facetious inputs that could cause traceback:

```console
$ rose date 20000101 -s= -1d
WARNING: "rose date" is deprecated, use the "isodatetime" command.
[WARN] This offset syntax is deprecated.
Invalid ISO 8601 date representation: \-P1D
$ rose date 20000101 -s= 
IndexError: list index out of range
```

These aren't a big concern but it would be good to handle them if possible. I managed to get around them by rejigging the if statements a little (needs review/testing):

```suggestion
            if '=' in arg:
                if LEGACY_OFFSET.match(arg.split('=')[1]):
                    offset = upgrade_offset(arg.split("=")[1])
                    args[index] = f'{arg.split("=")[0]}={offset}'

            # Case: --offset <offset> is two items in args list:
            elif (
                len(args) > index + 1
                and LEGACY_OFFSET.match(args[index + 1])
            ):
                args[index + 1] = upgrade_offset(args[index + 1])
```

oliver-sanders (approved):
Nicely done, well tested, manually tested on the CLI, two minor issues, otherwise all good.

(worth trying in your shell manually particularly with negative durations due to difference in escaping logic)

metomi/rose/date_cli.py
```diff
@@ -177,6 +177,40 @@
...
+    >>> _handle_old_offsets(['rose-date', '--offset=1d1s'])
+    [WARN] This offset syntax is deprecated.
+    ['rose-date', '--offset=P1DT0H0M1S']
+    >>> _handle_old_offsets(['rose-date', '-s', '1d1s'])
+    [WARN] This offset syntax is deprecated.
+    ['rose-date', '-s', 'P1DT0H0M1S']
+    """
+    for index in range(len(args)):
+        arg = args[index]
+        if '--offset' in arg or '-s' in arg:
```
MetRonnie, line 198:
```suggestion
        if arg.startswith(('--offset', '-s')):
```

wxtim:
> (worth trying in your shell manually particularly with negative durations due to difference in escaping logic)

Yes, this did my head in.

wxtim:
Re-drafted whilst I deal with first review

metomi/rose/date.py
```diff
@@ -427,5 +430,56 @@ def _convert_duration(date_time_oper, opts, args):
...
+        if unit == 'D':
+            days = int(value)
+        if unit == 'H':
+            hours = int(value)
+        if unit == 'M':
+            minutes = int(value)
+        if unit == 'S':
+            seconds = int(value)
+
+    # Week is not a built-in type:
```
MetRonnie, line 478: This comment is misleading; week is a built-in type in isodatetime, but it can only be used by itself (e.g. `P2W`).
```suggestion
```

metomi/rose/date.py
oliver-sanders: Should the regex start ^ and end $ to enforce a full match?

MetRonnie: I don't think that would work with the way this is used (`.findall()`)

oliver-sanders: then `re.match`?

MetRonnie: Would need to change the regex to something like `^-?((\d+)([wdhms]))+$`

wxtim:
@oliver-sanders  - could you double check you approval?

oliver-sanders (approved):

PR #2587, opened 2022-05-08
Add a scheduled check that the tests work

wxtim:
Having just discovered that the tests have been failing on master for 19 days since https://github.com/cylc/cylc-flow/pull/4823 was merged I propose the addition of a daily run of the test battery on weekdays so that repository subscribers get to see a problem early.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,8 @@ on:
       cylc_repo:
         description: The Cylc repo to test against
         required: false
+  schedule:
+  - cron: '37 04 * * 1-5' # 03:37, Monday-Friday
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -19,6 +19,8 @@ on:
       cylc_repo:
         description: The Cylc repo to test against
         required: false
+  schedule:
+  - cron: '37 04 * * 1-5' # 03:37, Monday-Friday
```
MetRonnie, line 23: I thought the idea was to run once or twice a week? Also the comment says 03:37 whereas it is actually 04:37

Issue #2575, opened 2022-03-31
fcm_make: app requires cylc.rose on remote hosts


oliver-sanders:
The fcm_make app currently imports `cylc.rose` on remote platforms which means we need cylc-rose installed on remote platforms which is not as intended.

If possible we should remove this dep, quick investigation to find out why it is imported and whether the import is needed in these situations.

wxtim:
I can't see it:

searched for 
- `from cylc`
- `import cylc`
- `cylc.rose`

```bash
> grep -rn cylc\.rose * --include "*.py" --color
metomi/rose/suite_engine_procs/cylc.py:80:        from cylc.rose.platform_utils import get_platforms_from_task_jobs
metomi/rose/suite_engine_procs/cylc.py:110:        from cylc.rose.platform_utils import get_platform_from_task_def
metomi/rose/tests/suite_engine_procs/test_cylc.py:24:    import cylc.rose.platform_utils
metomi/rose/tests/suite_engine_procs/test_cylc.py:26:    pytestmark = pytest.mark.skip(reason="cylc-rose not found")
metomi/rose/tests/suite_engine_procs/test_cylc.py:54:        cylc.rose.platform_utils,
metomi/rose/tests/suite_engine_procs/test_cylc.py:135:        cylc.rose.platform_utils,
```

The only place it's imported in the non-test parts of the code-base are in `get_suite_jobs_auths` (used by rose-prune) and `get_task_auth` which is used by fcm make, but only in the method `_run_orig`, which I hope isn't being called on the remote machine. (Is it possible that this is where the error is?).

Where it's used in the tests it's got logic for handling it's unavailability.

PR #2590, opened 2022-05-10
Allow get_task_auth (used by fcm make) to return None if

wxtim:
cylc.rose is not installed.

Proved horrendous to test - @dpmatthews  - can you test manually, whilst @oliver-sanders checks for Python goodness?

### DIFF
--- a/metomi/rose/suite_engine_procs/cylc.py
+++ b/metomi/rose/suite_engine_procs/cylc.py
@@ -107,7 +107,12 @@ class CylcProcessor(SuiteEngineProcessor):
         from cylc.flow.exceptions import WorkflowFilesError
         from cylc.flow.hostuserutil import is_remote_platform
         from cylc.flow.platforms import get_host_from_platform
-        from cylc.rose.platform_utils import get_platform_from_task_def
+
+        try:
+            from cylc.rose.platform_utils import get_platform_from_task_def
+        except ModuleNotFoundError:
+            return None
+
         try:
             platform = get_platform_from_task_def(suite_name, task_name)
         except KeyError:
--- a/metomi/rose/suite_engine_procs/cylc.py
+++ b/metomi/rose/suite_engine_procs/cylc.py
@@ -109,10 +109,10 @@ class CylcProcessor(SuiteEngineProcessor):
         from cylc.flow.platforms import get_host_from_platform
         try:
             from cylc.rose.platform_utils import get_platform_from_task_def
-        except ModuleImportError:
+        except ModuleNotFoundError:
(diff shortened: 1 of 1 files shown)


metomi/rose/suite_engine_procs/cylc.py
```diff
@@ -107,7 +107,12 @@ def get_task_auth(
         from cylc.flow.exceptions import WorkflowFilesError
         from cylc.flow.hostuserutil import is_remote_platform
         from cylc.flow.platforms import get_host_from_platform
-        from cylc.rose.platform_utils import get_platform_from_task_def
+
+        try:
+            from cylc.rose.platform_utils import get_platform_from_task_def
+        except ModuleNotFoundError:
+            return None
```
dpmatthews, line 114: Maybe add a comment to explain, e.g.
```
# Allow single stage fcm_make app to work without requiring cylc.rose
```

dpmatthews:
Works fine in my tests

oliver-sanders:
The solution is fine, however, it has the consequence that the interface will always return `None` if cylc-rose is not installed. What are the consequences of this?

dpmatthews:
> The solution is fine, however, it has the consequence that the interface will always return `None` if cylc-rose is not installed. What are the consequences of this?

This function is only used by the fcm_make built-in app. Returning `None` is equivalent to there being no fcm_make2 task found or no workflow file found which is fine - 2 stage fcm_make is only supported on the localhost install target (because the workflow files aren't mirrored).

metomi/rose/suite_engine_procs/cylc.py
```diff
@@ -107,7 +107,13 @@ def get_task_auth(
         from cylc.flow.exceptions import WorkflowFilesError
         from cylc.flow.hostuserutil import is_remote_platform
         from cylc.flow.platforms import get_host_from_platform
-        from cylc.rose.platform_utils import get_platform_from_task_def
+        try:
+            from cylc.rose.platform_utils import get_platform_from_task_def
+        except:
```
wxtim, line 112:
```suggestion
        except ModuleImportError:
```

dpmatthews (approved):

oliver-sanders:
Closes https://github.com/metomi/rose/issues/2575 ?

oliver-sanders (approved):

hjoliver:
> +1 @wxtim could you copy this comment into the docstring, would be good to capture this. The returns section of the docstring could do with an extra bullet point to match.

I've pushed a commit to do this, so I can merge for the forthcoming release...

PR #2591, opened 2022-05-11
Fix datetimes accepted by Rose (2019) Date, but not with isodatetime:

wxtim:
- Unix Format - "%a %b %d %H:%M:%S %Z %Y"
- Pre Cylc 5 Format - %Y%m%d%H

### DIFF
--- a/metomi/rose/date.py
+++ b/metomi/rose/date.py
@@ -30,6 +30,10 @@ from metomi.rose.reporter import Reporter
 
 
 LEGACY_OFFSET = re.compile(r'-?(?P<value>\d+)(?P<unit>[wdhms])')
+CYLC5_FORMAT = re.compile(r'\d{10}')
+UNIX_FORMAT = re.compile(
+    r'\w{3} \w{3} \d\d \d\d:\d\d:\d\d (?P<timezone>\w*\/?\w*) \d{4}'
+)
 
 
 class OffsetValueError(ValueError):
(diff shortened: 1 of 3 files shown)


metomi/rose/date_cli.py
```diff
@@ -213,6 +217,29 @@ def _handle_old_offsets(args: list) -> list:
...
+            not (
+                args[i].startswith('--')
+                and '=' not in args[i]
+            )
+            and not arg.startswith('--')
+        ):
+            if CYLC5_FORMAT.match(arg):
+                args[i + 1] = upgrade_cylc5_datetime(arg)
+            elif UNIX_FORMAT.match(arg):
+                args[i + 1] = upgrade_unix_datetime(arg)
```
MetRonnie, line 239: Possibly consider this change. I found it confusing that `i` did not match the index of `args`
```suggestion
    for i, arg in enumerate(args[1:], start=1):
        if (
            not (
                args[i - 1].startswith('--')
                and '=' not in args[i - 1]
            )
            and not arg.startswith('--')
        ):
            if CYLC5_FORMAT.match(arg):
                args[i] = upgrade_cylc5_datetime(arg)
            elif UNIX_FORMAT.match(arg):
                args[i] = upgrade_unix_datetime(arg)
```

metomi/rose/date.py
```diff
@@ -482,5 +486,47 @@ def upgrade_offset(offset: str) -> str:
...
+
+    Examples:
+        >>> upgrade_cylc5_datetime('2022010101')
+        '20220101T01'
+    """
+    upgraded = f'{datetime[:-2]}T{datetime[-2:]}'
+
+    print(
+        f'[WARN] This datetime syntax {datetime}'
+        f' is deprecated: Using {upgraded}',
```
MetRonnie, line 500: Maybe indicate that the user should take action?
```suggestion
        f' is deprecated. Use {upgraded} instead',
```

metomi/rose/date.py
```diff
@@ -482,5 +486,47 @@ def upgrade_offset(offset: str) -> str:
...
+        Support of Timezones seems to be dependent on system setup and
+        may be very variable.
+    """
+    upgraded = datetime.strptime(
+        datetime_str, "%a %b %d %H:%M:%S %Z %Y"
+    ).strftime("%Y-%m-%dT%H:%M:%S%z")
+
+    print(
+        f'[WARN] This datetime syntax {datetime_str} '
+        f'is deprecated: Using {upgraded}',
```
MetRonnie, line 524:
```suggestion
        f'is deprecated. Use {upgraded} instead',
```

MetRonnie (commented):
Partial review, still haven't looked at tests

metomi/rose/date_cli.py
```diff
@@ -135,7 +135,7 @@
     * ctime: `%a %b %d %H:%M:%S %Y`
     * Unix date: `%a %b %d %H:%M:%S %Z %Y`
     * Basic ISO8601: `%Y-%m-%dT%H:%M:%S`, `%Y%m%dT%H%M%S`
-    * Cylc: `%Y%m%d%H`
+    * Cylc: `%Y%m%d%H` (deprecated)
```
oliver-sanders, line 138:
```suggestion
    * Cylc 5: `%Y%m%d%H` (deprecated)
```

metomi/rose/date_cli.py
```diff
@@ -135,7 +135,7 @@
     * ctime: `%a %b %d %H:%M:%S %Y`
     * Unix date: `%a %b %d %H:%M:%S %Z %Y`
```
oliver-sanders, line 136: Are these not deprecated too?

(by which I mean does `isodatetime` support them)

oliver-sanders:
@wxtim 

The question label is on this issue but I'm not sure what the question is?

> the format "Unix date" %a %b %d %H:%M:%S %Z %Y seems extremely flaky at Cylc 7.

Why?

> I suspect is a dependence on local linux installations and the contents of /usr/share/zoneinfo.

Makes sense, what you get is dependent on local factors.

> Further evidence for this is the failure of tests at https://github.com/metomi/rose/commit/314610e03dee28a413f7e4c181c559c35266ef5f which pass locally for me.

Correct commit?

> Should we deprecate this format?

We are deprecating all of the old formats (that are no longer supported with isodatetime) by back-supporting them in `rose date`.

metomi/rose/date_cli.py
MetRonnie: Supported: https://github.com/metomi/isodatetime/blob/2e7cea4ee01c136e5077105ec5a8872b7df69f5c/metomi/isodatetime/datetimeoper.py#L41-L47

PR #2592, opened 2022-05-11
date: send warning to stderr not stdout

oliver-sanders:
Looks like the `Reporter()` instance didn't check out the way we would have expected it to, easier just to `print()` for the time being.


wxtim (approved):
Legit - But it looks like you'll need to swap the `capsys.readouterr().out` for a `capsys.readouterr().err` in the tests.

I'll have a look at  bringing my PR into line.

@oliver-sanders  -I'll fix

PR #2593, opened 2022-05-16
resource: fix CLI docs

oliver-sanders:


datamel (approved):
I have tested this is accurate.

PR #2604, opened 2022-08-08
Return a sensible warning popup if user presses play when Cylc Versio…

### LINKED ISSUE #2597: edit: play button and Cylc 8 workflows
The Rose Edit GUI has a play button which calls `rose suite-run`.

We need to check that pressing this button (when using the wrapper script with `CYLC_VERSION=8`) results in safe behaviour that the user can understand.

wxtim:
…n == 8.x

Closes #2597 

## Summary

If the user presses the `rose suite-run` button in `rose edit` or `rosie go` return a sensible warning.

## How to test:

With the branch checked out:

```
PYTHONPATH=$PWD/lib/python:$PYTHONPATH ./bin/rosie go
PYTHONPATH=$PWD/lib/python:$PYTHONPATH ./bin/rose edit
```

Try pressing play.

n.b. if you haven't got any rosie workflows you may need to use the create button in Rosie to create one.

### DIFF
--- a/lib/python/rose/gtk/run.py
+++ b/lib/python/rose/gtk/run.py
@@ -20,8 +20,9 @@
 """Miscellaneous gtk mini-applications."""
 
 import multiprocessing
+from subprocess import check_output
 
-from rose.gtk.dialog import DialogProcess
+from rose.gtk.dialog import DialogProcess, run_dialog, DIALOG_TYPE_WARNING
 from rose.opt_parse import RoseOptionParser
 from rose.suite_run import SuiteRunner
 from rose.reporter import Reporter, ReporterContextQueue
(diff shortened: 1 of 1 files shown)


datamel:
All good, I have tested rose edit and rosie go and the errors are raised when trying to rose suite run. One small query, for the option to launch suite control GUI, I get an error that the suite is not registered, should this raise a cylc version error instead?

datamel (approved):

wxtim:
> All good, I have tested rose edit and rosie go and the errors are raised when trying to rose suite run. One small query, for the option to launch suite control GUI, I get an error that the suite is not registered, should this raise a cylc version error instead?

I think that this is OK - if they were running a Cylc 7 workflow with Rose this would also be true until they run `rose-suite-run`. Note that it is greyed out on the command bar.

lib/python/rose/gtk/run.py
```diff
@@ -30,6 +31,17 @@
 def run_suite(*args):
     """Run "rose suite-run [args]" with a GTK dialog."""
 
+    # Check that we are not using Cylc 8: Else return a helpful message:
+    cylc_version = check_output(['cylc', 'version'])
```
oliver-sanders, line 35: There is an existing interface for fetching the Cylc version.

lib/python/rose/suite_engine_procs/cylc.py::get_version

oliver-sanders (commented):
LGTM, one comment, can save a bit of duplication.

oliver-sanders (approved):

PR #2605, opened 2022-08-09
Config proc handler bug

wxtim:
Discovered by running the Cylc Rose tests; To replicate run

```
cd cylc-rose
pytest
```

I suspect, but have not demo'd convincingly that `sys.modules` ([docs](https://docs.python.org/3/library/sys.html#sys.modules)) is a slippery customer, and that this function risks giving different answers depending on how it is being called.


datamel (approved):
I can confirm this fixes the locally failing tests for me on cylc-rose.

metomi/rose/config_processor.py
```diff
@@ -108,11 +106,11 @@ def __init__(self, event_handler=None, popen=None, fs_util=None):
...
-        path = os.path.dirname(
-            os.path.dirname(sys.modules["metomi.rose"].__file__)
-        )
         SchemeHandlersManager.__init__(
-            self, [path], "rose.config_processors", ["process"]
+            self,
+            [self.get_rose_path()],
+            "metomi.rose.config_processors",
+            ["process"]
         )
```
oliver-sanders, line 114: I think `path` is now one level higher than it was before.

```python
>>> import metomi.rose
>>> import sys
>>> init = sys.modules["metomi.rose"].__file__
>>> init
'~/rose/metomi/rose/__init__.py'
>>> import os
>>> from pathlib import Path

# before
>>> os.path.dirname(os.path.dirname(init))
'~/rose/metomi'

# after
>>> Path(init).parent.parent.parent
PosixPath('~/rose')
```

If was `dirname(dirname(__init__.py)`, now it's `Path(__init__.py).parent.parent.parent`.



https://github.com/metomi/rose/blob/75ae8fdfd39d19d57898d5c24e7a485b92007fd8/metomi/rose/scheme_handler.py#L129-L139

wxtim: Yes, it was - I think that it's easier to understand if the next line has the full path.

oliver-sanders: But doesn't this change the behaviour of SchemeHandlersManager?

wxtim: No

MetRonnie: I'm guessing it doesn't because of adding the `metomi.` to the beginning of the `ns` arg?

MetRonnie (approved):
Confirmed it fixes the cylc-rose tests

Found a few more instances that you didn't touch. Not sure if you wanted to update those too?

https://github.com/metomi/rose/blob/0e66d2be5259e671ecbcd73152cf6e8dd7460ff7/metomi/rose/app_run.py#L371-L376

https://github.com/metomi/rose/blob/48d5f6104b05fa0fda0a63eea3f2c5539ff71f95/metomi/rose/suite_engine_proc.py#L288-L294

https://github.com/metomi/rose/blob/0a37faec034cf42e908cd068245f4712705a9dd7/metomi/rose/apps/rose_arch.py#L132-L138

https://github.com/metomi/rose/blob/0a37faec034cf42e908cd068245f4712705a9dd7/metomi/rosie/svn_hook.py#L110-L112

https://github.com/metomi/rose/blob/05805e48f93ccde05ce588fbf87e459be60e1aec/metomi/rosie/svn_post_commit.py#L115-L118

https://github.com/metomi/rose/blob/05805e48f93ccde05ce588fbf87e459be60e1aec/metomi/rosie/svn_pre_commit.py#L67-L70

wxtim:
> Confirmed it fixes the cylc-rose tests
> 
> Found a few more instances that you didn't touch. Not sure if you wanted to update those too?


Honestly I'd prefer not to, although having created a one-size-fits all interface...

wxtim:
@MetRonnie @oliver-sanders @datamel  - Were any of you demanding I do something to this PR before merging?

PR #2608, opened 2022-08-25
Fix tutorials

### LINKED ISSUE #2607: Tutorials are out of date
Some of the tutorials are not working as they are out of date. E.g. `rose-suite-tutorial`, which relies on a copy of `cylc-forecasting-suite` which is out of date with its cylc-flow counterpart.

Are we going to manually copy over from cylc-flow or try and use some system to keep them up to date?

MetRonnie:
Closes #2607

- Copies of the tutorial workflows were out of date and broken. Manually copied tutorial workflows from cylc-flow.
- Sorted out a few other things in the tutorials
- Improved the dead-end command messages


metomi/rose/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
```diff
@@ -23,18 +37,24 @@ from datetime import datetime
...
+from mercator import get_offset, get_scale, pos_to_coord
 import util
 
 
 URL = ('http://datapoint.metoffice.gov.uk/public/data/layer/wxobs/'
        'RADAR_UK_Composite_Highres/png?TIME={time}&key={api_key}')
 
 
-class Rainfall:
+class Rainfall(object):
```
wxtim, line 57: [Not required at python3](https://stackoverflow.com/questions/4015417/why-do-python-classes-inherit-object)

If you are intending to make this version agnostic the shebang line suggests otherwise.

sphinx/tutorial/rose/configurations.rst
```diff
@@ -93,7 +90,7 @@ For example:
 
 .. hint::
 
-   In Rose configuration files settings do not normally require quotation.
+   In Rose configuration files, settings do not normally require quotation.
```
wxtim, line 93: No longer true - settings must be a valid Python or Jinja2 literal!

wxtim: A slightly broader matter, perhaps for another PR, but not if you agree and wish to do it: I think that "Why Use Rose Configurations" should be near the top of this file.

metomi/rose/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
MetRonnie: I don't think this needs to support python 2, because Rose 2 requires python 3.

But there isn't much point making changes here, better to do them on cylc-flow then rsync across

sphinx/tutorial/rose/metadata.rst
```diff
@@ -152,6 +152,9 @@ Metadata Items
       ``application-tutorial`` app we wrote in the
       :ref:`Rose application practical <rose-applications-practical>`.
 
+   In order to use ``rose config-edit``, you may find it easier to do this
+   whole practical using Rose 2019.
+
```
wxtim, line 157: Not too sure about this - users at sites using the Cylc wrapper script should be able to use Rose 2019 config edit without doing anything that we need to tell them here.

```suggestion
    If you get the following message:
    
    .. code-block:: bash
    
       The Rose configuration editor has been removed.
       The old Rose 2019 GUI remains compatible with Rose 2 configurations.
    
    You should open a terminal where ``CYLC_VERSION=7`` if you wish to use Rose 2019 graphical utilities.
    These utilities are compatible with Rose 2 configurations.
```

sphinx/tutorial/rose/configurations.rst
MetRonnie: > No longer true - settings must be a valid Python or Jinja2 literal!

I think that only applies to `[template variables]` settings?

wxtim (approved):
I've spotted a couple of things I'd like you to at least argue with me over.

sphinx/tutorial/rose/configurations.rst
wxtim: You are right, but since that it > 50% of use-cases I think that this is unhelpful

sphinx/tutorial/rose/furthertopics/date-time-manipulation.rst
```diff
@@ -21,13 +21,21 @@ directory.
...
 -------------------------
 
+.. _isodatetime: https://github.com/metomi/isodatetime
+
+.. note::
+
+   ``rose date`` has been superseded by `isodatetime`_ which offers a
+   compatible CLI interface with more features (just swap out ``rose date``
+   for ``isodatetime``), as well as a Python API.
+
```
oliver-sanders, line 31: Rather than documenting this deprecated command, how about `s/rose date/isodatetime/g`?

sphinx/tutorial/rose/suites.rst
```diff
@@ -239,63 +229,63 @@ Using a Rose workflow configuration with Cylc 8
 
       Open the :ref:`command-rose-config-edit` GUI. You should see
       :guilabel:`suite conf` in the panel on the left-hand side of the window.
-      This will contain the environment and template variables we have just
-      defined.
+      This will contain the template variables we have just defined.
 
-   #. **Use Suite Variables In The** ``flow.cylc`` **File.**
+   #. **Use suite variables in the** ``flow.cylc`` **file**
```
oliver-sanders, line 234: This heading was de-capitalised for some reason, but others in the same file were not?

oliver-sanders (commented):
LGTM: couple of small points.

sphinx/tutorial/rose/furthertopics/date-time-manipulation.rst
MetRonnie: Just had a look at doing this. But it turns out `rose date` accepts an option `-c` whereas `isodatetime` does note (even though it is mentioned in the help text)

oliver-sanders:
This one is still outstanding - https://github.com/metomi/rose/pull/2608#discussion_r959548907

MetRonnie:
Done in a2eea911130b1183376ed206c3ed42b8d4dd98ad

oliver-sanders (approved):

PR #2611, opened 2022-09-12
use any python>=2.7 for rsync check

### LINKED ISSUE #2609: rsync handler: reduce dependencies
The rsync handler requires `python3` to be in the `PATH` on any remote host it is run on (note neither Cylc nor Rose need to be installed on the remote hosts for this to work).

```
ssh <host> <opts> python3 <<< program-to-run
```

https://github.com/metomi/rose/blob/3309275978be70a8d7adaf024a6165efd214c0aa/metomi/rose/loc_handlers/rsync.py#L63-L65

If `python3` is not present, or not in the default environment the command will fail.

We should consider ways to make this more robust as it is not guaranteed that `python3` will be present in the system environment. Ideas:

1. Make the script Python 2/3 compatible (one of Python 2 or 3 are reasonably likely to be installed).
2. Re-write in Bash (highly likely to be installed).
3. Re-write in `/bin/sh` (guaranteed to be installed).

Notes for option (1):
* Octals (can move the octal stuff out of the remote code into the server-side code to avoid this issue).
* Print statements (can be handled via `from _future_ import print_function` in Python 2.7, but not in 2.6?).

wxtim:
test rsync loc handlers against many python version

closes #2609


dpmatthews (approved):
Looks good to me
(Note: I haven't reviewed the tests)

.github/workflows/test_rsync_test_script.yml
```diff
@@ -0,0 +1,47 @@
...
+
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      rose_ref:
+        description: The Rose branch to test against
+        required: true
+  schedule:
+  - cron: '37 04 * * 1-5' # 03:37, Monday-Friday
```
oliver-sanders, line 11: Having this both on `pull_request` and `schedule` is overkill.

How about running this on `pull_request` but only running it when either `metomi/rose/loc_handlers/rsync_remote_check.py` or this test file are modified.

t/test_rsync_loc_handlers.sh
```diff
@@ -0,0 +1,32 @@
...
+
+python metomi/rose/loc_handlers/rsync_remote_check.py $PWD/foo 'blob' 'tree' > out.txt
+test="it prints tree"
+grep tree out.txt > /dev/null && echo "PASS ${test}" || exit 1
+
+test="it prints foo/baz/qux"
+grep "foo/baz/qux" out.txt > /dev/null && echo "PASS ${test}" || exit 1
+
+test="it prints 4 lines"
+test "$(cat out.txt | wc -l)" == 4 && echo "PASS ${test}" || exit 1
```
oliver-sanders, line 21: How about performing a known-good-output test here rather than grepping for individual words. I think the output of this command should be stable (possibly access mode aside although that can be controlled when the files are created at the start of the test).

t/test_rsync_loc_handlers.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
oliver-sanders, line 1: Missing copyright header.

t/test_rsync_loc_handlers.sh
```diff
@@ -0,0 +1,32 @@
...
+
+python metomi/rose/loc_handlers/rsync_remote_check.py $PWD/foo/bar 'blob' 'tree' > out.txt
+
+test="it prints blob"
+grep blob out.txt > /dev/null && echo "PASS ${test}" || exit 1
+
+test="it prints foo/bar"
+grep "foo/bar" out.txt > /dev/null && echo "PASS ${test}" || exit 1
+
+test="it prints 2 lines"
```
oliver-sanders, line 31: This might be better implemented with `prove` as it has better reporting of errors etc.

.github/workflows/test_rsync_test_script.yml
```diff
@@ -0,0 +1,47 @@
+name: test rsync check
```
oliver-sanders, line 1: There's an empty `github/workflows/test_rsync_test_script.py` file here?

wxtim:
Will address comments from Oliver - focusing on rose release

t/test_rsync_loc_handlers.sh
wxtim: Yes

.github/workflows/test_rsync_test_script.yml
```diff
@@ -0,0 +1,47 @@
...
+
+on:
+  pull_request:
+    paths:
+      - 'metomi/rose/loc_handlers/rsync.py'
+      - 'metomi/rose/loc_handlers/rsync_remote_check.py'
```
oliver-sanders, line 7: I think we might need to list the action itself?

```suggestion
      - 'metomi/rose/loc_handlers/rsync_remote_check.py'
      - '.github/workflows/test_rsync_test_script.yml'
```

.github/workflows/test_rsync_test_script.yml
```diff
@@ -0,0 +1,47 @@
...
+  workflow_dispatch:
+    inputs:
+      rose_ref:
+        description: The Rose branch to test against
+        required: true
+
+defaults:
+  run:
+    shell: bash  # macos default shell is zsh
+
```
oliver-sanders, line 17: Not using Mac OS (no need to for this script):

```suggestion
```

t/rose-loc-handlers/test_rsync_loc_handlers.t
```diff
@@ -0,0 +1,57 @@
...
+SCRIPT="${ROSE_TEST_HOME}/rose/metomi/rose/loc_handlers/rsync_remote_check.py"
+
+# Run the script on a tree:
+TEST_KEY="${TEST_KEY_BASE} tree"
+run_pass "${TEST_KEY}" python "${SCRIPT}" $PWD/foo 'blob' 'tree'
+awk '{print $1, $3, $4}' > cut_test_file < "${TEST_KEY}.out"
+file_cmp "${TEST_KEY}-vs-kgo" "cut_test_file" <<__HERE__
+tree  
+- - $PWD/foo/baz
+33188 3 $PWD/foo/bar
```
oliver-sanders, line 46: What's 33188, is this the access mode?

If so it might be worth running `chmod` (or using a mask) on the HI THERE files to make this test more robust (as the default file permissions are an environment thing).

oliver-sanders (commented):
:1+:

t/rose-loc-handlers/test_rsync_loc_handlers.t
```diff
@@ -0,0 +1,59 @@
...
+
+# Print the Python version to stderr, just to check.
+python --version >&2
+
+# Create a folder containing both files and directories.
+mkdir -p foo/baz
+echo "HI" > foo/bar
+echo "THERE" > foo/baz/qux
+chmod 752 foo/baz/qux
+chmod +x foo/bar
```
oliver-sanders, line 36: You might want to use an absolute value here too (e.g. `755`) as `+x` just adds user execute (i.e. makes the first digit a `7`) but doesn't change the other fields.

oliver-sanders:
(docs failure unrelated)

oliver-sanders (approved):

PR #2618, opened 2022-09-28
2.0.x patch

oliver-sanders:


MetRonnie:
Commit history is a bit odd here, this can actually be rebased onto 2.0.x and it's only the latest commit that is different.

But do need to resolve changelog conflict

MetRonnie (approved):

PR #2632, opened 2022-10-17
GH Actions: fix order of pip installation

MetRonnie:
This was failing for master because when installing cylc-rose before rose, it was complaining it couldn't find rose 2.1

I've also added a missing changelog entry

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,15 @@ updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 --------------------------------------------------------------------------------
 
+## 2.0.2 (<span actions:bind='release-date'>Upcoming</span>)
+
+### Fixes
+
+[#2608](https://github.com/metomi/rose/pull/2608) - Fix problems with the
+tutorials.
+
+--------------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -10,6 +10,15 @@ updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 --------------------------------------------------------------------------------
 
+## 2.0.2 (<span actions:bind='release-date'>Upcoming</span>)
+
+### Fixes
+
+[#2608](https://github.com/metomi/rose/pull/2608) - Fix problems with the
```
wxtim, line 17: Should this be here?

MetRonnie: Yes, I forgot to add a changelog entrey in that PR

wxtim (approved):

oliver-sanders:
> This was failing for master because when installing cylc-rose before rose, it was complaining it couldn't find rose 2.1

~I think this was a problem with the cylc-rose metadata, not the metomi-rose GH action?~

No, because we haven't installed the local version yet :facepalm: 

We should really bundle the install of the chosen package along with its dependencies in install-cylc-components.

Something like this: https://github.com/cylc/release-actions/pull/158

Issue #2647, opened 2022-11-22
edit: capabilities document


oliver-sanders:
Extract existing `cylc rose` capabilities into a document for future reference.

Suggest the following sections:

* Configuration
* Metadata
* Search & Filter
* Widgets
* Misc

In due course we need to outline how these capabilities will map onto the new framework.

oliver-sanders:
(assigning myself for the metadata section)

Issue #2644, opened 2022-11-22
edit: choose a form generator


oliver-sanders:
We need to auto-generate forms from schemas. Whilst we may need to implement some components ourselves (e.g. STASH) there is reasonable potential for us to use an existing form generator.

Here's a list of JSON Schema form generators: https://json-schema.org/implementations.html#web-ui-generation

Considerations:
* Rose configurations are hierarchical, we need a form generator which will support this or which could be extended to support this.
* We need the ability to add custom widgets (e.g. STASH).
* We would like to implement advanced searching / filtering controls.
* We would like slick UI components!

oliver-sanders:
JupyterLab has started to use React JsonSchema Form (search for rjsf in Jupyter Lab issues/PRs).

https://github.com/rjsf-team/react-jsonschema-form

Issue #2643, opened 2022-11-22
edit: decide on a schema


oliver-sanders:
We are currently considering JSON Schema, however, we should have a quick look around to make sure this is the best choice.

Points in favour of JSON Schema:
* It is used by Jedi.
* There is good tooling around it.
* It may be easier to integrate with text editors via existing JSON Schema plugins.
* JSON Schema Form extends JSON Schema into form layout/control.

aenglyst:
The current version is 2020-12 

Specification links can be found here:  

[Specification Links | JSON Schema (json-schema.org)](https://json-schema.org/specification-links.html)

aenglyst:
**Pros** 

 


- **Standardised format** - JSON Schema is an IETF standard providing a format for what JSON data is required for a given application and how to interact with it. Applying such standards for a JSON document lets you enforce consistency and data validity across similar JSON data. 

-  **Future Proofed** - The JSON Schema format is a web standard meaning changes in Front End frameworks are decoupled from the data layer far downstream (Close to the clientside delivery) 
 
- **Validation** of both the front and backend can be done using the same schema. It is a powerful tool for validating the structure of JSON data 
 
- **Yaml compatible** 
 
- **Reusable** - Already used at MO. It is used by Jedi. Creating synergy between architectures used in different areas reduces maintenance and improves scalability. This will help to future-proof the technology stack.  
 
- **API compatibility** - The JSON Schema media types (application/schema+json and application/schema-instance+json) will be published as an IETF RFC, which has already been adopted by the HTTP APIs working group. 
 
- **Web validation tools** - https://json-schema.org/implementations.html#validator-web%20(online) 
 
- **Text editor integration** - It may be easier to integrate with text editors via existing JSON Schema plugins. 
 Further investigation is required. In particular for the use of LSPs for text editor integration. This is being looked at in tickets #2660, #2651. 
 
- **UI form integration** - JSON Schema Form extends JSON Schema into form layout/control. 
 There are a number of JSON Schema form extensions available. 
 A proof of concept application demonstrates the use of one of these: JSON Schema Form. 
 
    This provides a neat solution for end-to-end data editing and validation. Some more detailed investigation is required for addressing custom widgets and hierarchical data delivery. 
    Further investigation is provide in #2644. 
 
- **Good tooling is available** - Generally there is a lot of resources available for implementation. See:  
 
    [Implementations | JSON Schema (json-schema.org)](https://json-schema.org/implementations.html) 
    
    
     
    
[...]

PR #2651, opened 2022-11-28
rose edit capabilities document

oliver-sanders:
Partially addresses #2647 

Adds configuration and metadata sections to get this document started.

There are a couple of TODOs where further research is required.

### DIFF
--- a/etc/rose-edit-capabilities.md
+++ b/etc/rose-edit-capabilities.md
@@ -228,9 +228,9 @@ be filtered via the view menu.
 ||||||
 | Menu options|||||
 | File >|||||
-| Open| Open configuration <br/> `@cylc_ls.feature(TEXT_DOCUMENT_DID_OPEN)` ||||
+| Open| Open configuration |Pygls can register a function to this Language Server Protocol (LSP) feature using a decorator and a type from @microsoft/lsprotocol `@my_ls.feature(lsprotocol.types.TEXT_DOCUMENT_DID_OPEN)`|||
 | Save| Save changes||||
-| Check And Save| Check and save changes| Check all metadata incl. fail-if/warn-if and validator macros, then save. |||
+| Check And Save| Check and save changes| Check all metadata incl. fail-if/warn-if and validator macros, then save.<br/> LSP: `TEXT_DOCUMENT_DID_SAVE` |||
 | Load All Apps| Load All Apps| Option is enabled for rose-suite.conf but errors when selecting this option. Option is disabled for rose-app.conf|||
 | Quit| Quits rose edit||||
 | Edit >|||||


aenglyst (commented):
A good summary of Configuration and metadata descriptions.

Perhaps as we get a little further along we can use the descriptions from the rose metadata items to create an end-to-end mapping scheme, i.e to map fields all the way from rose metadata > jsonSchema > jsonForms. This has already been mapped for a number of metadata items, however, in the longer term it may be useful to get a complete mapping so all gaps can be identified. Where the requirement for this in the UI is essential, alternative ways to get this info into the UI can then be factored into functional specifications. 

If the lack of a direct mapping is an issue for a number of fields, this should be factored into the research to be undertaken in #2643 for choice of schema and #2644 for choice of form generator.

Perhaps it is worth looking into whether there are any custom mapping solutions available to help where there is no direct mapping from metadata onto json-schema / json-form. For example this would apply to comment support, after assessing whether or not this feature is actually a requirement for Rose II.

This would apply to any other schemas chosen over these that come out of the technology choices investigations.

oliver-sanders:
> create an end-to-end mapping scheme

That's what I'm trying to do listing the corresponding fields with each metadata item. There's probably a better way to represent this...

> it may be useful to get a complete mapping so all gaps can be identified

That's the aim, just got a few gaps to fill.

> Perhaps it is worth looking into whether there are any custom mapping solutions available

I think the remaining things which have no direct mapping onto JSON schema will fall into one of two camps:

1. Things we can template into the description or work around in the metadata somehow.
   * E.G. rather than handing this in the form directly we could handle it in a metadata component associated with the form input (e.g. via an info button).
2. Things we must handle async server-side.
   * If there are ways to extend JSON schema, great! However, some things such as `fail-if`/`warn-if` & macros require Python logic to evaluate so could not feasibly be implemented client side.

oliver-sanders:
Looks like `$vocabulary` can be used to extend the schema, this might be useful for representing Rose extensions:

http://json-schema.org/draft/2020-12/json-schema-core.html#name-example-meta-schema-with-vo

aenglyst:
That is good to see most of the mappings can be handled directly and the two approaches for where there is no direct mapping provide viable solutions, either client side or server side, which brings it closer to a complete end to end solution.

The use of $vocabulary is interesting and looks useful. This may require further investigation.

etc/rose-edit-capabilities.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Search & Filter
+
+Rose Edit has a global-search bar which uses input names and allows inputs to
+be filtered via the view menu.
+
+> **TODO:**
+> Explore the view options and search features and record the key capabilities.
+
+
```
aenglyst, line 226: [Rose Edit Capabilities.xlsx](https://github.com/metomi/rose/files/10195599/Rose.Edit.Capabilities.xlsx)

aenglyst (approved):
This all looks fine. 
Some comments added and further points for discussion.

aenglyst (approved):

oliver-sanders:
LGTM.

Answered a couple of questions and filled in some TODOs in https://github.com/metomi/rose/pull/2651/commits/8960c9f0da485bcf25c39a81168b2288cd655503

aenglyst (approved):
I've added some notes on exploring the widgets and associated capabilities

oliver-sanders:
Assigning @wxtim who is going to try to map functionalities onto Language Server Protocol (LSP) interfaces.

A couple of questions to be resolved:

* [ ] Can we register commands on a per-file basis with LSP?
   LSP has some functionality for registering commands which looks like it will be a good fit for Rose macros. However, the list of Rose macros available depends on the file being edited (because different apps have different macros).  So we need to be able to vary the response based on the file being edited.
* [ ] Can LSP validate files incrementally?
   Can we validate the file diff by diff rather than validating the whole file top to bottom each time something changes?

And a couple more questions about the `pygls` LSP implementation (which I think is the only generic LSP implementation in Python):

* [ ] Does `pygls` synchronise the file being edited with the client?
   LSP has an interfaces for synchronising the file being edited between the client and the server. Can we get at this internal representation of the file from `pygls` or can we only see the file as it exists on the filesystem?
* [ ] Does `pygls` support all of the interfaces we require?

wxtim:
## Incremental Changes

> Does pygls synchronise the file being edited with the client?
> LSP has an interfaces for synchronising the file being edited between the client and the server. Can we get at this internal representation of the file from pygls or can we only see the file as it exists on the filesystem?

> Does pygls synchronise the file being edited with the client?
> LSP has an interfaces for synchronising the file being edited between the client and the server. Can we get at this internal representation of the file from pygls or can we only see the file as it exists on the filesystem?

As far as I can tell we can see the diff in the case of `TEXT_DOCUMENT_DID_CHANGE` event as one of the arguments the decorator function provides as `params`:

```
DidChangeTextDocumentParams(
    text_document=VersionedTextDocumentIdentifier(
        version=5, uri='file:///net/home/h02/tpilling/cylc-src/x9/flow.cylc'
    ),
    content_changes=[TextDocumentContentChangeEvent_Type1(range=6:4-6:4, text='Hello World', range_length=0)]
)
```
It looks like the server holds a `workspace` object which contains representations of `Document` objects which contains a full representation of the doc - i.e.:

```python
>>> ls.workspace.documents
{'file:///net/home/h02.../flow.cylc': <pygls.workspace.Doc...54201ce20>}

>>> list(ls.workspace.documents.values())[0].lines
['#!jinja2\n', '\n', '\n', '\n', '[scheduling]\n', '\tinitial cycle point = 1066\n', '    Hello World    [[graph]]\n', '        R1 = task\n', '[runtime] \n', '[meta]']
list(ls.workspace.documents.values())[0].uri
'file:///net/home/h02/tpilling/cylc-src/x9/flow.cylc'
```

Futher, it looks like the `TEXT_DOCUMENT_DID_CHANGE` method itself uses an incremental update to the internal representation

wxtim:
Recording this link - I'm not sure it's relevent, but I want to record it https://github.com/microsoft/language-server-protocol/issues/642

PR #2667, opened 2023-02-02
GH Actions: update libs to install

MetRonnie:

PR #2670, opened 2023-02-08
Macro validation fix

### LINKED ISSUE #2662: macro: invalid trigger expression: this ...
Spotted in the wild, a `trigger` expression behaving differently with Rose 2019 and 2.

```ini
# rose-app.conf
[env]    
FOO=0                                                                                                                              
BAR=2  
```

```ini
# meta/rose-meta.conf
[env=FOO]                                                                                                                          
 
[env=BAR]    
trigger=env=FOO: this < 1
```

```console
$ CYLC_VERSION=7 rose macro -V
[V] rose.macros.DefaultValidators: issues: 1
    env=FOO=0
        State should be trig-ignored
$ CYLC_VERSION=8 rose macro -V
[V] metomi.rose.macros.DefaultValidators: issues: 1
    env=BAR=2
        Invalid trigger expression: this < 1
```

wxtim:
Closes #2662 .

Following the instructions in the issue the following error is seen:

```
Invalid trigger expression: this < 1
```
The underlying cause is at `metomi/rose/macros/rule.py:214` is that type comparisons are not the same at Python 3 for base types as they are at Python 2.

```
(Pdb) template.render(rule_id_values)
*** TypeError: '<' not supported between instances of 'str' and 'int'
```

@oliver-sanders  suggested the following workarounds (in order of preference):

> Order of preference:
> 1a. Try to overwrite the str. float, bool & int, __lt__, __gt__, __ge__, __le__ methods used by Jinja2. 
> 1b. You may be able to do this by subclassing str, float, int etc as I started.
>       You may be able to do this by redefining them in the Python environment before calling Jinja2.
>
> 2. try/except cascade 
> try/except(TypeError)If this == '', then this=-999999999999999999999999999999try/except(TypeError)this = Falsetry/except(TypeError)raise
>
> 3. Cry

This PR uses approach 1a. 😃

### DIFF
--- a/.github/actions/install-libs/action.yml
+++ b/.github/actions/install-libs/action.yml
@@ -28,8 +28,8 @@ runs:
 
         # NOTE: hard-coded branches need to be manually updated as necessary.
         if [[ "$BASE_REF" == '2.0.x' ]]; then
-          : ${CYLC_FLOW_REF:='8.0.x'}
-          : ${CYLC_ROSE_REF:='1.1.x'}
+          : ${CYLC_FLOW_REF:='8.1.x'}
+          : ${CYLC_ROSE_REF:='master'}
         fi
 
         if [[ -n "$CYLC_FLOW_REF" ]]; then
--- a/.github/actions/install-libs/action.yml
+++ b/.github/actions/install-libs/action.yml
@@ -27,10 +27,10 @@ runs:
         BASE_REF="${GITHUB_BASE_REF:-$GITHUB_REF_NAME}"
 
         # NOTE: hard-coded branches need to be manually updated as necessary.
-        if [[ "$BASE_REF" == '2.0.x' ]]; then
-          : ${CYLC_FLOW_REF:='8.1.x'}
-          : ${CYLC_ROSE_REF:='master'}
-        fi
+        # if [[ "$BASE_REF" == '2.0.x' ]]; then
+        : ${CYLC_FLOW_REF:='8.1.x'}
+        : ${CYLC_ROSE_REF:='master'}
+        # fi
 
         if [[ -n "$CYLC_FLOW_REF" ]]; then
           CYLC_FLOW_GITHUB="${CYLC_FLOW_GITHUB}@${CYLC_FLOW_REF}"
--- a/metomi/rose/macros/rule.py
+++ b/metomi/rose/macros/rule.py
@@ -68,6 +68,8 @@ class MyInt(int):
             False
             >>> MyInt(42) < 42
             False
+            >>> MyInt(77) < ''
+            False
         """
         if isinstance(other, (int, float, MyFloat, MyInt)):
             return float(self) < float(other)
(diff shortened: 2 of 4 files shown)


.github/actions/install-libs/action.yml
```diff
@@ -28,8 +28,8 @@ runs:
 
         # NOTE: hard-coded branches need to be manually updated as necessary.
         if [[ "$BASE_REF" == '2.0.x' ]]; then
-          : ${CYLC_FLOW_REF:='8.0.x'}
-          : ${CYLC_ROSE_REF:='1.1.x'}
+          : ${CYLC_FLOW_REF:='8.1.x'}
```
wxtim, line 31: Should have been updated on release.

metomi/rose/macros/rule.py
```diff
@@ -50,6 +50,180 @@
...
+    """
+    def __lt__(self, other):
+        """
+        Examples:
+            >>> MyInt(4) < MyInt(6)
+            True
+            >>> MyInt(4) < MyFloat(6.1)
+            True
+            >>> MyInt(4) < MyStr('Zaphod Beeblebrox')
+            True
```
oliver-sanders, line 64: May be worth trying an empty string example.

wxtim:
Failing on the same 6 tests as master https://github.com/metomi/rose/actions/runs/4122813340/jobs/7120050963

.github/actions/install-libs/action.yml
oliver-sanders: See #2667

oliver-sanders: Now merged to 2.0.x.

metomi/rose/macros/rule.py
```diff
@@ -50,6 +50,198 @@
...
+            False
+        """
+        if isinstance(other, (int, float, MyFloat, MyInt)):
+            return float(self) < float(other)
+        elif isinstance(other, MyStr):
+            return True
+        else:
+            return int(self) == other
+
+    def __eq__(self, other):
```
oliver-sanders, line 81: Overwriting the builtin `__eq__`!

metomi/rose/macros/rule.py
```diff
@@ -50,6 +50,198 @@
...
+            False
+            >>> MyInt(-77) == 'Viltvodle VI'
+            False
+        """
+        if isinstance(other, (int, float, MyFloat, MyInt)):
+            return float(self) == float(other)
+        elif isinstance(other, MyStr):
+            return False
+        else:
+            return int(self) == other
```
oliver-sanders, line 102: Another option here might be:

```python
try:
    return self == other
except TypeError:
    return False
```

I think this would be more robust and avoid casting to `float` which could alter results?

Ditto for the others?

metomi/rose/macros/rule.py
```diff
@@ -50,6 +50,198 @@
...
+            False
+        """
+        if isinstance(other, (int, float, MyFloat, MyInt)):
+            return float(self) < float(other)
+        elif isinstance(other, MyStr):
+            return True
+        else:
+            return float(self) < other
+
+    def __eq__(self, other):
```
oliver-sanders, line 151: Overriding the builtin `__eq__`!

oliver-sanders:
I have run this through the reported example and everything checks out, phew, thanks @wxtim.

One thing I did notice is that the `MyStr` class name can appear in user-facing output which I hadn't expected.

Suggest renaming `MyStr` to `Str`, etc. This way the error messages will appear almost identical to the user, just with the first letter capitalised.

wxtim:
> I have run this through the reported example and everything checks out, phew, thanks @wxtim.

Thank you for doing this check.

> One thing I did notice is that the `MyStr` class name can appear in user-facing output which I hadn't expected.
> 
> Suggest renaming `MyStr` to `Str`, etc. This way the error messages will appear almost identical to the user, just with the first letter capitalised.

I have now done this

datamel (approved):
I have verified the problem on master and compared to the behaviour on Rose 2019. I have manually tested the change with the example in the issue, run the tests and read the code. No problems spotted form me

oliver-sanders (approved):

PR #2673, opened 2023-02-10
Disable failing tests until sqlalchemy can be upgraded

datamel:
Disable tests until sqlalchemy version can be upgraded.

Diagnosis from @oliver-sanders:
 - sqlalchemy has released version 2 which has some breaking changes.
 - to help people migrate they added some deprecation messages to v1.4
 - these messages break tests which check that nothing is written to stderr.

Note various other attempts at suppressing the warnings failed, including setting env variables and using python warning suppression.

Built on https://github.com/metomi/rose/pull/2667

### DIFF
--- a/t/rosa-svn-post-commit/00-basic.t
+++ b/t/rosa-svn-post-commit/00-basic.t
@@ -25,6 +25,7 @@
 if ! python3 -c 'import sqlalchemy' 2>/dev/null; then
     skip_all '"sqlalchemy" not installed'
 fi
+skip_all "sqlalchemy upgrade required"
 tests 71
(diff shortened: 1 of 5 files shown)


t/rosa-svn-post-commit/00-basic.t
```diff
@@ -25,6 +25,7 @@
 if ! python3 -c 'import sqlalchemy' 2>/dev/null; then
     skip_all '"sqlalchemy" not installed'
 fi
+skip_all "sqlalchemy upgrade required"
```
oliver-sanders, line 28: Could we just skip the broken stderr tests.

PR #2682, opened 2023-03-24
job runner: limit concurrency

oliver-sanders:
* Restore the Rose 2019 concurrency limit of 6.
* Closes #2680

Had to jig the code around a bit to move away from the maximum concurrency model.

For review, please ensure:

* It limits concurrency.
* It still handles errors correctly.
* It runs all the jobs (I think some jobs can be inter-dependent?).

TODO: Need to figure out which branch to put this on => Monday problem.

PR #2684, opened 2023-03-27
job runner: limit concurrency

### LINKED ISSUE #2680: fileinstall: restore parallism limiting
We used to manage file installation processes (e.g. `rsync`s) in a worker pool. We switched to an asyncio implementation, however, the limit was lost in the process:

https://github.com/metomi/rose/pull/2325/files#diff-1a4d9963cf73c867e4dcfdce40251652a0674c8220771f24138b54ef9d4eaf3cL152

We should restore this parallel limit to reduce the risk of issues which could result from it.

oliver-sanders:
* Restore the Rose 2019 concurrency limit of 6.
* Closes #2680

This is a cherry-pick of https://github.com/metomi/rose/pull/2682 from master, onto the 2.0.x branch.

### DIFF
--- a/metomi/rose/job_runner.py
+++ b/metomi/rose/job_runner.py
@@ -148,6 +148,8 @@ class JobProxy:
 class JobRunner:
     """Runs JobProxy objects with pool of workers."""
 
+    ASYNC_SLEEP_TIME = 0.1
+
     def __init__(self, job_processor, nproc=None):
         """
         Initialise a job runner.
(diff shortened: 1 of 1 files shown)


oliver-sanders:
A simple test would be to write a Rose configuration file which installs a bunch of resources e.g:

```ini
[command]
    default=true

[file:foo]
    source=~/temptest/foo

[file:bar]
    source=~/temptest/bar

...
```

Then run it something like:

```
$ cd $(mktemp -d)
$ rose app-run -C /path/to/app
```

You could jam a "job started" and "job stopped" message into the code to confirm that the concurrency is respected and test that the files requested were installed into `$PWD`.

MetRonnie (approved):
Ok I've had a play with it and no more than 6 asyncio tasks are created at any time

PR #2693, opened 2023-04-27
actions: test with python up to 3.11

oliver-sanders:


metomi/rose/__init__.py
```diff
@@ -17,7 +17,6 @@
 """This package contains the Python code for Rose utilities.
 
 This module contains the constants that are used globally within these.
-
```
oliver-sanders, line 20: Whitespace change to force the tests to re-run.

PR #2699, opened 2023-05-17
Bunch: Incremental mode tweaks

### LINKED ISSUE #2698: Bunch: Not working incrementally on `cylc trigger`
### Synopsis

Cylc Triggering a previously run rose bunch task appears to run non-incrementally.

### How to replicate:

```bash
#!/bin/bash
# Complete issue replication:

WORKFLOW="example.rose-bunch"

mkdir -p "${WORKFLOW}/app/bunch"

# Create a workflow:
cat > "${WORKFLOW}/flow.cylc" <<__HERE__
[scheduling]
    initial cycle point = 2069
    [[graph]]
        R1 = bunch

[runtime]
    [[bunch]]
        script = rose task-run

__HERE__

# An empty rose-suite.conf:
touch "${WORKFLOW}/rose-suite.conf"

# This rose-app.conf:
cat > "${WORKFLOW}/app/bunch/rose-app.conf" <<__HERE__
mode=rose_bunch
meta=rose_bunch

[bunch-args]
arg1=true false true

[bunch]
command-format=%(arg1)s
incremental=true
fail-mode=continue

__HERE__

# install and play the workflow
cylc vip ./"${WORKFLOW}" --host=localhost

# Wait for first job to finish
while true; do
    if grep TOTAL "${HOME}/cylc-run/${WORKFLOW}/runN/log/job/20690101T0000Z/bunch/01/job.out"; then
        break
    else
        sleep 5
    fi
done

cylc trigger "${WORKFLOW}//*/*"

# Wait for second attempt to finish
while true; do
    if grep TOTAL "${HOME}/cylc-run/${WORKFLOW}/runN/log/job/20690101T0000Z/bunch/02/job.out"; then
        break
    else
        sleep 5
    fi
done

# This output should show that only one job ran second time around, but in reality all three ran.
cat "${HOME}/cylc-run/${WORKFLOW}/runN/log/job/20690101T0000Z/02/job.out"
```

### Investigation

In https://github.com/metomi/rose/pull/1865 a check for `CYLC_TASK_SUBMIT_NUMBER` was added to prevent new runs picking up old rose_bunch database files and assuming that they could run a rose bunch task in incremental mode.

It was changed to `CYLC_TASK_TRY_NUMBER` at https://github.com/metomi/rose/commit/7757cf02df04bee1152b9b8737cef4f06eac1571, although the reason for this is unclear - @oliver-sanders  can you interpret your own work?

The original reason in #1865 is now irrelevant, because Cylc install should delete the bunch database. *_BUT_* reflows may cause a similar problem, so we probably need to retain this code, and return it to `CYLC_TASK_SUBMIT_NUMBER`

Ideally we should probably run in incremental mode only if this is the first run of this task in a given flow, but at present there is no way of working out whether this is the first time that we have run  task in a given flow, the database does not contain information about the environment of the previous run (apart from the PATH variable).

### Actions:

Pending further discussion on possibly storing flow numbers in the database, we should:
- Put the `CYLC_TASK_SUBMIT_NUMBER` back.
- Turn off incremental mode if flow is not 1?
- Document that in a new flow all bunch tasks will run - this is a fail-safe way of working.

wxtim:
Closes #2698 

Two related issues:
- [x] Previously rose bunch is was using CYLC_TASK_TRY_NUMBER, causing it to treat each submit as the first run of the job and turning off incremental running - this needed changing to CYLC_TASK_SUBMIT_NUMBER.
- [x] rose bunch needs awareness of flow numbers. A new submit should not run in incremental mode if the task belongs to a different flow to the previous run -- full details of investigation in #2698 .
- [x] Add documentation about how incremental mode works.

### DIFF
--- a/metomi/rose/apps/rose_bunch.py
+++ b/metomi/rose/apps/rose_bunch.py
@@ -291,6 +291,10 @@ class RoseBunchApp(BuiltinApp):
         # Set max number of processes to run at once
         max_procs = conf_tree.node.get_value([self.BUNCH_SECTION, "pool-size"])
 
+        # Add CYLC_TASK_FLOW_NUMBERS to configuration:
+        cylc_task_flow_numbers = os.environ.get('CYLC_TASK_FLOW_NUMBERS', "0")
+        conf_tree.node.set(['CYLC_TASK_FLOW_NUMBERS'], cylc_task_flow_numbers)
+
         if max_procs:
             max_procs = int(metomi.rose.env.env_var_process(max_procs))
         else:
@@ -475,7 +479,7 @@ class RoseBunchDAO:
     def create_tables(self):
         """Create tables as appropriate"""
         existing = []
-        first_run = os.environ.get("CYLC_TASK_TRY_NUMBER") == "1"
+        first_run = os.environ.get("CYLC_TASK_SUBMIT_NUMBER") == "1"
 
         for row in self.conn.execute(
             "SELECT name FROM sqlite_master " + "WHERE type=='table'"
@@ -566,7 +570,6 @@ class RoseBunchDAO:
 
     def record_config(self, config, clear_db=False):
         """Take in a conf_tree object and record the entries"""
-
         if clear_db:
             d_stmt = "DELETE FROM " + self.TABLE_CONFIG
             self.conn.execute(d_stmt)
--- a/sphinx/api/built-in/rose_bunch.rst
+++ b/sphinx/api/built-in/rose_bunch.rst
@@ -12,12 +12,12 @@ its output directed to separate ``.out`` and ``.err`` files of the form:
 .. code-block:: sub
 
    bunch.<name>.out
-   
+
 Should you need separate working directories you should configure your
 command to create the appropriate subdirectory for working in.
 
 .. note::
-   
+
    Under load balancing systems such as PBS or Slurm, you will need to
    set resource requests to reflect the resources required by running
    multiple commands at once e.g. if one command would require 1GB
@@ -91,7 +91,7 @@ sections of the :rose:file:`rose-app.conf` file, but
       .. rose:conf:: fail-mode=continue|abort
 
          :default: continue
-         
+
          Specify what action you want the job to take on the failure of a
          command that it is trying to run. If set to continue all command
          variants will be run by the job and the job will return a non-zero
@@ -111,7 +111,7 @@ sections of the :rose:file:`rose-app.conf` file, but
       .. rose:conf:: incremental=true|false
 
          :default: true
-         
+
          If set to ``true`` then only failed commands will be re-run on
          retrying running of the job. If any changes are made to the
          configuration being run then all variants will be re-run. Similarly,
@@ -120,6 +120,10 @@ sections of the :rose:file:`rose-app.conf` file, but
          will result in all commands being run. In verbose mode the app
          will report commands that won't be run due to previous successes
          in the job output with a ``[PASS]`` prefix.
+
+         .. seealso::
+
+            :ref:`rosebunch.CylcTasks`
 
       .. rose:conf:: names=name1 name2 ...
 
@@ -167,3 +171,14 @@ sections of the :rose:file:`rose-app.conf` file, but
          ``command-instances`` and ``COMMAND_INSTANCES``, which are reserved
          for the auto-generated list of instances when the
          :rose:conf:`[bunch]command-instances=N` option is used.
+
+.. _roseBunch.CylcTasks:
+
+Incremental Mode In Cylc Tasks
+------------------------------
+
+When incremental mode is turned on, only failed commands will be re-run
+if the task is run again (e.g. by manual triggering, or automatic retries).
+
+If the task is run again as part of a new flow (e.g. ``--flow=new``),
+then all commands will be re-run.
--- a/t/rose-task-run/31-app-bunch.t
+++ b/t/rose-task-run/31-app-bunch.t
@@ -22,7 +22,7 @@
 . $(dirname $0)/test_header
 
 #-------------------------------------------------------------------------------
-tests 75
+tests 78
 #-------------------------------------------------------------------------------
 # Define some constant patterns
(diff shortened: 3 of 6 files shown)


t/rose-task-run/43-bunch-check-flows.t
```diff
@@ -0,0 +1,78 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Rose. If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test "rose_prune" built-in application:
+# Prune items in job hosts sharing a common file system (but not with the suite
+# host).
```
wxtim, line 22:
```suggestion
Test "rose_bunch" built-in application:
Test that CYLC_FLOW_TASK_NUMBERS is stored in the rose_bunch database
and incremental mode turned off if changes made to its value in a later run.
```

sphinx/api/built-in/rose_bunch.rst
```diff
@@ -31,9 +31,7 @@ Example
 .. code-block:: rose
 
    meta=rose_bunch
-   mode=rose_bunch
-
-   [bunch]
+   mode=rose_bunchincre
```
wxtim, line 34:
```suggestion
   mode=rose_bunch

   [bunch]
```

sphinx/api/built-in/rose_bunch.rst
```diff
@@ -167,3 +171,18 @@ sections of the :rose:file:`rose-app.conf` file, but
...
+
+Rose Bunch examines environment variables set in the
+:ref:`Cylc job script variables<Task Job Script Variables>` to produce
+sensible behavior in Rose Bunch tasks if they are re-run by Cylc:
+
+- If a Cylc task is re-submitted as part of a new flow *all* jobs will be run.
+
+- If a Cylc task is re-submitted in the same flow as it previously ran
+  incremental mode will be switched on by default; so jobs which have
+  previously succeeded will be skipped.
```
oliver-sanders, line 188: This is a little unclear to me, suggest something along the lines of:

```rest
Incremental Mode In Cylc Tasks
---

When incremental mode is turned on, only failed commands will be re-run if the task is run again (e.g. by manual triggering, or automatic retries).

If the task is run again as part of a new flow (e.g. `--flow=new`), then all commands will be re-run.
```

t/rose-task-run/43-bunch-check-flows.t
```diff
@@ -0,0 +1,78 @@
...
+
+[bunch]
+command-format=%(arg1)s
+incremental=true
+fail-mode=continue
+__HERE__
+
+
+# App outside Cylc task
+unset CYLC_FLOW_TASK_NUMBERS
```
oliver-sanders, line 57: Wrong env var?

oliver-sanders (commented):
LGTM, could do with testing the flow interaction.

sphinx/api/built-in/rose_bunch.rst
wxtim: Happy with suggestion, taken it.

t/rose-task-run/31-app-bunch.t
```diff
@@ -111,6 +116,19 @@ file_grep $TEST_KEY_PREFIX-ran-2 \
...
+
+# Ensure that the re-triggered task is done:
+<<<<<<< HEAD
+while true; do
+    if grep "TOTAL" "$FILE" 2>&1 /dev/null ; then
+        break
+    fi
+=======
+while ! grep "TOTAL" "$FILE" 2>&1 /dev/null; do
+>>>>>>> bab70fcbe1206a173b8f0f0cd7c824d707e4c27c
```
wxtim, line 128:
```suggestion
while ! grep "TOTAL" "$FILE" 2>&1 /dev/null; do
```

t/rose-task-run/44-bunch-incremental.t
```diff
@@ -0,0 +1,139 @@
...
+poll ! grep TOTAL "${APP_LOG_PATH}/01/job.out" 2>&1 /dev/null
+grep_incrementals "${APP_LOG_PATH}/01" 1 1 0
+
+# Re trigger task in same flow - bunch should skip previously succeeded tasks:
+ensure_trigger "2"
+poll ! grep TOTAL "${APP_LOG_PATH}/02/job.out" 2>&1 /dev/null
+grep_incrementals "${APP_LOG_PATH}/02" 0 1 1
+
+# Trigger the task a third time which doesn't trigger rose bunch
+# Allowing the N=1 window to move on.
```
wxtim, line 128: This seems to be demonstrably correct - should it be?

sphinx/api/built-in/rose_bunch.rst
oliver-sanders: (not resolved)

t/rose-task-run/44-bunch-incremental.t
oliver-sanders: By "which doesn't trigger rose bunch" do you mean that rose_bunch doesn't re-run and commands because they have all succeeded?

wxtim: No, in this version of the test the task used CYLC_FLOW_SUBMIT_NUMBER to decide whether to rose app-run or to `true`.

MetRonnie (approved):

metomi/rose/apps/rose_bunch.py
```diff
@@ -291,6 +291,10 @@ def run(self, app_runner, conf_tree, opts, args, uuid, work_files):
         # Set max number of processes to run at once
         max_procs = conf_tree.node.get_value([self.BUNCH_SECTION, "pool-size"])
 
+        # Add CYLC_TASK_FLOW_NUMBERS to configuration:
+        cylc_task_flow_numbers = os.environ.get('CYLC_TASK_FLOW_NUMBERS', "0")
+        conf_tree.node.set(['CYLC_TASK_FLOW_NUMBERS'], cylc_task_flow_numbers)
```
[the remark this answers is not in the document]
oliver-sanders: Yes, that shouldn't happen.

oliver-sanders (approved):

PR #2700, opened 2023-05-24
Fix Rosie password caching

### LINKED ISSUE #2661: rosie: investigate password caching
Investigate password caching e.g. with GPG agent w/ Rose 2 deployed via Conda environments.

Rosie uses password caching to allow it to interact with SVN repositories without users having to type their passwords every time they do so. Try out the latest version of Rosie with caching systems like gpg-agent to make sure it still works and add CI testing where possible.

MetRonnie:
Closes #2661 

- Replace `libsecret` with `keyring`
- Remove broken zenity dialogs
- Fix GPG Agent Store type mishandling leftover from Python 2->3 upgrade

#### Checklist
- [x] Dependencies updated
- [x] Changelog entry included

### DIFF
--- a/metomi/rosie/ws_client_auth.py
+++ b/metomi/rosie/ws_client_auth.py
@@ -107,7 +107,7 @@ class GPGAgentStore(BaseStore):
     RECV_BUFSIZE = 4096
 
     @classmethod
-    def usable(cls):
+    def usable(cls) -> bool:
         """Can this store be used?"""
         try:
             cls.get_socket()
@@ -116,7 +116,7 @@ class GPGAgentStore(BaseStore):
         return True
 
     @classmethod
-    def get_socket(cls):
+    def get_socket(cls) -> socket.socket:
         """Get a connected, ready-to-use socket for gpg-agent."""
         agent_info = os.environ.get("GPG_AGENT_INFO")
         if agent_info is None:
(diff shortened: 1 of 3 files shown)


dpmatthews:
Seems to be working OK for both gpg-agent and GNOME keyring

metomi/rosie/ws_client_auth.py
```diff
@@ -33,23 +34,13 @@
...
-        import gtk
-    import gnomekeyring
+    import keyring
+    KEYRING_FLAG = True
 except ImportError:
-    pass
+    KEYRING_FLAG = False
+else:
+    import keyring.errors
+    from keyring.backends.fail import Keyring as NoAvailableKeyring
```
wxtim, line 43: This is linguistically confusing - and looking at the class docstring makes me wonder why you'd want to import this object. I think a comment would be in order.

```suggestion
    # Fallback non-functional keyring: Used to check that the module.get_keyring()
    # has given us a keyring we can use.
    from keyring.backends.fail import Keyring as NoAvailableKeyring
```

wxtim (approved):
It all looks pretty sensible. I haven't worked out a way to functionally test it.

Might it be worth unit testing `GPGAgentStore.get_password`? It looks like it's doing some parsing, and I'd quite like to see examples of what we are expecting `replylines` to be.

MetRonnie:
> Might it be worth unit testing `GPGAgentStore.get_password`? It looks like it's doing some parsing, and I'd quite like to see examples of what we are expecting `replylines` to be.

We decided this system is on its way out so no effort will be spent on tests

wxtim:
I'm happy. Once Dave has signed off for his functional review, I'm happy to merge.

PR #2711, opened 2023-07-24
Remove release candidate warning from docs

wxtim:
Do we still want this?

Perhaps it's worth me reading this cover to cover at some point and jotting down anything else we might want rid of.

I've just noted a reference to file installation in the slides, for example. Will create a new issue later, but I think that this is required now.

### DIFF
--- a/sphinx/index.rst
+++ b/sphinx/index.rst
@@ -11,11 +11,6 @@
 Rose Documentation
 ==================
 
-.. warning::
-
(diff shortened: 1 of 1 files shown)


MetRonnie (approved):

MetRonnie:
Hang on, this warning isn't present in the deployed docs: https://metomi.github.io/rose/doc/html/index.html

Is this file actually being used?

PR #2712, opened 2023-07-24
2.1.0 docs

wxtim:
Might be worth waiting for #2711

See `git diff 2.0.4 2.1.0 sphinx` for what has actually changed in the docs - not much, but probably enough.
Only sig change is

```diff
--- a/sphinx/api/built-in/rose_bunch.rst
+++ b/sphinx/api/built-in/rose_bunch.rst
@@ -167,3 +171,14 @@ sections of the :rose:file:`rose-app.conf` file, but
          ``command-instances`` and ``COMMAND_INSTANCES``, which are reserved
          for the auto-generated list of instances when the
          :rose:conf:`[bunch]command-instances=N` option is used.
+
+.. _roseBunch.CylcTasks:
+
+Incremental Mode In Cylc Tasks
+------------------------------
+
+When incremental mode is turned on, only failed commands will be re-run
+if the task is run again (e.g. by manual triggering, or automatic retries).
+
+If the task is run again as part of a new flow (e.g. ``--flow=new``),
+then all commands will be re-run.
```

Other question
Is it worth either removing or replacing 2.0.0 with 2.0.4, or deleting it? It looks like there are a fair few changes between 2.0.0 and 2.0.4 - none super sig, but if we've made the edits?


MetRonnie (approved):

oliver-sanders:
We should leave the docs for the old version kicking around for a little while, makes sense to have, say the last three minor versions documented. There's no harm to us in keeping old versions documented. I tend to purge them every year or so.

oliver-sanders:
I think this might need a symlink change to make it the default?

oliver-sanders:
https://github.com/metomi/rose/blob/gh-pages/doc

wxtim:
Responded to your comments @oliver-sanders

MetRonnie (approved):

PR #2713, opened 2023-07-25
added a custom gh pages deploy action

wxtim:
Replace the default GH Action to create a custom workflow which doesn't try to run Jekyll. Once merged this will need to be enabled in the settings page.

### DIFF
--- a/.github/workflows/deploy_pages.yml
+++ b/.github/workflows/deploy_pages.yml
@@ -1,0 +1,45 @@
+# Simple workflow for deploying static content to GitHub Pages
+name: Deploy static content to Pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["gh-actions"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  # Single deploy job since we're just deploying
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: 'gh-pages'
+      - name: Setup Pages
+        uses: actions/configure-pages@v3
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v1
+        with:
+          # Upload entire repository
+          path: '.'
+      - name: Deploy to GitHub Pages
+        id: deployment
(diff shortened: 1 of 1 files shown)


MetRonnie (commented):
Strange that this has started failing here but not for cylc-doc. From https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#static-site-generators what happens if we just put a `.nojekyll` file in the root of the `gh-pages` branch?

wxtim:
`.nojekyll` also works, but @oliver-sanders seems enthusiastic about taking more control of what's happening here.

oliver-sanders:
> @oliver-sanders seems enthusiastic about taking more control of what's happening here.

Eh?

We did use the nojekyll approach once, I can't remember why we stopped, perhaps it just seemed nicer to whitelist the Sphinx directories and let Jekyll do its thing rather than hyde it away.

MetRonnie:
> rather than hyde it away.

I see what you did there 😀

wxtim:
FWIW I think `.nojekyll` does work with this repo, but I think we had to drop it for a good reason for at least one of the Cylc repos.

wxtim:
Closed because (IMO) `.nojekyll` is simpler. We might want to resurrect this laters.

wxtim:
We need this PR because the "doc" symlink breaks the default build. I cannot give a clear explanation for the failure, or replicate it locally.

- Default "pages build and deployment" workflow without "doc" symlink passes. The webpage doesn't work, of course - the redirect breaks.

I tried the tar command which is failing locally and it was fine. But there is one possible explanation on this GH post. The tarred file is only 189Mb which is much less than the 1Gb GH pages size limit.

MetRonnie:
I don't think that affects us?

> The majority of Pages sites **will not be affected by this change since [building with GitHub Actions is the default](https://github.blog/2022-08-10-github-pages-now-uses-actions-by-default/)**. If any of your Pages sites is affected, you will receive a build failure email with instructions how to fix the error on your next build.

wxtim:
> building with GitHub Actions is the default.

@MetRonnie  - I think you may be right - my thought process was that
- Build failed, apparently because of the symlinks.
- Therefore when it talks about using GH actions it meant explicitly setting it to GHA, not using the default GH Pages deploy action.

I'm still getting failures which can be prevented by removing symlinks. It looks like GH has decided that it doesn't like them much. Attempting to explicitly use the actions which GH provides for deployment of pages just gives me more explicit error message `Error: Artifact could not be deployed. Please ensure the content does not contain any hard links, symlinks and total size is less than 10GB.` ([Run](https://github.com/wxtim/rose/actions/runs/5950290413/job/16137757603))

MetRonnie:
> `Error: Artifact could not be deployed. Please ensure the content does not contain any hard links, symlinks and total size is less than 10GB.` ([Run](https://github.com/wxtim/rose/actions/runs/5950290413/job/16137757603))

What workflow is that? It should just be a built-in workflow called "pages-build-deployment" which doesn't have a problem with symlinks on cylc-doc's `gh-pages` branch

wxtim:
> What workflow is that? It should just be a built-in workflow called "pages-build-deployment" which doesn't have a problem with symlinks on cylc-doc's `gh-pages` branch

https://github.com/wxtim/rose/actions/runs/5940627765/job/16109634884 fails in an unhelpfully mysterious manner.

wxtim:
Supplanted by #2722

PR #2716, opened 2023-08-02
flake8

oliver-sanders:


oliver-sanders:
Applied the flake8 fix but it failed for Python 3.7/8, pushing a test up to see in the functional tests pass with a noqa to keep flake8 happy.

metomi/rose/config.py
```diff
@@ -217,8 +217,8 @@ def __eq__(self, other):
         try:
             for keys_1, node_1 in self.walk(no_ignore=True):
                 node_2 = other.get(keys_1, no_ignore=True)
-                if (
-                    type(node_1) != type(node_2)
+                if (  # noqa: E721
+                    type(node_1) != type(node_2)  #  noqa: E721
```
wxtim, line 221: I think that this is what flake8 wants you to do...

```suggestion
                    type(node_1) != type(node_2)  # noqa: E721
```

wxtim:
I fixed the cause of the test failure - I think I'm happy to say you can merge if you're happy with my change to your change. Else ask Ronnie for a quick once over.

oliver-sanders:
Np,

So `type(node_1) is not type(node_2)` caused test failures for newer versions of Python which is interesting, so have gone down the noqa route. The comment is on two lines as the line reported appeared to change with Python version.

Please squash merge.

PR #2723, opened 2023-08-25
Rose Stem tutorial without svn version control

wxtim:
Updates to make it clearer that is possible (and reasonably straightforward) to run rose stem suites where sources are not version controlled with FCM. 

### Rose stem tutorial

I chose not to create an entirely separate tutorial (even if it were mostly a copy) to avoid duplication. IMO we should also disentangle this from the reliance on FCM for Fortran building (but I haven't done that here).

### Rose stem docs

Rose stem is currently most thoroughly documented in `sphinx/tutorial/rose/furthertopics/rose-stem.rst`. This seems like a slightly odd place for it, because the script is stored in `cylc-rose`, and this document is not a tutorial.

### Review

I've tagged @oliver-sanders for a first review, since he reported this documentation bug.

IMO @dpmatthews would make a good reviewer in terms of general overview, but might be busy. 

If @markgrahamdawson can follow the instructions (avoid doing the FCM version control stuff) in the tutorial that might indicate that they are well written.

### DIFF
--- a/sphinx/tutorial/rose/furthertopics/rose-stem-tutorial.rst
+++ b/sphinx/tutorial/rose/furthertopics/rose-stem-tutorial.rst
@@ -20,9 +20,28 @@ This tutorial will walk you through creating a simple example of the
 Rose Stem testing system which will involve piloting a spaceship
 through space.
 
+.. admonition:: Without Version Control
+   :class: hint
+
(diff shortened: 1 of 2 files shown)


wxtim:
Not merged after a month.

PR #2733, opened 2023-10-23
linux.die.net anti crawler settings cause linkcheck failure

wxtim:
Copy of a "fix"* from https://github.com/cylc/cylc-doc/pull/653.

\* The fix is to ignore the link.


oliver-sanders:
> /home/runner/work/rose/rose/sphinx/tutorial/rose/furthertopics/rose-arch.rst:229:broken link: https://linux.die.net/man/1/rsync (403 Client Error: Forbidden for url: https://linux.die.net/man/1/rsync)

oliver-sanders (approved):

PR #2744, opened 2023-12-06
fileinstall: run coros in the background if loop already running

oliver-sanders:
One out of the triplet of PRs which closes https://github.com/cylc/cylc-rose/issues/274
* https://github.com/cylc/cylc-flow/pull/5868
* https://github.com/cylc/cylc-rose/pull/276

---

* Allow Rose file installation to be called by code which already has an event loop running by scheduling coroutines to run in the background (i.e. schedule but don't await).
* The calling code can list these tasks using `asyncio.all_tasks()` and await them as appropriate.
* Addresses https://github.com/cylc/cylc-rose/issues/274

---

For testing instructions see the cylc-flow PR.

### DIFF
--- a/metomi/rose/job_runner.py
+++ b/metomi/rose/job_runner.py
@@ -17,7 +17,15 @@
 """A multiprocessing runner of jobs with dependencies."""
 
 import asyncio
+
 from metomi.rose.reporter import Event
+
+
+# set containing references to "background" coroutines that are not referenced
+# from any code (i.e. are not directly awaited), adding them to this list
+# avoids the potential for garbage collection to delete them whilst they are
+# running
+_BACKGROUND_TASKS = set()
 
 
 class JobEvent(Event):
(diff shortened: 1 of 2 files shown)


metomi/rose/job_runner.py
```diff
@@ -17,9 +17,17 @@
...
 import asyncio
+
 from metomi.rose.reporter import Event
 
 
+# set containing references to "background" coroutines that are not referenced
+# from any code (i.e. are not directly awaited), adding them to this list
+# avoids the potential for garbage collection to delete them whilst they are
+# running
+_BACKGROUND_TASKS = set()
```
oliver-sanders, line 28: This ensures that background tasks are always referenced whilst they are running.

That is so long as the module doesn't get housekept :see_no_evil: 

For details: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

wxtim:
MyPy Failure is the same as on master!

wxtim (approved):
- [x] Read changes
- [x] Run tests locally
- [x] Tried to break various components of this

MetRonnie (approved):

PR #2765, opened 2024-02-28
Fix unescaped content injection in rosie disco

MetRonnie:
By having double quotes in rose-suite.info you could inject arbitrary HTML into rosie disco.

In this example, the description contained
```
           =the PDF document is displayed in the task "add_link_to_plots".
           =--------------------------------------------------------------------
           =There follows a description of template variables in
```

You can see how everything after 'in the task' is cut (and actually every word following this has become an attribute on the HTML element and caused the status bit to bleed into the element content)

image

I've tested out this fix by just running `lib/html/template/rosie-disco/prefix-index.html` through Jinja2 before and after

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,12 +3,22 @@
 Go to https://github.com/metomi/rose/milestones?state=closed
 for a full listing of issues for each release.
 
+## 2019.01.9 (Upcoming)
+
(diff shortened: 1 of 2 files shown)


oliver-sanders (approved):
LGTM, haven't tested.

PR #2779, opened 2024-05-10
Test that Cylc Rose workflow config retrieval respect compat mode

wxtim:
Tests for the combined effect of https://github.com/cylc/cylc-rose/pull/322 and https://github.com/cylc/cylc-flow/pull/6097 in closing https://github.com/cylc/cylc-rose/issues/319.


oliver-sanders (commented):
Confirmed that this test fails (without the cylc-flow & cylc-rose changes) with the following string appearing in the `job.err` file:

```
Output hello-make:succeeded can't be both required and optional
```

But passes in combination with the associated cylc-flow & cylc-rose branches :+1:.

oliver-sanders:
Re-tested against PRs merged into master branches, all good.

oliver-sanders:
Small problem, although this test works as expected locally, the CI tests have passed *before* the upstream PRs were merged.

Presumably these tests are skipped in CI?

oliver-sanders (approved):

t/rose-task-run/45-app-fcm-make-compat_mode.t
```diff
@@ -0,0 +1,62 @@
...
+if ! gfortran --version 1>/dev/null 2>&1; then
+    skip_all '"gfortran" unavailable'
+fi
+JOB_HOST="$(rose config --default= 't' 'job-host')"
+if [[ -n "${JOB_HOST}" ]]; then
+    JOB_HOST="$(rose host-select -q "${JOB_HOST}")"
+fi
+if [[ -z "${JOB_HOST}" ]]; then
+    skip_all '"[t]job-host" not defined or not available'
+fi
```
oliver-sanders, line 39: I don't think a job host is required to test this functionality?

wxtim:
> Small problem, although this test works as expected locally, the CI tests have passed _before_ the upstream PRs were merged.
> 
> Presumably these tests are skipped in CI?
> 
> ... Yes:
> 
> ```
> ./t/rose-task-run/45-app-fcm-make-compat_mode.t .................... skipped: "[t]job-host" not defined or not available
> ```

Yes - they rely on having a remote platform setup.

t/rose-task-run/45-app-fcm-make-compat_mode.t
wxtim: I tried it without and couldn't replicate the problem...

oliver-sanders: I managed to replicate this with the `[remote]host` logic commented out.

wxtim: Yes, but it was failing with the other branches checked out for an unrelated reason.

wxtim:
> I don't think that's right, `fcm_make` must load the workflow config in order to check whether there is a corresponding `fcm_make2` task. If `fcm_make2` is present, we would probably expect `fcm_make` to be local (it's pushing local data).

Yes, I've fixed the test. The problem was caused by my installing a PoC Yaml Config plugin which recent api changes in Cylc have broken.

oliver-sanders:
(tests should pass once upstream PRs are merged)

t/rose-task-run/45-app-fcm-make-compat_mode/app/make/file/src/hello.f90
```diff
@@ -0,0 +1,3 @@
+program hello
+write(*, '(a)') 'Hello World!'
+end program hello
```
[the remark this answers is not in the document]
MetRonnie: Is this Fortran? My first exposure to it..!

MetRonnie (commented):
Can't get the tests to pass, don't think I am set up for FCM correctly

oliver-sanders:
What error are you getting (check the job.err file)?

MetRonnie (approved):
At least I'm not getting the failure that the issue is about with https://github.com/cylc/cylc-rose/pull/322 and https://github.com/cylc/cylc-flow/pull/6097 checked out

oliver-sanders:
Upstream PRs are in, so I would expect the tests to pass, but they have failed again, @wxtim could you take a look.

t/rose-task-run/45-app-fcm-make-compat_mode/app/make/file/src/hello.f90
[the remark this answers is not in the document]
wxtim: Yes. Do you feel dizzy? Nauseous?

wxtim:
> Upstream PRs are in, so I would expect the tests to pass, but they have failed again, @wxtim could you take a look.

Unfortunately what was happening was that if fcm_make2 task was not set up to be on a remote machine the fcm make task would fail before or after the changes.

wxtim:
> I don't think that's right, `fcm_make` must load the workflow config in order to check whether there is a corresponding `fcm_make2` task. If `fcm_make2` is present, we would probably expect `fcm_make` to be local (it's pushing local data).

FCM make is local, but apparently needs to be convinced that make 2 is not, even if make 2 is never called!

t/rose-task-run/45-app-fcm-make-compat_mode/suite.rc
```diff
@@ -0,0 +1,16 @@
...
+    [[dependencies]]
+        graph = """
+            fcm_make:finish => fin
+            fcm_make:fail => fin
+        """
+
+[runtime]
+    [[fcm_make]]
+        script = rose task-run --app-key=fcm_make --new
+    [[fcm_make2]]
```
wxtim, line 14: Despite the `fcm_make2` not being in the graph, its config is what the `fcm_make` task is using Cylc Rose to carry out the lookup on, so it needs defined.

t/rose-task-run/45-app-fcm-make-compat_mode.t
```diff
@@ -0,0 +1,58 @@
...
+run_fail "${TEST_KEY_BASE}-play" \
+    timeout 120 \
+        cylc play \
+            "${FLOW}" \
+            --abort-if-any-task-fails \
+            --host='localhost' \
+            --no-detach \
+            --debug
+
+# It is a compat mode error.
```
wxtim, line 50:
```suggestion
# It is an error you would expect in compat mode.
```

wxtim:
I've changed this test: It now checks that the error logged by the `fcm_make` task refers to the platform (for `fcm_make2`) being garbage. Without the fix you instead get an error referring to the graph being wrong (which it is, unless you are in compat mode).

MetRonnie (approved):
Test now passes for me. And still fails if using prior commit of cylc-rose without the fix

oliver-sanders:
Mac OS failures unrelated ([latest run on master](https://github.com/metomi/rose/actions/runs/9476832638/job/26110380736))

oliver-sanders (approved):

PR #2783, opened 2024-06-06
Replace "template variables" with "template_variables"

### LINKED ISSUE #2737: template variables: space in section title incompatible with mini language
```ini
# rose-suite.conf
[jinja2:suite.rc]
FOO=1 
BAR=1 
```

```ini
# meta/rose_meta.conf
[jinja2:suite.rc=FOO]          
fail-if=jinja2:suite.rc=FOO == 1 and jinja2:suite.rc=BAR == 1     

[jinja2:suite.rc=BAR]
```

```console
$ rose metadata-check -C meta/
$ rose macro -V
[V] metomi.rose.macros.DefaultValidators: issues: 1
    jinja2:suite.rc=FOO=1
        failed because: jinja2:suite.rc=FOO == 1 and jinja2:suite.rc=BAR == 1
```

Now, take this same example and `s/jinja2:suite.rc/template variables`:

```console
$ rose metadata-check -C meta/
[V] rose.metadata_check.MetadataChecker: issues: 1
    template variables=FOO=fail-if=template variables=FOO == 1 and template variables=BAR == 1
        Not found: variables=BAR, variables=FOO
```

Rose doesn't like the space in `template variables` :cry:

wxtim:
... at the start of rose metadata-check to prevent the space being a violation of the rose mini-language.

closes #2737

### DIFF
--- a/metomi/rose/macros/rule.py
+++ b/metomi/rose/macros/rule.py
@@ -18,6 +18,8 @@
 import ast
 import re
 
+from functools import partial
+
 import jinja2
 import jinja2.exceptions
 
@@ -370,11 +372,8 @@ class RuleEvaluator(metomi.rose.macro.MacroBase):
         if log_ids is None:
             get_value_from_id = self._get_value_from_id
         else:
-            get_value_from_id = (
-                lambda id_, conf, m_conf, p_id: self._log_id_usage(
-                    id_, conf, m_conf, p_id, log_ids
-                )
-            )
+            get_value_from_id = partial(self._log_id_usage, id_set=log_ids)
+
         if not (rule.startswith('{%') or rule.startswith('{-%')):
             rule = "{% if " + rule + " %}True{% else %}False{% endif %}"
 
@@ -388,8 +387,13 @@ class RuleEvaluator(metomi.rose.macro.MacroBase):
         sci_num_count = -1
 
         # any/all processing.
+        rule = rule.replace('template variables', 'template_variables')
         for array_func_key, rec_regex in self.REC_ARRAY.items():
             for search_result in rec_regex.findall(rule):
+                search_result = (
+                    i.replace('template_variables', 'template variables')
+                    for i in search_result
+                )
                 start, var_id, operator, value, end = search_result
                 if var_id == "this":
                     var_id = setting_id
@@ -411,6 +415,10 @@ class RuleEvaluator(metomi.rose.macro.MacroBase):
 
         # len(...) processing.
         for search_result in self.REC_LEN_FUNC.findall(rule):
+            search_result = (
+                i.replace('template_variables', 'template variables')
+                for i in search_result
+            )
             start, var_id, end = search_result
             if var_id == "this":
                 var_id = setting_id
@@ -455,11 +463,17 @@ class RuleEvaluator(metomi.rose.macro.MacroBase):
                 x_id_num_str = search_result.replace("this", "").strip('()')
                 key = self.INTERNAL_ID_THIS_SETTING.format(x_id_num_str)
                 local_map[key] = value_string
+            search_result = search_result.replace(
+                'template variables', 'template_variables')
             rule = rule.replace(search_result, key, 1)
 
         # Replace ids (namelist:foo=bar) with their cast values.
         config_id_count = -1
+        rule = rule.replace('template variables', 'template_variables')
+
         for search_result in self.REC_CONFIG_ID.findall(rule):
+            search_result = search_result.replace(
+                'template_variables', 'template variables')
             value_string = get_value_from_id(
                 search_result, config, meta_config, setting_id
             )
@@ -470,6 +484,8 @@ class RuleEvaluator(metomi.rose.macro.MacroBase):
                 config_id_count += 1
                 key = self.INTERNAL_ID_SETTING.format(config_id_count)
                 local_map[key] = value_string
+            search_result = search_result.replace(
+                'template variables', 'template_variables')
             rule = rule.replace(search_result, key, 1)
 
         # Return the now valid Jinja2 template with a map of variables.
--- a/metomi/rose/macros/rule.py
+++ b/metomi/rose/macros/rule.py
@@ -387,6 +387,9 @@ class RuleEvaluator(metomi.rose.macro.MacroBase):
         sci_num_count = -1
 
         # any/all processing.
+        # n.b. Because we've made "template variables" (with space) the
+        # section header at Rose 2 we need to hack the rose mini language
+        # to allow a space in this one case:
         rule = rule.replace('template variables', 'template_variables')
         for array_func_key, rec_regex in self.REC_ARRAY.items():
             for search_result in rec_regex.findall(rule):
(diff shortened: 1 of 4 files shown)


oliver-sanders:
@wxtim, this issue may affect things other than `rose metadata-check` so I think this fix will need to be applied at the more-fundamental mini-language level, not the command level.

metomi/rose/macros/rule.py
```diff
@@ -370,11 +372,8 @@ def _process_rule(
         if log_ids is None:
             get_value_from_id = self._get_value_from_id
         else:
-            get_value_from_id = (
-                lambda id_, conf, m_conf, p_id: self._log_id_usage(
-                    id_, conf, m_conf, p_id, log_ids
-                )
-            )
+            get_value_from_id = partial(self._log_id_usage, id_set=log_ids)
```
wxtim, line 375: This is a refactor, but IMO so much easier to read that it's worth it.

t/rose-metadata-check/12-template-variables.t
```diff
@@ -0,0 +1,56 @@
...
+for ((i = 0; i < ${#CHECKS[@]}; i++)); do
+    CHECK="${CHECKS[$i]}"
+
+    TEST_KEY="$TEST_KEY_BASE::${CHECK}"
+    init <<__META_CONFIG__
+[template variables=FOO]
+fail-if=${CHECK}
+__META_CONFIG__
+
+    run_pass "$TEST_KEY" rose metadata-check -C ../config
```
oliver-sanders, line 50: Can you try adding a `rose macro -V` test to this to make sure this fix works for other commands too.

oliver-sanders (commented):
LGTM

oliver-sanders:
(oh, we should probably put this on 2.3.x)

oliver-sanders:
@wxtim, can you rebase to drop the "bump version" commit (doesn't belong on the 2.3.x branch) and squash the remainder.

metomi/rose/macros/rule.py
```diff
@@ -455,11 +463,17 @@ def _process_rule(
                 x_id_num_str = search_result.replace("this", "").strip('()')
                 key = self.INTERNAL_ID_THIS_SETTING.format(x_id_num_str)
                 local_map[key] = value_string
+            search_result = search_result.replace(
+                'template variables', 'template_variables')
             rule = rule.replace(search_result, key, 1)
 
         # Replace ids (namelist:foo=bar) with their cast values.
```
oliver-sanders, line 470: You've added this `search_result.replace` fix into most, but not all of the sections in this method.

These two remain unpatched:

```
# Number-like-strings into numbers.
# Strings into proper string variables.
```

It's ok if they are not needed there, but could you explain why it's not needed there for review.

oliver-sanders:
Tested with this config:

```
# ./meta/rose-meta.conf
[template variables=FOO]          
fail-if=template variables=FOO == 1 and template variables=BAR == 1     

[template variables=BAR]

[template variables=BAZ]
length=3
fail-if=template variables=BAZ(1) == 1 or len(template variables=BAZ) == 2 or any(template variables=BAZ == 4)

# ./rose-suite.conf
[template variables]
FOO=2
BAR=1
BAZ=2,2,3
```

Worked great.

metomi/rose/macros/rule.py
wxtim: Those loops where I've made a change the Regex will return `template variables=something` - the loops where I haven't the regexes only return the bit after the `=`:

```python
# For example the following are broken with `template variables`
>>> RuleEvaluator.REC_LEN_FUNC.findall('{% if len(template variables=FOO) == 1 %}True{% else %}False{% endif %}')
[]
>>> RuleEvaluator.REC_CONFIG_ID.findall('{% if len(template variables=FOO) == 1 %}True{% else %}False{% endif %}')
['variables=FOO']

# And fixed by adding `_`
>>> RuleEvaluator.REC_CONFIG_ID.findall('{% if template_variables=FOO == "2e1" %}True{% else %}False{% endif %}')
['template_variables=FOO']
>>> RuleEvaluator.REC_LEN_FUNC.findall('{% if len(template_variables=FOO) == 1 %}True{% else %}False{% endif %}')
[(' ', 'template_variables=FOO', ' ')]

# Wheras
>>> RuleEvaluator.REC_SCI_NUM.findall('{% if template_variables=FOO == "2e1" %}True{% else %}False{% endif %}')
['2e1']
>>> RuleEvaluator.REC_VALUE.findall('{% if template_variables=FOO == "2e1" %}True{% else %}False{% endif %}')
['"2e1"']

Comment (oliver-sanders) [reply metomi/rose/macros/rule.py:470]: Makes sense, just needed to understand why.

**Review by oliver-sanders — APPROVED** [review]

File: metomi/rose/macros/rule.py [hunk 4]
```diff
@@ -388,8 +387,13 @@ def _process_rule(
         sci_num_count = -1
 
         # any/all processing.
+        rule = rule.replace('template variables', 'template_variables')
         for array_func_key, rec_regex in self.REC_ARRAY.items():
             for search_result in rec_regex.findall(rule):
+                search_result = (
+                    i.replace('template_variables', 'template variables')
+                    for i in search_result
+                )
```
MetRonnie, line 396: Maybe add a comment explaining why you are replacing one thing one way and then doing the opposite for the other

metomi/rose/macros/rule.py
```diff
@@ -470,6 +484,8 @@ def _process_rule(
                 config_id_count += 1
                 key = self.INTERNAL_ID_SETTING.format(config_id_count)
                 local_map[key] = value_string
+            search_result = search_result.replace(
+                'template variables', 'template_variables')
```
MetRonnie, line 488: Again why replace it one way and then the opposite?

metomi/rose/macros/rule.py
```diff
@@ -388,8 +387,16 @@ def _process_rule(
         sci_num_count = -1
 
         # any/all processing.
+        # n.b. Because we've made "template variables" (with space) the
+        # section header at Rose 2 we need to hack the rose mini language
+        # to allow a space in this one case:
```
MetRonnie, line 392: Sorry but as someone not familiar with this bit of rose, I still don't understand why do space -> underscore followed by underscore -> space (and even once again space -> underscore down below)?

MetRonnie: So the reason for the back-and-forth is to hack around a regex not supporting spaces followed by a config lookup which has to have the exact setting name?

oliver-sanders: Yes, however, I think it's probably deeper than just a single Regex, the entire expression is executed in Jinja2 (which does not allow variable names to contain spaces).

wxtim: This function is trying to parse rules into jinja2 so if you put in 

```
rule='template variables=FOO == 2'
```

you will get out (assuming that you have set `FOO=1` in your `rose-suite.conf` and are running `rose macro -V`.

```
rule='{% if this == 2 %}True{% else %}False{% endif %}', local_map={'this': 1.0}
```

IMO this function would be easier to understand with unit tests showing its working. I'm not going to spend a lot of time on it but I'll have a quick look.

MetRonnie (approved):

PR #2795, opened 2024-06-28
eslint: run via npx

oliver-sanders:
We really don't need to maintain an eslint lock file, we just need a vaguely up-to-date version.


package.json
```diff
@@ -1,20 +0,0 @@
...
-  "name": "metomi-rose",
-  "private": true,
-  "scripts": {
-    "lint": "eslint -c .eslintrc.js sphinx/"
-  },
-  "dependencies": {
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-promise": "^4.2.1",
-    "eslint-plugin-standard": "^5.0.0"
```
MetRonnie, line 11: I'm not sure running `npx eslint` would have any way of knowing about these plugin dependencies anymore

.github/workflows/test.yml
```diff
@@ -68,7 +67,8 @@ jobs:
         run: etc/bin/shellchecker
 
       - name: ESLint
-        run: yarn run lint
+        run: |
+          npx --package=eslint -- eslint -c .eslintrc.js sphinx/
```
MetRonnie, line 71: Would have to be v8 max
```suggestion
          npx --package=eslint@8.x -- eslint -c .eslintrc.js sphinx/
```

package.json
MetRonnie: Would have to add a violation to one of the rules supplied by one of these plugins to see any difference

oliver-sanders: That is what I tried yes.

MetRonnie: In the latest run

> ESLint couldn't find the config "standard" to extend from.

PR #2797, opened 2024-07-16
GH Actions: patch DNS on MacOS

MetRonnie:
Spotted in #2783 we are getting the DNS issues here as well

https://github.com/actions/runner-images/issues/8649


MetRonnie:
Tests now failing on MacOS for other reasons...

wxtim (approved):

PR #2820, opened 2024-10-29
Add Joseph Abram to contributing.md & updating mailmap

J-J-Abram:


J-J-Abram:
Thanks Ollie, I've updated it and pushed it through

PR #2847, opened 2025-01-09
deploy: 2.4.0

wxtim:
Do we want to undeploy 2.1.0?


oliver-sanders:
> Do we want to undeploy 2.1.0?

Can tidy up old versions whenever.

oliver-sanders (approved):

PR #2848, opened 2025-01-09
added 2.4.0 to versions.json

wxtim:


oliver-sanders (approved):
Dammit, I should have spotted that

oliver-sanders:
Was the default version symlink chsnged in the prev PR?

oliver-sanders (changes requested):
I think we need to update the `doc` symlink to point at thr new versiom

oliver-sanders (approved):

oliver-sanders:
That aufght to do it

PR #2860, opened 2025-02-25
`setup.cfg` -> `pyproject.toml`

MetRonnie:
> This email is notifying you of an upcoming deprecation that we have determined may affect you as a result of your recent upload to 'metomi-rose'.
> 
> In the future, PyPI will require all newly uploaded source distribution filenames to comply with PEP 625. Any source distributions already uploaded will remain in place as-is and do not need to be updated.
> 
> Specifically, your recent upload of 'metomi-rose-2.4.1.tar.gz' is incompatible with PEP 625 because it does not contain the normalized project name 'metomi_rose'.
> 
> In most cases, this can be resolved by upgrading the version of your build tooling to a later version that supports PEP 625 and produces compliant filenames. You do not need to remove the file.
> 
> If you have questions, you can email admin@pypi.org to communicate with the PyPI admin@pypi.org to communicate with the PyPI administrators.


.github/workflows/2_auto_publish_release.yml
```diff
@@ -29,8 +29,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v5
       with:
-        # return to 3.x once cylc-flow is compatible with 3.10+ (pyuv)
-        python-version: '3.9'
+        python-version: '3.x'
```
MetRonnie, line 32: I thought we needed to specify the `setuptools` version in `pyproject.toml[build-system]` but it turns out it was just this after all

pyproject.toml
```diff
@@ -0,0 +1,145 @@
...
+
+[project]
+name = "metomi-rose"
+description = "Rose, a framework for meteorological suites."
+authors = [
+    { name = "British Crown (Met Office) & Contributors", email = "metomi@metoffice.gov.uk" }
+]
+readme = "README.md"
+license = { text = "GPL" }
+keywords = ["hpc", "weather-modelling", "weather-model", "meteorological-suites", "meteorological-modelling", "meteorological-models"]
```
wxtim, line 30: Only semi relevent, but we should consider adding "workflows" and "workflow automation" to this list.

MetRonnie:
(3.x test failure fixed by https://github.com/metomi/rose/pull/2861)

wxtim (approved):

MetRonnie:
Kicking tests

MetRonnie:
@wxtim Do you want a 2nd reviewer?

wxtim:
> @wxtim Do you want a 2nd reviewer?

I think it's worth it.

pyproject.toml
```diff
@@ -87,7 +87,7 @@ tests = [
     # version is pinned to the previous rose release)
 ]
 lint = [
-    "flake8>=4.0.0",
+    "flake8>=6.0.0; python_version > '3.7'",
```
MetRonnie, line 90: Flake8 v5 is the last version compatible with Python 3.7. However this version has a pin on `importlib-metadata` which conflicts with our own. So we can't run Flake8 on Python 3.7 anymore.

MetRonnie:
Build checks need https://github.com/cylc/release-actions/pull/102

oliver-sanders (approved):
LGTM.

Tested running the tutorial workflow from a build to make sure files are being packaged correctly.

Issue #2853, opened 2025-02-20
upgrade: namelist added in wrong state


oliver-sanders:
A bug where new settings are added into a namelist in an optional config by an upgrade macro, but the namelist is created in the wrong state.

This causes `rose app-upgrade` to produce a config which `rose macro -V` will identify as violating triggering rules.

Adding the namelist into the optional config in the trigger-ignored state *before* upgrade, does not seem to help.

https://gist.github.com/oliver-sanders/a1cd6e9ae1f6146a248e58519ce06bb3

<details>

<summary>[edit] gists loose the FS structure, here's a text version of the example</summary>

```
# ./app/example_app/opt/rose-app-demo.conf
[namelist:namelist_3]
existing_value=10.0

# ./app/example_app/rose-app.conf
meta=example_meta/vn1.0

[command]
default=true

[namelist:namelist_1]
use_namelist_2=.false.

[!!namelist:namelist_2]
!!other_value=1.0

[namelist:namelist_3]
existing_value=5.0

# ./rose-meta/example_meta/HEAD/rose-meta.conf
[namelist:namelist_1]
compulsory=true
title=Namelist 1

[namelist:namelist_1=use_namelist_2]
compulsory=true
description=logical to trigger namelist_2
trigger=namelist:namelist_2: .true. ;
       =namelist:namelist_2=other_value: .true. ;
       =namelist:namelist_2=new_value: .true. ;
type=logical

[namelist:namelist_2]
compulsory=true
title=Namelist 2

[namelist:namelist_2=other_value]
compulsory=true
description=other nml2 value, not used
type=real

[namelist:namelist_2=new_value]
compulsory=true
description=new value from upgrade macro. Is equal to nml3:existing_values
type=real

[namelist:namelist_3]
compulsory=true
title=Namelist 3

[namelist:namelist_3=existing_value]
compulsory=true
description=existing value to set nml2:new_value
type=real
# ./rose-meta/example_meta/tmp
from . import versions

# ./rose-meta/example_meta/vn1.0/rose-meta.conf
[namelist:namelist_1]
compulsory=true
title=Namelist 1

[namelist:namelist_1=use_namelist_2]
compulsory=true
description=logical to trigger namelist_2
trigger=namelist:namelist_2: .true. ;
       =namelist:namelist_2=other_value: .true. ;
       =namelist:namelist_2=new_value: .true. ;
type=logical

[namelist:namelist_2]
compulsory=true
title=Namelist 2

[...]

PR #2867, opened 2025-03-13
ConfigNodeDiff.get_as_opt_config: preserve parent attributes

oliver-sanders:
* If a setting (e.g. `namelist:a=b`) is present in a ConfigNodeDiff, but its parent (e.g. `namelist:a`) is not. Then the state and comments of parent node may not be present in the diff if they have not changed.
* As a result the state and comments of the parent node may be incorrect in the output generated by `get_as_opt_config`.
* This commit detects parent nodes that are not present in the diff and adds an option to transfer them from a provided node.
* closes https://github.com/metomi/rose/issues/2853

### DIFF
--- a/metomi/rose/config.py
+++ b/metomi/rose/config.py
@@ -771,25 +771,62 @@ class ConfigNodeDiff:
                     keys, settings_1[keys], settings_2[keys]
                 )
 
-    def get_as_opt_config(self):
+    def get_as_opt_config(self, base_config=None):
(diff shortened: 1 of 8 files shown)


wxtim (approved):

oliver-sanders:
(note to self, synchronise a cylc-rose bugfix release with the metomi-rose bugfix release we create for this fix)

MetRonnie (approved):

PR #2818, opened 2024-10-22
Fix traceback when git is not installed

MetRonnie:
Closes #2809 - fixes `rose app-run` and `rose task-run` crashing if git is not installed

Issue #2809, opened 2024-09-12
`rose app-run` fails if git is not installed


MetRonnie:
```
$ ./test-app-run 
+ cat rose-app.conf
[command]
default=echo Hello
+ rose app-run
Traceback (most recent call last):
  File "..../python3.9/site-packages/metomi/rose/popen.py", line 267, in run_bg
    proc = Popen(args, text=True, **kwargs)
  File "..../python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "..../python3.9/subprocess.py", line 1837, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'git'
 
During handling of the above exception, another exception occurred:
 
Traceback (most recent call last):
  File "/data/local/fcm/cylc-8.3.3-2/bin/rose", line 10, in <module>
    sys.exit(rose())
  File "..../python3.9/site-packages/metomi/rose/rose.py", line 281, in rose
    main(
  File "..../python3.9/site-packages/metomi/rose/rose.py", line 396, in main
    exec_sub_cmd(ns, sub_cmd, cmd_args)
  File "..../python3.9/site-packages/metomi/rose/rose.py", line 194, in exec_sub_cmd
    _exec_python(
  File "..../python3.9/site-packages/metomi/rose/rose.py", line 230, in _exec_python
    fcn()
  File "..../python3.9/site-packages/metomi/rose/app_run.py", line 553, in main
    runner = AppRunner(event_handler)
  File "..../python3.9/site-packages/metomi/rose/app_run.py", line 370, in __init__
    Runner.__init__(self, *args, **kwargs)
  File "..../python3.9/site-packages/metomi/rose/run.py", line 125, in __init__
    config_pm = ConfigProcessorsManager(event_handler, popen, fs_util)
  File "..../python3.9/site-packages/metomi/rose/config_processor.py", line 109, in __init__
    SchemeHandlersManager.__init__(
  File "..../python3.9/site-packages/metomi/rose/scheme_handler.py", line 98, in __init__
    handler = class_(*args, **kwargs)
  File "..../python3.9/site-packages/metomi/rose/config_processors/fileinstall.py", line 65, in __init__
    self.loc_handlers_manager = PullableLocHandlersManager(
  File "..../python3.9/site-packages/metomi/rose/config_processors/fileinstall.py", line 868, in __init__
[...]

PR #2869, opened 2025-03-25
Display failure messages from comms with Git remotes

wxtim:
I've been having some problems with my Github setup. Whilst it's still unclear what the cause is, I believe that this diff is worth pushing back up to master to help others who follow this way in future.


oliver-sanders (approved):
Tested with:

```
[file:foo]
source=git:foobar::x::y
```

**Before:**

```
[FAIL] file:foo=source=git:foobar::x::y: ls-remote: could not locate 'foobar'
```

**After:**

```
[FAIL] file:foo=source=git:foobar::x::y: ls-remote: could not locate 'foobar':
[FAIL]     fatal: 'foobar' does not appear to be a git repository
[FAIL]     fatal: Could not read from remote repository.
[FAIL] 
[FAIL]     Please make sure you have the correct access rights
[FAIL]     and the repository exists.
```

oliver-sanders:
@wxtim Want to put this on a bugfix branch?

t/rose-app-run/28-git.t
```diff
@@ -241,7 +241,12 @@ source=git:$TEST_DIR/zz9+zα/::earth/::v1
 __CONFIG__
 run_fail "$TEST_KEY" rose app-run --config=../config
 file_cmp "$TEST_KEY.err" "$TEST_KEY.err" <<__ERROR__
-[FAIL] file:hello/fruit_main=source=git:$TEST_DIR/zz9+zα/::earth/::v1: ls-remote: could not locate '$TEST_DIR/zz9+zα/'
+[FAIL] file:hello/fruit_main=source=git:$TEST_DIR/zz9+zα/::earth/::v1: ls-remote: could not locate '$TEST_DIR/zz9+zα/':
```
wxtim, line 244: I'm mildly concerned that this is vulnerable to Git error messages changing. Would it be better as a `grep_ok`?

oliver-sanders: I'm surprised this change didn't break this test?

oliver-sanders:
Might want to check the other `git` commands that are being run to make sure we aren't swallowing any other stderr.

wxtim:
> Might want to check the other `git` commands that are being run to make sure we aren't swallowing any other stderr.

## Other cases where git is called out for

### `git version`

[Here](https://github.com/metomi/rose/blob/f21144db816ad598055089ed6beef33fb93265e2/metomi/rose/loc_handlers/git.py#L48-L57)

I don't like that you get a failure to warn the user if git not installed:

```
> sed -i 's/GIT = "git"/GIT = "garbage"/'  metomi/rose/loc_handlers/git.py 
> python
Python 3.9.21 | packaged by conda-forge | (main, Dec  5 2024, 13:51:40) 
[GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from metomi.rose.config_processors.fileinstall import PullableLocHandlersManager
>>> from metomi.rose.loc_handlers.git import GitLocHandler
>>> h = GitLocHandler(PullableLocHandlersManager())
>>> h.git_version
# ... no warnings, no error, no version.... :(
```

### Other cases

Are all wrapped in `PullableLocHandlersManager.popen.run_ok_async` and not `.run`, which, looking at that code appears to fail in a less obscure way.

wxtim:
> Don't worry about the Python API, this isn't really public, so long as the CLI reports as expected.

I'm not, I'm just demoing what happens if no git.

MetRonnie:
> I think @MetRonnie 's suppression of the traceback inhttps://github.com/[/pull/2818](https://github.com/metomi/rose/pull/2818) also suppressed any error message at all.

It only suppresses the git not found error. If `git version` fails for some other reason then it will raise. Although I didn't think to check what happens if you try to *use* git as a file source when git is not installed... probably worth checking

wxtim:
> > I think @MetRonnie 's suppression of the traceback inhttps://github.com/[/pull/2818](https://github.com/metomi/rose/pull/2818) also suppressed any error message at all.
> 
> It only suppresses the git not found error. If `git version` fails for some other reason then it will raise. Although I didn't think to check what happens if you try to _use_ git as a file source when git is not installed... probably worth checking

I don't think this will occur very often, but if I've understood the code right, it will carry on and probably raise an error later when it can't use Git, then raise a potentially obscure error there.

MetRonnie:
> it will carry on and probably raise an error later when it can't use Git

I think that sounds ok, as long as the error isn't obscure!

I'm not sure that we should be logging anything when starting `rose app-run` or `rose task-run` if git isn't installed, there should not be an expectation that git is installed from what I gather

> `rose task-run` appears to require `git` to exist to run. This just hit me, and we really didn't want to have to have `git` installed on a compute node which is meant to be pretty minimal.

https://github.com/metomi/rose/issues/2809#issuecomment-2428202632

wxtim:
> I think that sounds ok, as long as the error isn't obscure!

I think we're good then. On reflection, I'm reasonably happy that any error won't be _too_ obscure.

oliver-sanders:
There are other git commands which could fail, e.g:

https://github.com/metomi/rose/blob/c9ee60dc9b9365b249e4bcedd382e21578a3cb25/metomi/rose/loc_handlers/git.py#L131-L143

Should we also log the stderr from these commands?

oliver-sanders (approved):

MetRonnie (approved):

PR #2892, opened 2025-04-24
attempt to handle rose rsync can_pull errors better

wxtim:
Fixes problem not recorded in an issue:

## Issue

Rose has a routine to work out which `file_loc_handler` should be used based on the content of the `source=` field. However, the function called by the `rsync` loc handler will attempt to run

```bash
# loc handler = foo:bar
ssh -n foo test -e bar
```
As part of the method used to decide if the source is an Rsync handler. This fails in a difficult to debug way, as it's unclear whether the rsync check routine failed because the host was unreachable, or the file didn't exist, or the `source` was not meant to be a location reached by `rsync`, or was just garbage.

## This PR

* Ensures that Rsync is always the last handler tried.
* Allows the rsync handler to raise more informative errors.
* Give the user more useful detail if the loc handler cannot be identified because this ssh check fails.


oliver-sanders:
Did you manage to work out why the command stderr was not logged?

wxtim:
> Did you manage to work out why the command stderr was not logged?

Because no one made Rose log it.

When I was testing it in front of you and there was no output, that was because `test -e nonexistant` doesn't return any stdout/err. 

On consideration, I should probably have a specific error message for this outcome!

oliver-sanders:
From a quick code skim, it looks like there is already logic to ensure that the commands stderr is reported:

https://github.com/metomi/rose/blob/26a6066e9f8d53caa89b7d976a0310e355e52187/metomi/rose/popen.py#L44-L61

So, I think all we need to do is `raise` this exception rather than swallow it?

metomi/rose/config_processors/fileinstall.py
```diff
@@ -891,12 +894,35 @@ def parse(self, loc, conf_tree):
...
+                        handler = self.guess_handler(loc)
+
+                    # The rsync handler will attempt to run
+                    # ssh host test -e path
+                    # This logic allows us to report failure nicely.
+                    except PreRsyncCheck255Error as exc:
+                        raise ValueError(
+                            exc.BASE_MESSAGE
+                            + 'If it is then host'
+                            f' \"{loc.name.split(":")[0]}\"'
```
MetRonnie, line 914: Unnecessary escapes?
```suggestion
                            f' "{loc.name.split(":")[0]}"'
```

metomi/rose/config_processors/fileinstall.py
```diff
@@ -891,12 +894,35 @@ def parse(self, loc, conf_tree):
...
+                            ' is uncontactable (ssh 255 error).'
+                        )
+                    except PreRsyncCheckError as exc:
+                        raise ValueError(
+                            exc.BASE_MESSAGE
+                            + f'`{exc.cmd}` failed with:'
+                            f'\n\t returncode: {exc.returncode}'
+                            f'\n\t stdout:     {exc.stdout}'
+                            f'\n\t stderr:     {exc.stderr}'
```
MetRonnie, line 923: If stdout or stderr have newlines then this might wrap awkwardly. Could use `textwrap.indent()` to nicefy

metomi/rose/loc_handlers/rsync.py
```diff
@@ -50,10 +86,15 @@ def can_pull(self, loc):
                 sleep(0.1)
             if proc.poll():
                 proc.kill()
-        except RosePopenError:
-            return False
+        except RosePopenError as exc:
+            return 'error' + ' '.join(exc.args)
```
MetRonnie, line 90: This is now truthy!

metomi/rose/loc_handlers/rsync.py
```diff
@@ -24,6 +25,41 @@
...
+
+        # Handle ``test -e nonexistant`` where no useful error is
+        # provided:
+        if (
+            self.returncode == 1
+            and self.stderr == ''
+            and self.stdout == ''
+        ):
+            self.stderr = f'File "{cmd[-1]}" does not exist.'
+        super().__init__()
```
MetRonnie, line 56: I don't think this does anything
```suggestion
```

metomi/rose/loc_handlers/rsync.py
```diff
@@ -24,6 +25,41 @@
 from metomi.rose.popen import RosePopenError
 
 
+class PreRsyncCheckError(Exception):
+    """Error to raise if the checks"""
```
MetRonnie, line 29: if the checks...?

metomi/rose/config_processors/fileinstall.py
```diff
@@ -891,12 +894,35 @@ def parse(self, loc, conf_tree):
...
+                        )
```
MetRonnie, line 924: Actually, wouldn't it make sense to put these error messages inside the exception classes?

metomi/rose/loc_handlers/rsync.py
wxtim: ...throw the papal envoy out of a window?

Will make this make sense.

metomi/rose/loc_handlers/rsync.py
wxtim: Whoops - I'd intended to check for this in the calling function and raise. TBH, might be better just raising?

metomi/rose/loc_handlers/rsync.py
```diff
@@ -24,6 +26,62 @@
...
+                self.BASE_MESSAGE
+                + f'`{self.cmd}` failed with:'
+                + indent(
+                    f'\nreturncode: {self.returncode}'
+                    f'\nstdout:     {self.stdout}'
+                    f'\nstderr:     {self.stderr}',
+                    prefix='    ',
+                )
+            )
+        super().__init__(self.message)
```
MetRonnie, line 78: Why this `super.__init__`?

metomi/rose/loc_handlers/rsync.py
```diff
@@ -24,6 +26,62 @@
...
+
+
+class PreRsyncCheck255Error(PreRsyncCheckError):
+    ...
+
+
```
MetRonnie, line 84: This one not used anymore in latest force push
```suggestion
```

metomi/rose/loc_handlers/rsync.py
wxtim: Mistake - it doesn't need to be any more

```suggestion
```

wxtim: I thought that there might be a better way - will mod.

t/rose-app-run/05-file.t
```diff
@@ -70,7 +70,8 @@ run_fail "$TEST_KEY" rose app-run --config=../config -q \
     --define='[file:hello4]source=stuff:ing'
 file_cmp "$TEST_KEY.out" "$TEST_KEY.out" </dev/null
 file_cmp "$TEST_KEY.err" "$TEST_KEY.err" <<'__CONTENT__'
-[FAIL] file:hello4=source=stuff:ing: don't know how to process stuff:ing
+[FAIL] Rose tried all other file install handlers and decided this must be an Rsync handler.
+[FAIL] 	If it is then host "stuff" is uncontactable (ssh 255 error).
```
MetRonnie, line 74: This has lost the context of `file:hello4=source=stuff:ing:` at the beginning?

MetRonnie (approved):

wxtim:
@oliver-sanders  Poke

PR #2893, opened 2025-04-30
Add zstd and xz support

david-rundle:
This commit adds support in rose for zstd and xz (with default options) using the same syntax as for gzip, including for compressing tarballs of various flavours.

It adds tests for zstd (46) and xz (47) that are clones of the gzip rose_arch test (32) and a slightly beefier test for zstd (48) that is a clone of (07).


david-rundle:
I also note that there's obviously a lot of very-near-duplication of code here and it would likely be better to implement something more generic (especially given the reliance on external utilities which effectively have identical command line arguments (by design, I think?)). But I am keen to get this in in some form initially and potentially revisit when I have more time (and more skills!) if it has the potential to be beneficial.

.mailmap
```diff
@@ -39,3 +39,4 @@ Joe Marsh Rossney    <17361029+jmarshrossney@users.noreply.github.com>
 Joe Marsh Rossney    <17361029+jmarshrossney@users.noreply.github.com> <17361029+marshrossney@users.noreply.github.com>
 Joseph Abram         <joseph.abram@metoffice.gov.uk>      J-J-Abram            <joseph.abram@metoffice.gov.uk>
 Joseph Abram         <joseph.abram@metoffice.gov.uk>      J-J-Abram            <98320699+J-J-Abram@users.noreply.github.com>
+David Rundle         <david.rundle@metoffice.gov.uk>      david-rundle          <david.rundle@metoffice.gov.uk>
```
wxtim, line 42:
```suggestion
David Rundle         <david.rundle@metoffice.gov.uk>      david-rundle          <david.rundle@metoffice.gov.uk>

```

sphinx/api/built-in/rose_arch.rst
```diff
@@ -206,6 +206,22 @@ with names in the form ``data_001.txt``:
...
+where multi-threaded compression can significantly improve throughput.
+
+.. code-block:: rose
+
+   [arch:large-data.tar.zst]
+   compress=zst
+   compress-cores=8
+   source=large-data/*
+
+In this example, the `zstd` will use 8 CPU cores.
```
oliver-sanders, line 223: I'm guessing this configures virtual cores as opposed to physical cores?

oliver-sanders (commented):
Thanks for this PR (and special thanks for sorting out the tests and documentation too :clap:)!

We'll try to get this in soon, but be aware, we've got a very heavy review workload right now, so apologies if it takes longer.

sphinx/api/built-in/rose_arch.rst
```diff
@@ -206,6 +206,22 @@ with names in the form ``data_001.txt``:
    rename-parser=^//some//path//data_(?P<serial_number>[0-9]{3})(?P<name_tail>.*)$
    rename-format=hello/%(cycle)s-%(name_head)s%(name_tail)s
 
+Using Multiple threads for Compression (zstd only)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
oliver-sanders, line 210: Error from Sphinx build:

> /home/runner/work/rose/rose/sphinx/api/built-in/rose_arch.rst:210:Title underline too short.

metomi/rose/apps/rose_arch_compressions/rose_arch_tar.py
```diff
@@ -70,3 +76,23 @@ def compress_sources(self, target, work_dir):
...
             self.app_runner.popen.run_simple(command, shell=True)
             self.app_runner.fs_util.delete(tar_name)
+
+        if target.compress_scheme in self.ZSTD_EXTS:
+            fdsec, zst_name = mkstemp(
+                suffix="." + target.compress_scheme, dir=work_dir
+            )
+            os.close(fdsec)
+            target.work_source_path = zst_name
+            command = f"zstd --rm -T{cores} -c '{tar_name}' >'{zst_name}'"
```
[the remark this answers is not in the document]
david-rundle: Thanks @oliver-sanders . I agree and I hope that the default being to not use this is a good trade-off between exposing useful functionality and maintaining order!

I have updated the documentation to note that this should be used with caution on shared resources.

sphinx/api/built-in/rose_arch.rst
david-rundle: I've updated the documentation to clarify (and correct) that it controls the number of threads that will be used during compression. I have noted that (a) the automatic setting [-T0] sets the number of threads to the number of physical cores and (b) that it is recommended not to exceed that number.

I think it is likely that there is little to be gained using hyperthreading (and in my tests it has been worse).

metomi/rose/apps/rose_arch_compressions/rose_arch_tar.py
```diff
@@ -70,3 +76,23 @@ def compress_sources(self, target, work_dir):
             command = "gzip -c '%s' >'%s'" % (tar_name, gz_name)
             self.app_runner.popen.run_simple(command, shell=True)
             self.app_runner.fs_util.delete(tar_name)
+
+        if target.compress_scheme in self.ZSTD_EXTS:
```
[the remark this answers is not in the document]
david-rundle: fixed thanks!

metomi/rose/apps/rose_arch_compressions/rose_arch_tar.py
```diff
@@ -70,3 +76,23 @@ def compress_sources(self, target, work_dir):
...
+            fdsec, zst_name = mkstemp(
+                suffix="." + target.compress_scheme, dir=work_dir
+            )
+            os.close(fdsec)
+            target.work_source_path = zst_name
+            command = f"zstd --rm -T{cores} -c '{tar_name}' >'{zst_name}'"
+            self.app_runner.popen.run_simple(command, shell=True)
+            self.app_runner.fs_util.delete(tar_name)
+
+        if target.compress_scheme in self.XZ_EXTS:
```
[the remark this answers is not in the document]
david-rundle: fixed thanks!

sphinx/api/built-in/rose_arch.rst
david-rundle: fixed thanks!

oliver-sanders:
The "test / docs" test seems to be failing for reasons unrelated to this change, please ignore.

The wonderfully named "test / test" tests, however, seem to be failing for legitimate reasons. It looks like `t/rose-task-run/07-app-arch.t` has become broken?

metomi/rose/apps/rose_arch.py
```diff
@@ -398,7 +411,8 @@ def _run_target_update(cls, dao, app_runner, compress_manager, target):
             # Compress sources
             if target.compress_scheme:
                 handler = compress_manager.get_handler(target.compress_scheme)
-                handler.compress_sources(target, work_dir)
+                compress_args = {"threads": target.compress_threads}
+                handler.compress_sources(target, work_dir, **compress_args)
```
wxtim, line 415: Tried a very simple check

```ini
mode=rose_arch

[arch]
# rose-app.conf
command-format=cp %(sources)s %(target)s
target-prefix=/home/users/tim.pillinger/cylc-src/rose-apps/arch/archive/
source-prefix=/home/users/tim.pillinger/cylc-src/rose-apps/arch/source/

[arch:world.out]
source='world.out'

[arch:gunzipme.gz]
source='gunzipme.out'

[arch:targunzipme.tar.gz]
source='targunzipme.out'
```

```bash
export CYLC_WORKFLOW_ID='hippo'
export CYLC_TASK_ID='task-run'
export CYLC_TASK_NAME='task-run'
export CYLC_TASK_CYCLE_POINT='task-run'
export CYLC_TASK_LOG_ROOT="${HERE}/log"

echo Running app from "${HERE}/app"
rose task-run --config="${HERE}/app"
```

 and got an error:

```
[FAIL] RoseArchGzip.compress_sources() got an unexpected keyword argument 'threads'
```
I think that you need to add the threads argument to `rose_arch_gzip.py` and possibly other items in that folder. You may want to consider emitting a warning `if threads != 1 and program_is_single_thread:`

david-rundle:
> The "test / docs" test seems to be failing for reasons unrelated to this change, please ignore.
> 
> The wonderfully named "test / test" tests, however, seem to be failing for legitimate reasons. It looks like `t/rose-task-run/07-app-arch.t` has become broken?

This test works for me locally? Is that the most recent version of the code?

oliver-sanders:
Hmm, testing locally (cazldf...), this test passes on upstream/master but fails on david-rundle:feature/add-zstd-xz, so this does look like a genuine error. Do you have zstd and xz installed on your box / in your environment? Does the test need them to be installed?

If the test does need these utils installed, let us know and we can work out where to install them in CI.

The error can be found from this line onwards:

https://github.com/metomi/rose/actions/runs/14859530104/job/41720858720#step:13:14

I've pulled out a couple of error messages that might be pertinent:

```
tar: /home/runner/cylc-run/rtb.20250506T123046Z/07-app-arch/hLA3QX/foo/20130101T0000Z/hello/worlds/unknown/stuff.pax: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
```

```
tar (child): /home/runner/cylc-run/rtb.20250506T123046Z/07-app-arch/hLA3QX/foo/20130101T1200Z/hello/worlds/planet-n.tar.gz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
```

metomi/rose/apps/rose_arch_compressions/rose_arch_tar.py
```diff
@@ -25,14 +25,20 @@ class RoseArchTarGzip:
...
+                      "pax.xz": tarfile.PAX_FORMAT}
     GZIP_EXTS = ["pax.gz", "tar.gz", "tgz"]
+    ZSTD_EXTS = ["pax.zst", "tar.zst"]
+    XZ_EXTS = ["pax.xz", "tar.xz", "txz"]
 
     def __init__(self, app_runner, *args, **kwargs):
         self.app_runner = app_runner
 
-    def compress_sources(self, target, work_dir):
+    def compress_sources(self, target, work_dir, threads="1"):
```
wxtim, line 41: IMO threads should be cast to int as soon as we parse the config.

wxtim:
I've written you a couple of integration tests at https://github.com/david-rundle/rose/pull/1

Tests running at https://github.com/wxtim/rose/actions/runs/14864492636

wxtim:
@david-rundle  - Is there anything I can do to help you move this forward?

david-rundle:
Just looking at this again now having merged your changes and am getting loads of errors like:

```./etc/bin/rose-test-battery -vv t/rose-task-run/46*.t

ok 1 - 46-app-arch-zstd-install
not ok 2 - 46-app-arch-zstd-play
ERROR - Incomplete tasks:
    * 1/archive did not complete the required outputs:
      ⨯ ┆  succeeded
DEBUG - stopping zmq replier...
DEBUG - ...stopped
DEBUG - stopping zmq publisher...
DEBUG - ...stopped
DEBUG - auth received API command b'TERMINATE'
DEBUG - Removing authentication keys from scheduler
INFO - DONE
not ok 3 - 46-app-arch-zstd-job.status-archive-01
t/rose-task-run/46-app-arch-zstd.t: line 51: cd: /home/users/david.rundle/cylc-run/rtb.20250731T115019Z/46-app-arch-zstd/uSHLqn/share/backup: No such file or directory
--- -	2025-07-31 12:50:30.909814205 +0100
+++ 46-app-arch-zstd-find.out	2025-07-31 12:50:30.905487529 +0100
@@ -1,2 +0,0 @@
-./archive.d/2016.txt.zst
-./archive.d/whatever.tar.zst
not ok 4 - 46-app-arch-zstd-find.out
Failed 3/4 subtests 

Test Summary Report
-------------------
t/rose-task-run/46-app-arch-zstd.t (Wstat: 0 Tests: 4 Failed: 3)
  Failed tests:  2-4
Files=1, Tests=4,  9 wallclock secs ( 0.02 usr  0.00 sys +  1.37 cusr  0.73 csys =  2.12 CPU)
Result: FAIL
```

and it's been a while so I don't really remember any of the workflow for running the tests so may need some help dusting this off!

wxtim:
@david-rundle - I've fixed the conflict for you, and retriggered the tests.

I have replicated the test failure locally, but I'm not clear on the causes.

Anyway - to run the test locally

```
conda env create --name cylc.dev python=3.9 pygraphviz pip        # Or use the same env you used before

cat > ~/conda-envs/cylc.dev/etc/conda/activate.d/cylc.sh <<__HERE__
#!/usr/bin/env bash
CYLC_ENV_NAME="$(basename "$CONDA_DEFAULT_ENV")"
export CYLC_ENV_NAME
__HERE__

cat > ~/conda-envs/cylc.dev/etc/conda/deactivate.d/cylc.sh <<__HERE__
#!/usr/bin/env bash
unset CYLC_ENV_NAME
__HERE__

conda install cylc-rose
cd ~/path/to/your/copy/of/rose

pip install -e .

./etc/bin/rose-test-battery t/rose-task-run/46-app-arch-zstd.t
```

I've tried running the tests and I'm getting the same failure - it looks like the workflow is failing because the task is failing, and you can look at the details of the failure at `~/cylc-run/rtb.<datestamp>/46-app-arch-zstd/p6oswd/log/job/1/archive/NN/job.err`. Mine is failing with

```
/home/users/tim.pillinger/cylc-run/rtb.20250731T155430Z/46-app-arch-zstd/p6oswd/log/job/1/archive/01/job: line 42: zstd: command not found
```

oliver-sanders:
New developments...

Python has now added support for zstd in version 3.14, this makes it much more attractive for use in Rose as we would no longer be relying on an external dependency (which may or may not be installed). Although this functionality would only be present when Python >= 3.14.0 is installed.

https://docs.python.org/3/library/compression.zstd.html

oliver-sanders:
(unassigning this to re-focus my review queue, feel free to re-assign when undrafted)

oliver-sanders:
Got some unhappy tests :(

* Please ignore the docs tests, they're broken for other reasons.
* The "check contributor list" check is failing because the commits on this branch appear to have been signed with different email addresses (making it look like they've come from different authors). Either fiddle the commits so the email address / name format match, or add a `.mailmap` entry to map the second name/address onto the first.
* The other tests appear to be failing for genuine reasons. If not already done, you might want to update the branch against `upstream/master` before investigating. You *should* get the same errors locally (let me know if you don't as this can happen):
  * To run the unit tests: `pytest`
  * To run the functional tests: `etc/bin/rose-test-battery` (use the `-j` argument to parallelise).

samuel-denton:
Looking good to me, just need to sort the contributor checks.

metomi/rose/apps/rose_arch_compressions/compression_util.py
```diff
@@ -0,0 +1,228 @@
...
+# the available memory, so they are never read in one go.
+#
+# This only applies to single-threaded compression (xz, or zstd when only
+# one thread is requested) - multi-threaded zstd compression uses the
+# "zstandard" distribution or the command line tool instead, both of
+# which manage their own internal chunking. A moderate value is used
+# here: too small (e.g. 1 MiB) adds unnecessary Python call overhead, and
+# going much bigger gives no benefit while risking memory pressure of its
+# own on constrained systems.
+CHUNK_SIZE = 16 * 1024 * 1024
```
oliver-sanders, line 54: Would be helpful to add the units into the comment.

sphinx/api/built-in/rose_arch.rst
```diff
@@ -272,19 +288,92 @@ Configuration
...
+
+            For multi-threaded ``zstd`` compression specifically, a Python
+            library is only used if the ``zstandard`` package is installed;
+            the standard library ``compression.zstd`` module (Python 3.14+)
+            is only used for single-threaded compression, as it has been
+            found not to scale reliably across threads. Setting the
+            ``ROSE_ARCH_ZSTD_FORCE_CLI`` environment variable (to any
+            non-empty value) forces the ``zstd`` command line tool to be
+            used unconditionally for zstd compression, regardless of
+            thread count or what is installed.
```
oliver-sanders, line 336: Should this say "since it cannot use multiple threads to compress a single source"?

I haven't dug too deep into this myself, I don't know whether Python's breaking of the GIL (search "freethreadded python") will have any impact here?

metomi/rose/apps/rose_arch_compressions/compression_util.py
```diff
@@ -0,0 +1,228 @@
...
+            from compression import zstd  # Python 3.14 and above
+        except ImportError:
+            pass
+        else:
+            return partial(_copy_compressed, zstd.ZstdCompressor)
+    try:
+        import zstandard  # the "zstandard" distribution
+    except ImportError:
+        return None
+    return zstandard.ZstdCompressor(threads=threads).copy_stream
```
oliver-sanders: TODO: you are here!

metomi/rose/apps/rose_arch_compressions/compression_util.py
oliver-sanders, line 1: Since this logic applies to all compressors, consider moving this code into `__init__.py`.

metomi/rose/apps/rose_arch_compressions/compression_util.py
```diff
@@ -0,0 +1,228 @@
...
+            import lzma
+        except ImportError:
+            return None
+        return partial(_copy_compressed, lzma.LZMACompressor)
+    if compressor == ZSTD:
+        return _get_zstd_compress_func(threads)
+    raise KeyError(compressor)
+
+
+def _get_zstd_compress_func(threads):
```
oliver-sanders: It would be better to locate the logic specific to each compressor within its own module for isolation and extendability.

I.e, the `_get_compress_func` could be implemented on each `RoseArchCompressor` implementation:

```python
class RoseArchZstd(RoseArchCompressor):

    """Compress archive sources in zstd."""

    SCHEMES = ["zst", "zstd"]
    COMPRESSOR = ZSTD

    def get_compress_func(threads):
        if os.environ.get(ZSTD_FORCE_CLI):
            return None
        ...
```

metomi/rose/apps/rose_arch_compressions/compression_util.py
```diff
@@ -0,0 +1,228 @@
...
+"""
+
+from functools import partial
+import os
+from shlex import quote
+
+# Compressor names. These are also the names of the command line fallbacks.
+GZIP = "gzip"
+XZ = "xz"
+ZSTD = "zstd"
```
oliver-sanders, line 31: The implementation of Rose is fairly modular, with abstract interfaces and concrete implementations.

The abstract interfaces (e.g, `RoseArchCompressor`) should hold the abstract interface and any generic logic. Things like `CHUNK_SIZE`, `compress` and `_copy_compressed` are good examples of generic boilerplate that are not specific to any one compressor in particular.

However, this module also contains aspects of the concrete implementations, e.g, rather than having a global list called `MULTI_THREADED`, the OO way would be to have an interface called `is_multi_threaded` to be implemented on each compressor individually. Other examples are `ZSTD_FORCE_CLI`, `_get_compress_func` and `_get_zstd_compress_func`.

It would be better to move the compressor-specific stuff into the compressor modules, translating these global implementations into interfaces for each compressor to implement.

metomi/rose/apps/rose_arch.py
```diff
@@ -314,6 +314,44 @@ def _run_target_setup(
...
+        )
+        if not target.compress_threads.isdigit():
+            app_runner.handle_event(
+                ConfigValueError(
+                    [t_key, "compress-threads"],
+                    target.compress_threads,
+                    ValueError(
+                        "compress-threads must be 0 (automatic) or a"
+                        " positive integer"
+                    ),
```
oliver-sanders, line 329: Question: Should we report this error (and fallback to a default), or should we just raise this error (and reject this configuration)?

It's likely that warnings might be missed if the application succeeds.

metomi/rose/tests/apps/rose_arch/test_rose_arch_compressions.py
```diff
@@ -0,0 +1,262 @@
...
+    if compressor == XZ:
+        import lzma
+
+        return lzma.decompress(data)
+    try:
+        from compression import zstd
+    except ImportError:
+        import zstandard
+
+        return zstandard.ZstdDecompressor().decompressobj().decompress(data)
```
oliver-sanders, line 166: With this approach we will test either `lzma` OR `zstandard` but not both which means that one could easily become broken without us realising.

This work introduces 5 new interfaces, 3 python, 2 CLI. We shouldn't expect changes to these interfaces going forward (🤞), but there is a risk of them becoming broken in the future.

What we can do, is to write tests for all of them, but add `pytest.mark.skip`s as appropriate to allow tests to skip if the relevant component is not installed. That way, we could test all of them in our CI, even if we only test the available ones locally.

oliver-sanders (commented):
Thanks for this PR.

The functionality looks good to me, tests and docs are nice :+1:.

We'll need to give a little thought to packaging here:
* We've added optional dependence on some external Python libraries.
  * We should probably make this a formal optional dependence by registering the dependency along with any minimum compatible version number in the [`pyproject.toml`](https://github.com/metomi/rose/blob/2373730caf9936d45d1438a83c9decb505af2f37/pyproject.toml#L66).
  * This helps us to manage our inventory of optional dependencies.
  * But also helps with [installing these dependences in CI](https://github.com/metomi/rose/blob/2373730caf9936d45d1438a83c9decb505af2f37/.github/workflows/test.yml#L165).
* We've also added optional dependence on some external CLI tools:
  * We should define these optional dependencies in the [`conda-environment.yml`](https://github.com/metomi/rose/blob/master/conda-environment.yml) along with any minimum compatible versions.
  * Should these be included with the default Rose conda package? (we provide default and a "bare bones" packages on conda-forge).
  * For CI testing, the dependencies [can be added here](https://github.com/metomi/rose/blob/2373730caf9936d45d1438a83c9decb505af2f37/.github/workflows/test.yml#L117-L128).

metomi/rose/apps/rose_arch_compressions/compression_util.py
david-rundle: yes, agreed. Done

sphinx/api/built-in/rose_arch.rst
david-rundle: I think it's not so much an issue that it _cannot_, because it appears to expose `nb_workers`. It's just that what I've tested with at the Met Office _does not_. There is some discussion about this (e.g. [here](https://github.com/astral-sh/python-build-standalone/issues/761)) but I haven't followed up on the details.

I think it's something to keep an eye on but the CLI is the tool to beat. It helps that it's basically a drop-in replacement for gzip too.

metomi/rose/apps/rose_arch_compressions/compression_util.py
david-rundle: yes, good idea. Done.

metomi/rose/apps/rose_arch_compressions/compression_util.py
david-rundle: You're right. I've done this now too.

metomi/rose/tests/apps/rose_arch/test_rose_arch_compressions.py
david-rundle: Yes ok. I have added the `lzma` tests which were missing and I think they should skip when the components are not installed.

PR #2894, opened 2025-05-07
Explicitly include testing Python 3.7 on legacy Ubuntu

wxtim:
Python 3.7 is no longer supported on Ubuntu latest.

A couple more websites are resisisting automatic link checks. 😞

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,10 +96,12 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest']
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', '3']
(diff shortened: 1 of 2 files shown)


oliver-sanders:
@wxtim, we are provisioning Python from Conda now. As it stands, we should be testing Python 3.7 against ubuntu-latest.

No need for legacy any more!

wxtim:
> @wxtim, we are provisioning Python from Conda now. As it stands, we should be testing Python 3.7 against ubuntu-latest.
> 
> No need for legacy any more!

Why is it not working on master?

MetRonnie:
But it is working on master, apart from linkcheck? https://github.com/metomi/rose/actions/runs/14875383672

wxtim:
My Brain no worky.

PR #2903, opened 2025-06-26
rosie lookup: quick improvements

oliver-sanders:
The `rosie lookup` interface is more important than it used to be due to the loss of `rosie go`.

However, the CLI help is not especially informative, especially on the query format.

Additionally, the tabulated output suffers from line wrapping which makes it very hard to read.

This is a quick slapdash attempt to improve the situation for users.

This PR:
* Improves the docs.
* Fixes the line wrap issue (caused by broken terminal size detection) - note lines will clip rather than wrapping.
* Changes the default format to a word-wrapped table (avoids the clipping problem).


metomi/rosie/cli_utils.py
```diff
@@ -0,0 +1,179 @@
...
+from subprocess import Popen, PIPE
+import textwrap
+from typing import Optional, List
+
+# NOTE: Box Drawing characters
+# t, b, l, r -> top, bottom, left, right
+# h, v -> horizontal, vertical
+# l, x -> line, cross (i.e. vertex)
+
+# unicode (U_) chars
```
oliver-sanders, line 27: Heavily inspired by the Cylc `--box` format system.

Search the internet for "box drawing characters" for more info.

metomi/rosie/ws_client_cli.py
```diff
@@ -295,13 +350,7 @@ def _display_maps(opts, ws_client, dict_rows, url=None):
     """Display returned suite details."""
     report = ws_client.event_handler
 
-    try:
-        terminal_cols = int(ws_client.popen("stty", "size")[0].split()[1])
-    except (IndexError, RosePopenError, ValueError):
-        terminal_cols = None
-
-    if terminal_cols == 0:
-        terminal_cols = None
```
oliver-sanders, line 304: The `stty size` command just does not work when run via the Rose Popener module.

I did try hacking it, but I couldn't figure out what Popener was doing to aggravate stty and gave in.

metomi/rosie/cli_utils.py
```diff
@@ -0,0 +1,208 @@
...
+
+    Args:
+        rows: 2D table composed of lists.
+        header: Optional 1D list to use as the table header.
+        max_width: Maximum permitted width for the whole table including
+            borders.
+        unicode: Use unicode characters if True, else fallback to ascii.
+
+    Examples:
+        >>> print(table(['a', 'b', 'c']))
```
wxtim, line 82: Should this work too? (It doesn't)
```suggestion
        >>> print(table(['aleph', 'bandicoot', 'coot']))
```

oliver-sanders: Removed the 1D table example.

metomi/rosie/cli_utils.py
```diff
@@ -0,0 +1,208 @@
...
+            for ind, col in enumerate(row):
+                widths[ind] = max(widths[ind], len(col))
+
+    def calc_width():
+        return (sum(widths) + (len(widths) * 3) + 1)
+
+    # resize cols to make them fix the max_width
+    if max_width:
+        overhang = max_width - calc_width()
+        while overhang < 0:
```
wxtim, line 142:
```python
print(table([["B", "Supercalifragilisticexpialgocious"]], header='hw', max_width=6))
```
dumps you into an infinite loop (`overhang == -1`)

oliver-sanders: Ooh, good spot!

oliver-sanders: Limited to 15 iterations and set a minimum column width of 1.

The table will now exceed the `max_width` in this situation rather than breaking.

wxtim (approved):

ChrisPaulBennett:
What is causing the doc check failure? Sphinx is complaining about a dead link, but the link is fine.

PR #2908, opened 2025-07-22
Remove upper pin on keyring dependency

### LINKED ISSUE #2794: setup: keyring
Upgrade from v23 to latest.

Note, v23 has not been built against Python 2.12 on conda-forge.

MetRonnie:
Closes #2794

~It appears keyring uses calendar versioning anyway~

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - python >= 3.7.0
   - aiofiles
   - jinja2 >=2.10.1
-  - keyring >=23,<24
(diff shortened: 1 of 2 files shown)


oliver-sanders:
It's a little risky as we have no tests for this, but it looks like a relatively slow moving package with few deprecations in the changelog.

I'm happy with this provided we've manually tested versions 23.0.0 and 25.6.0.

oliver-sanders (approved):
(make sure you've tested with the oldest/newest versions in the compat range before merging)

PR #2918, opened 2025-08-19
disable branch sync *to* 2019.01.x

oliver-sanders:


.github/workflows/branch_sync.yml
```diff
@@ -17,4 +17,5 @@ jobs:
     uses: cylc/release-actions/.github/workflows/branch-sync.yml@v1
     with:
       head_branch: ${{ inputs.head_branch }}
+      ignore_branches: 2019.01.x
```
MetRonnie, line 20: In case we end up with another 2019 branch for whatever reason
```suggestion
      ignore_branches: '2019.*'
```

MetRonnie (approved):

PR #2925, opened 2025-09-16
rosie ls: fix command breakage

oliver-sanders:
Fix breakage caused by https://github.com/metomi/rose/pull/2903

Turns out this code is called by other command(s).


oliver-sanders:
To test try:

```
rosie checkout u-aa001
rosie checkout u-aa002
rosie ls
```

wxtim (approved):

PR #2948, opened 2025-10-24
Fileinstall: Permit directories containing broken symlinks

### LINKED ISSUE #2946: Fileinstall fails if installing broken link
A [user report ](https://cylc.discourse.group/t/rose-fails-to-clone-with-broken-links/1257) of Rose fileinstall failing when in the source contains a broken link.

To replicate 

```bash
mkdir -p "${TMP}/."

# Install a file from a git source
cat > "${TMP}/./rose-app.conf" <<__ICI__
[command]
default=true

[file:git-source]
source=git:${TMP}/src::./::main
__ICI__

# Create that source
mkdir -p "${TMP}/src"
cd "${TMP}/src"
git init .
ln -s missing link
git add .
git commit -a -m "Add broken link"

# Run the rose app
mkdir -p "${TMP}/run"
cd "${TMP}/run"
rose app-run -C ../ --debug
```

This behaviour also appears to apply to rsync fileinstall, and may apply to others.

[broken-link-bug-git.sh](https://github.com/user-attachments/files/23020749/broken-link-bug-git.sh)
[broken-link-bug-rs.sh](https://github.com/user-attachments/files/23020748/broken-link-bug-rs.sh)

wxtim:
When complete will close #2946 .

After discussion with @dpmatthews  we think that broken links inside directories being installed should not cause a failure.

### DIFF
--- a/metomi/rose/checksum.py
+++ b/metomi/rose/checksum.py
@@ -70,8 +70,16 @@ def get_checksum(name, checksum_func=None):
             for filename in filenames:
                 filepath = os.path.join(path, filename)
                 source = os.path.join(name, filepath)
-                checksum = checksum_func(source, name)
-                mode = os.stat(os.path.realpath(source)).st_mode
+
+                if os.path.islink(source) and not os.path.exists(source):
+                    # If the source is a broken link within a directory
+                    # install without failing (unlike a single file)
+                    checksum = os.path.realpath(source)
+                    mode = os.lstat(source).st_mode
+                else:
+                    checksum = checksum_func(source, name)
+                    mode = os.stat(os.path.realpath(source)).st_mode
+
                 path_and_checksum_list.append((filepath, checksum, mode))
     return path_and_checksum_list
 
--- a/metomi/rose/loc_handlers/rsync_remote_check.py
+++ b/metomi/rose/loc_handlers/rsync_remote_check.py
@@ -62,7 +62,10 @@ def main(path, str_blob, str_tree):
                 if filename.startswith("."):
                     continue
                 name = os.path.join(dirpath, filename)
-                stat = os.stat(name)
+                if os.path.islink(name) and not os.path.exists(name):
+                    stat = os.lstat(name)
+                else:
+                    stat = os.stat(name)
                 print(stat.st_mode, stat.st_mtime, stat.st_size, name)
     elif os.path.isfile(path):
         print(str_blob)
--- a/metomi/rose/tests/test_checksum.py
+++ b/metomi/rose/tests/test_checksum.py
@@ -1,0 +1,82 @@
+# Copyright (C) British Crown (Met Office) & Contributors.
+# This file is part of Rose, a framework for meteorological suites.
+#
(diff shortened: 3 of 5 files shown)


metomi/rose/tests/test_checksum.py
wxtim: There was no prior unit testing at all for this module.

metomi/rose/checksum.py
```diff
@@ -70,8 +70,16 @@ def get_checksum(name, checksum_func=None):
...
-                mode = os.stat(os.path.realpath(source)).st_mode
+
+                if os.path.islink(source) and not os.path.exists(source):
+                    # If the source is a broken link within a directory
+                    # install without failing (unlike a single file)
+                    checksum = os.path.realpath(source)
+                    mode = os.lstat(source).st_mode
+                else:
+                    checksum = checksum_func(source, name)
+                    mode = os.stat(os.path.realpath(source)).st_mode
```
oliver-sanders: >  When installing directories using Git or Rsync broken symlinks will be copied.

This change targets the `get_checksum` function in general, not the git / rsync methods specificity.

Are we sure this method is only called for these methods, or should we pass the method name through to this function to ensure this logic is only activated when intended?

Also, this impacts Git, does it also impact SVN?

wxtim: > Are we sure this method is only called for these methods, or should we pass the method name through to this function to ensure this logic is only activated when intended?

I think that it makes sense to be consistent. The only other place this logic is used is Rose arch. I think it makes sense there too.

> Also, this impacts Git, does it also impact SVN?

I hope so - I would have tested it (and still can if you insist) but setting up the test would have taken extra time.

@dpmatthews  - what do you think?

metomi/rose/loc_handlers/rsync_remote_check.py
```diff
@@ -62,6 +62,9 @@ def main(path, str_blob, str_tree):
                 if filename.startswith("."):
                     continue
                 name = os.path.join(dirpath, filename)
+                # Bypass this check if rsyncing a broken symlink.
+                if os.path.islink(name) and not os.path.exists(name):
+                    continue
```
dpmatthews: Wouldn't it be better to use the same logic here as in checksum.py?

wxtim: Thus? (see changed code)

sphinx/api/configuration/file-creation.rst
```diff
@@ -116,6 +116,11 @@ root directory to install file targets with a relative path:
          ``~/.ssh/config`` to specify the user ID for logging into ``HOST``
          if required.
 
+      .. note::
+
+         * When installing files, broken symbolic links are rejected.
+         * When installing directories using Git or Rsync broken symlinks
```
dpmatthews: Remove "using Git or Rsync" - it applies in all cases (including svn).
Also add comma.

metomi/rose/checksum.py
wxtim: @oliver-sanders  please resolve if you're happy...

PR #2959, opened 2025-11-05
Update Rose Tutorial to use Canned Amazon Data

wxtim:
Update tutorials after https://github.com/cylc/cylc-flow/pull/7044, changing rainfall data source to Amazon.

N.b. https://github.com/cylc/cylc-flow/pull/7044 and subsequently [the update tutorials Workflow](https://github.com/metomi/rose/actions/workflows/update_tutorials.yml) will be required prior to this passing automated tests.

To test locally I've built this version of the docs at https://localserver/myID/doc/rose/killdatapoint/html/


metomi/rose/etc/tutorial/test-data/wind_20171101T0000Z_x.csv
oliver-sanders: We wave goodbye to that storm from 2017 :wave:

wxtim:
> Does the Rose tutorial now require the new HDF library thinggy?

Unclear - the only tutorial involving this new data which is ever going to work without Cylc/Cylc-Rose being installed is the `rose task-run -O test` tutorial, and that uses the data after it has been extracted to CSV.

I think that there is an implicit assumption that `cylc-flow[tutorial]` has been installed. Is there a reason that we haven't made this assumption explicit? Is there a risk of creating circular dependencies?

PR #2962, opened 2025-11-24
Move rose install to last step in test setup

wxtim:


wxtim:
Problem different to what it was assumed to be.

PR #2998, opened 2026-02-03
rose edit: license and lint

oliver-sanders:
* Update license headers to match changes to format on master.
* Remove Python 2 shebang line.
* Remove legacy Python 2 `class X(object)` inheritance.
* Run `ruff` auto fixers for a selection of lint codes used in cylc-flow.


wxtim (approved):
I've left one suggestion about how to make an if more legible. Otherwise this is fine. 👍🏼 

Pls sqsh mrge

PR #3003, opened 2026-02-13
3002.rose edit.traceback

### LINKED ISSUE #3002: Rose Edit: Traceback
### Reproduce thus

```bash
cd rose
rose edit -C demo/rose-config-edit/demo_meta/ 01-types
```

wxtim:
Closes #3002 

Turns out that if you try to close the splash screen twice you get traceback. I've added a handler to make this safer rather than trying to fix the circumstances in which it gets called twice.


oliver-sanders (approved):
Reproduced and confirmed as fixed :+1:

PR #3022, opened 2026-03-06
macro: fix return code

oliver-sanders:
Fix a return code reported as a bug.

This was reported as a bug in a situation where an upgrader macro was exiting `0` despite aborting before the upgrade could be performed.

This appears to be a situation where a critical error is handled with `sys.exit(0)`. The macro makes it no further through the code, but the return code doesn't suggest that anything went wrong.

### DIFF
--- a/changes.d/3022.fix.md
+++ b/changes.d/3022.fix.md
@@ -1,0 +1,1 @@
+The `rose macro` command will now exit 1 in an error scenario it previously exited 0 in.
--- a/changes.d/3022.fix.md
+++ b/changes.d/3022.fix.md
@@ -1,1 +1,1 @@
-The `rose macro` command will now exit 1 in an error scenario it previously exited 0 in.
+The `rose macro` command will now exit 1 instead of 0 in an error scenario.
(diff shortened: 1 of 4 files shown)


metomi/rose/macro.py
```diff
@@ -1081,7 +1081,7 @@ def pretty_format_config(config, ignore_error=False):
                             keylist[1], new_keylist[1]
                         )
                     )
-                    sys.exit(0)
```
oliver-sanders, line 1084: The code was exiting `0` at this point.

Whether exiting is a useful behaviour or not, exiting `0` is definitely wrong.

samuel-denton (approved):
Change makes sense to me.

changes.d/3022.fix.md
```diff
@@ -0,0 +1 @@
+The `rose macro` command will now exit 1 in an error scenario it previously exited 0 in.
```
MetRonnie:
```suggestion
The `rose macro` command will now exit 1 instead of 0 in an error scenario.
```

MetRonnie (approved):

PR #3030, opened 2026-03-25
Make rose check-software pick a sensible environment

wxtim:


metomi/rose/check_software.py
```diff
@@ -87,8 +114,15 @@ def cmd_version(command, command_template='--version',
...
-                   stderr=PIPE, text=True).communicate()[outfile - 1].strip()
+
+    cylc_env = get_cylc_env()
+
+    output = Popen(
+        ['conda', 'run', '-p', cylc_env] + [command] + command_template,
+        stdout=PIPE,
+        stderr=PIPE,
+        text=True,
+    ).communicate()[outfile - 1].strip()
```
oliver-sanders, line 96: @dpmatthews, this is a Cylc [wrapper script bug](https://github.com/cylc/cylc-flow/issues/7073), not an issue with this script:

* `conda run -n <env> rose check-software` will work
* `/path/to/wrapper check-software` will not.

To solve this issue, hack the wrapper script by adding another exception like this: https://github.com/cylc/cylc-flow/blob/b171201c456fb2ac43caf23e2cd17e2b347edd38/cylc/flow/etc/cylc#L131-L135

To fix the issue in this way, Tim has had to re-implement a chunk of the wrapper script logic inside this command and make the assumption that the user has installed Rose in a Conda environment, but they might not have installed it in an environment at all, Conda or otherwise.

oliver-sanders (changes requested):
This fixes the case where the `rose check-software` command is being run behind the Cylc wapper script and Rose is installed in a Conda environment.

However, it breaks all cases where `rose check-software` is installed via `pip` or `uv`.

metomi/rose/check_software.py
dpmatthews: Sorry, already discussed with Tim. There is no need to run an external command to check the Python version in which case most of this change is not needed.

oliver-sanders (approved):
LGTM.

I'm guessing we want this up against a bugfix branch?

oliver-sanders:
Superseded by #3045 (the same thing rebased onto the bugfix branch).

PR #2838, opened 2024-11-28
Fix.orphan

wxtim:
* Avoid displaying `:orphan:` directive in built pages.
* Remove unecessary includes for `hyperlinks.rst`.

Built copy at: https://wwwspice/~tim.pillinger/doc/rose/rose%202.4.0.dev/html/tutorial/rose/configurations.html

### Background

According to [this thread](https://github.com/readthedocs/readthedocs.org/issues/809) on readthedocs this is a problem with including RST metadata.

PR #3039, opened 2026-04-28
Fix docs

MetRonnie:
Follow-up to #2838

Also fixes broken link to image

### DIFF
--- a/sphinx/tutorial/rose/configurations.rst
+++ b/sphinx/tutorial/rose/configurations.rst
@@ -1,7 +1,3 @@
-
-    :start-line: 1
-
-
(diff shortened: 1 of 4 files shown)


samuel-denton (approved):
LGTM
Single review ok here @MetRonnie ?

MetRonnie:
Yes, for a small change like this. Another image link was also broken so I've force-pushed a fix, please merge after tests pass

samuel-denton (approved):
Both images now work for me.
the `:start-line: 1` blocks I don't believe would be doing anything outside a directive block so makes sense to remove them.

PR #3040, opened 2026-04-30
rosie disco: change service root to 'rosie'

dpmatthews:
- We have always used `/rosie` as the service root when hosting via Apache - this PR changes "service root mode" to match.
- The static content also needs to be served via the service root in order to run rosie disco as a service accessed via a proxy.

### DIFF
--- a/metomi/rosie/ws.py
+++ b/metomi/rosie/ws.py
@@ -86,6 +86,10 @@ class RosieDiscoServiceApplication(tornado.web.Application):
                     ".", 1
                 )[0]
         self.props["rose_version"] = ROSE_VERSION
+        if self.service_root_mode:
+            self.props["static_url_prefix"] = "/%s/static/" % (self.NAMESPACE)
+        else:
+            self.props["static_url_prefix"] = "/static/"
 
         # Get location of HTML files from package
         rosie_lib = str(files('metomi.rosie').joinpath(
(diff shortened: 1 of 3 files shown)


metomi/rosie/ws.py
```diff
@@ -116,7 +120,7 @@ def __init__(self, service_root_mode=False, *args, **kwargs):
         self.db_url_map = db_url_map
 
         # Specify the root URL for the handlers and template.
-        ROOT = "%s-%s" % (self.NAMESPACE, self.UTIL)
+        ROOT = "%s" % (self.NAMESPACE)
```
MetRonnie, line 123: This change seems to be causing the test failures. I tested by reverting this and the similar change below.

MetRonnie: Fixed by
```diff
$ git diff
diff --git a/t/lib/bash/test_header b/t/lib/bash/test_header
index 07a5915f3..5db8c611e 100644
--- a/t/lib/bash/test_header
+++ b/t/lib/bash/test_header
@@ -402,7 +402,7 @@ rose_ws_init() {
     TEST_ROSE_WS_PORT="${PORT}"
     if port_is_busy "${TEST_ROSE_WS_PORT}"; then
         pass "${TEST_KEY}"
-        TEST_ROSE_WS_URL="http://$(hostname -f):${TEST_ROSE_WS_PORT}/${NS}-${UTIL}"
+        TEST_ROSE_WS_URL="http://$(hostname -f):${TEST_ROSE_WS_PORT}/${NS}"
     else
         fail "${TEST_KEY}"
         rose_ws_kill
```

dpmatthews: Thanks

oliver-sanders (approved):
Looks reasonable, have tested server startup w/wo the `-R` option, seems to work.

PR #2366, opened 2019-07-09
Make Rosie Work under packaging (branched from #2360)

wxtim:
PR #2360 deactivated tests for rosie and rosa in the packaged version of rose. This ticket is to re-activate them:

- [x] Reactivate all skipped tests and check that these work.
  - [x] Rosa
    - [x] rosa svn-pre-commit
    - [x] rosa svn-post-commit
    - [x] rosa db-create
  - [x] rosie
    - [x] checkout
    - [x] create
    - [x] delete
    - [x] disco
    - [x] graph
    - [x] hello
    - [x] id
    - [x] lookup
    - [x] ls
    - [x] .db

PR #3042, opened 2026-05-06
rosie: fix ldap interface + simplify hello response

dpmatthews:


metomi/rosie/usertools/ldaptool.py
```diff
@@ -20,7 +20,7 @@
 """User information via LDAP."""
 
 try:
-    import ldap3 as ldap
```
dpmatthews, line 23: Changing back to `ldap`  (it was changed to `ldap3` in error in #2366)

metomi/rosie/ws.py
```diff
@@ -315,18 +315,17 @@ def get(self, *args):
 
 class HelloHandler(RosieDiscoService):
 
-    """Writes a 'Hello' message to the current logged-in user, else 'user'."""
+    """Writes a 'Hello' message."""
```
dpmatthews, line 317: The old code which reported the logged in user was removed when CherryPy was replaced.

metomi/rosie/ws.py
```diff
@@ -315,18 +315,17 @@ def get(self, *args):
...
 
-    HELLO = "Hello %s\n"
+    HELLO = "Hello\n"
```
dpmatthews, line 319: It is not desirable to report the user running the service

oliver-sanders (approved):
Have performed a basic `rosie hello` check (not actually covered by existing tests).

Looks good.

PR #3044, opened 2026-05-11
typing++

oliver-sanders:
MyPy has been updated, need to add a couple of types to keep it happy.


samuel-denton (approved):
LGTM - MyPy was failing on those lines and is now happy with them.

ChrisPaulBennett (approved):
LGTM

PR #3047, opened 2026-05-15
Updating tutorials to match config consolidation tutorials in cylc

samuel-denton:
https://wwwspice/~samuel.denton/rose-doc/consolidation_tutorial_update/html/tutorial/rose/applications.html
Few small changes to diff text and example code to realign with the cylc tutorial it follows on from.

### DIFF
--- a/sphinx/tutorial/rose/applications.rst
+++ b/sphinx/tutorial/rose/applications.rst
@@ -152,7 +152,7 @@ An application can be run using the :ref:`command-rose-app-run` command:
       come later).
    .. note::
       :class: hint
-      
+
       This practical can stand on it's own, but it will be easier to follow if
       you have already completed the
       :ref:`Cylc tutorials <tutorial-cylc-runtime-forecasting-workflow>`.
@@ -245,9 +245,9 @@ An application can be run using the :ref:`command-rose-app-run` command:
       * ``MAP_TEMPLATE``
 
       We will now add these into the application. This way, all of the
-      configuration specific to the application live within it.  The variable names
+      configuration specific to the application live within it. The variable names
       and values are provided here in case you have not completed the previous Cylc
-      tutorial.  We also adjust some of the values to point at the test data files
+      tutorial. We also adjust some of the values to point at the test data files
       so ensure you copy these values rather than from the previous tutorial.
 
       Add the following lines to the :rose:conf:`rose-app.conf[env]` section:
--- a/sphinx/tutorial/rose/applications.rst
+++ b/sphinx/tutorial/rose/applications.rst
@@ -176,7 +176,7 @@ An application can be run using the :ref:`command-rose-app-run` command:
 
       * The ``bin/forecast`` script.
       * The ``bin/util.py`` Python library.
-      * The ``file/map.html`` HTML template.
+      * The ``file/map-template.html`` HTML template.
 
       We will move all these resources into a single application directory.
 
--- a/sphinx/tutorial/rose/applications.rst
+++ b/sphinx/tutorial/rose/applications.rst
@@ -245,10 +245,7 @@ An application can be run using the :ref:`command-rose-app-run` command:
       * ``MAP_TEMPLATE``
 
       We will now add these into the application. This way, all of the
-      configuration specific to the application live within it. The variable names
-      and values are provided here in case you have not completed the previous Cylc
-      tutorial. We also adjust some of the values to point at the test data files
-      so ensure you copy these values rather than from the previous tutorial.
+      configuration specific to the application live within it.
 
       Add the following lines to the :rose:conf:`rose-app.conf[env]` section:
 
@@ -268,6 +265,12 @@ An application can be run using the :ref:`command-rose-app-run` command:
          MAP_FILE=map.html
          # The path to the HTML map template file.
          MAP_TEMPLATE=map-template.html
+
+      .. note::
+         :class: hint
+
+         The values given here are slightly different to those in the previous
+         Cylc tutorial; use these values.
 
       Note that the ``WIND_FILE_TEMPLATE`` and ``RAINFALL_FILE`` environment
       variables are pointing at files in the ``test-data`` directory.
--- a/sphinx/tutorial/rose/applications.rst
+++ b/sphinx/tutorial/rose/applications.rst
@@ -151,10 +151,16 @@ An application can be run using the :ref:`command-rose-app-run` command:
       the application standalone (running it as a task in the workflow will
       come later).
 
-   #. **Create a directory on your filesystem called ``rose-tutorial``::**
+      This practical can stand on it's own, but it will be easier to follow if
+      you have already completed the
+      :ref:`Cylc tutorials <tutorial-cylc-runtime-forecasting-workflow>`.
 
-      mkdir ~/rose-tutorial
-      cd ~/rose-tutorial
+   #. **Create a directory on your filesystem called** ``rose-tutorial``::
+
+      This directory will be used for multiple Rose tutorials.
+
+        mkdir ~/rose-tutorial
+        cd ~/rose-tutorial
 
    #. **Create a Rose application**
 
@@ -170,8 +176,8 @@ An application can be run using the :ref:`command-rose-app-run` command:
       The application gets three resources from different places:
 
       * The ``bin/forecast`` script.
-      * The ``lib/python/util.py`` Python library.
-      * The ``lib/template/map.html`` HTML template.
+      * The ``bin/util.py`` Python library.
+      * The ``file/map.html`` HTML template.
 
       We will move all these resources into a single application directory.
 
@@ -227,7 +233,8 @@ An application can be run using the :ref:`command-rose-app-run` command:
    #. **Move environment variables defined in the** ``flow.cylc`` **file.**
 
       In the ``[runtime][forecast][environment]`` section of the ``flow.cylc``
-      file in the weather-forecasting workflow we set a few environment variables:
+      file in the :ref:`Cylc tutorials <tutorial-cylc-runtime-forecasting-workflow>`
+      we set a few environment variables:
 
       * ``WIND_FILE_TEMPLATE``
       * ``WIND_CYCLES``
@@ -236,7 +243,10 @@ An application can be run using the :ref:`command-rose-app-run` command:
       * ``MAP_TEMPLATE``
 
       We will now add these into the application. This way, all of the
-      configuration specific to the application live within it.
+      configuration specific to the application live within it.  The variable names
+      and values are provided here in case you have not completed the previous Cylc
+      tutorial.  We also adjust some of the values to point at the test data files
+      so ensure you copy these values rather than from the previous tutorial.
 
       Add the following lines to the :rose:conf:`rose-app.conf[env]` section:
 
@@ -267,7 +277,7 @@ An application can be run using the :ref:`command-rose-app-run` command:
       file as well as the ``CYLC_TASK_CYCLE_POINT`` environment variable
       provided by Cylc when it runs a task.
 
-      Add the following lines to the :rose:file:`rose-app.conf`:
+      Add the following lines to the :rose:conf:`rose-app.conf[env]` section:
 
       .. code-block:: rose
 
@@ -293,5 +303,7 @@ An application can be run using the :ref:`command-rose-app-run` command:
          cd run
          rose app-run -C ../
 
+      ../ here is the path to the application directory, assuming you created
+      the run directory inside the application directory.
       The application should run successfully, leaving behind some files. Try
       opening the ``map.html`` file in a web browser.
--- a/sphinx/tutorial/rose/applications.rst
+++ b/sphinx/tutorial/rose/applications.rst
@@ -305,7 +305,10 @@ An application can be run using the :ref:`command-rose-app-run` command:
          cd run
          rose app-run -C ../
 
-      ../ here is the path to the application directory, assuming you created
-      the run directory inside the application directory.
+      .. note::
+         :class: hint
+
+         ``../`` is the path to the Rose application directory.
+
       The application should run successfully, leaving behind some files. Try
       opening the ``map.html`` file in a web browser.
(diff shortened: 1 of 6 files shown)


samuel-denton:
Not sure why MyPy is failing, some type annotation missing I think?

oliver-sanders:
MyPy fixes have just been merged to master. A quick close/reopen kicks off CI which will run on a fresh merge commit from master.

oliver-sanders:
@samuel-denton, please could you rebase this PR onto the 2.7.x branch for inclusion in the bugfix release.

samuel-denton:
Rebased and should be ready to review.

MetRonnie:
@samuel-denton Just needs deconfliction since #3049 has been merged

sphinx/tutorial/rose/applications.rst
```diff
@@ -150,11 +150,16 @@ An application can be run using the :ref:`command-rose-app-run` command:
       into a standalone Rose application. By the end you will be able to run
       the application standalone (running it as a task in the workflow will
       come later).
+      This practical can stand on it's own, but it will be easier to follow if
+      you have already completed the
+      :ref:`Cylc tutorials <tutorial-cylc-runtime-forecasting-workflow>`.
```
oliver-sanders, line 158: We've considered the Cylc tutorial to be a pre-requsite to the Rose tutorial, it is "fairly" standalone, but you'll need to understand Cylc to get what we're doing.

So we could leave blank here (but explain it [at the beginning](https://metomi.github.io/rose/doc/html/tutorial/rose/index.html)):

```suggestion
```

Otherwise, please stick a blankline between this and the rubric above, otherwise they'll get merged into one enormous header. Or stick it into a note/hint/tip/whatever:

```suggestion
   .. note::
      :class: hint
      
      This practical can stand on it's own, but it will be easier to follow if
      you have already completed the
      :ref:`Cylc tutorials <tutorial-cylc-runtime-forecasting-workflow>`.
```

sphinx/tutorial/rose/applications.rst
```diff
@@ -170,8 +175,8 @@ An application can be run using the :ref:`command-rose-app-run` command:
       The application gets three resources from different places:
 
       * The ``bin/forecast`` script.
-      * The ``lib/python/util.py`` Python library.
-      * The ``lib/template/map.html`` HTML template.
+      * The ``bin/util.py`` Python library.
+      * The ``file/map.html`` HTML template.
```
oliver-sanders:
```suggestion
      * The ``file/map-template.html`` HTML template.
```

sphinx/tutorial/rose/applications.rst
```diff
@@ -236,7 +242,10 @@ An application can be run using the :ref:`command-rose-app-run` command:
       * ``MAP_TEMPLATE``
 
       We will now add these into the application. This way, all of the
-      configuration specific to the application live within it.
+      configuration specific to the application live within it.  The variable names
```
oliver-sanders: FYI: The old double space after a full-stop typing convention does nothing in markup files, etc.

sphinx/tutorial/rose/applications.rst
```diff
@@ -236,7 +242,10 @@ An application can be run using the :ref:`command-rose-app-run` command:
...
+      and values are provided here in case you have not completed the previous Cylc
+      tutorial.  We also adjust some of the values to point at the test data files
+      so ensure you copy these values rather than from the previous tutorial.
```
oliver-sanders: What issue are we trying to prevent here?

The instruction is:

```
 Add the following lines to the :rose:conf:`rose-app.conf[env]` section:
```

sphinx/tutorial/rose/applications.rst
```diff
@@ -293,5 +302,7 @@ An application can be run using the :ref:`command-rose-app-run` command:
          cd run
          rose app-run -C ../
 
+      ../ here is the path to the application directory, assuming you created
+      the run directory inside the application directory.
```
oliver-sanders: > assuming you created the run directory inside the application directory.

The above command guarantees this, so I'm not sure it's helpful to bring this up?

Perhaps:

```suggestion
      .. note::
         :class: hint

         ``../`` is the path to the Rose application directory.

```

Or something along the lines of:

```suggestion
      .. note::
         :class: hint

         ``../`` is the path to the directory above (the one which contains the .... 

```

?

sphinx/tutorial/rose/metadata.rst
```diff
@@ -224,6 +227,8 @@ Metadata Items
       Next reload the metadata in the :ref:`command-rose-config-edit` window
       using the :menuselection:`Metadata --> Refresh Metadata` menu item.
       The descriptions should now display under each environment variable.
+      Comments removed from the :rose:file:`rose-app.conf` file may not disappear
+      on the refresh so you may need to close and reopen the window.
```
oliver-sanders: Good point!

Comments are part of the "configuration", not the "metadata". So refreshing the metadata does not remove the comments.

So we can change the cautious "may not" to a definite "wont".

I think we should chip this off of the previous paragraph too, along the lines of (plz edit to taste):

```suggestion

      .. note::
         :class: hint

         The comments won't disappear from the GUI until it is closed/reopened.
```

sphinx/tutorial/rose/suites.rst
```diff
@@ -477,6 +477,13 @@ Rose Applications In Rose Suite Configurations
          RAINFALL_FILE=$CYLC_WORKFLOW_WORK_DIR/$CYLC_TASK_CYCLE_POINT/get_rainfall/rainfall.csv
          MAP_FILE=${CYLC_TASK_LOG_ROOT}-map.html
 
+      .. tip::
+
+          We have removed the ``CYLC_TASK_CYCLE_POINT``, ``RESOLUTION`` and ``DOMAIN`` environment
+          variables; the ``RESOLUTION`` and ``DOMAIN`` are provided by the suite configuration and
+          since we are not using the test data, we want the ``CYLC_TASK_CYCLE_POINT`` variable to
+          be provided by Cylc rather than hard coded.
```
oliver-sanders: I struggle to parse this sentence.

I think we're trying to convey that there are two potential sources of these environment varaibles:

* When we run this application as part of a Cylc workflow, the `CYLC_TASK_CYCLE_POINT`, `RESOLUTION`` and `DOMAIN` environment variables; will come from Cylc.
* When we run this application standalone from the test data, they will come from the test optional configuration.

sphinx/tutorial/rose/applications.rst
samuel-denton: Yeah that's a strong habit that will take some breaking.  I guess less chars is better but not that big a deal?  Ill correct where I spot.

sphinx/tutorial/rose/applications.rst
samuel-denton: It was to prevent anyone trying to use the values they already have from the previous tutorials.
Not strictly needed given the instruction I guess, just being explicit.

sphinx/tutorial/rose/suites.rst
samuel-denton: That was there because, when I ran though the tutorial myself, I couldn't work out for ages why it didn't work, until I spotted I still had an old CYLC_TASK_CYCLE_POINT env var set in this config from an earlier step, which made it keep trying to run in the past and then failed to use the new data properly.  
It's there to catch the same issue for others, with an additional line explain which other configs shouldn't be present in this file any longer.

It could be re-worded to something like: 
"Note the env vars ``CYLC_TASK_CYCLE_POINT``, ``RESOLUTION`` and ``DOMAIN`` should no longer be set in this file. The ``RESOLUTION`` and ``DOMAIN`` are provided by the suite configuration, and the ``CYLC_TASK_CYCLE_POINT`` should be provided by Cylc."

oliver-sanders: > It could be re-worded to something like:

That makes sense. Perhaps clarify "this file" (these env vars should be in the opt conf, but not in the regular conf).

sphinx/tutorial/rose/applications.rst
```diff
@@ -170,8 +178,8 @@ An application can be run using the :ref:`command-rose-app-run` command:
       The application gets three resources from different places:
 
       * The ``bin/forecast`` script.
-      * The ``lib/python/util.py`` Python library.
-      * The ``lib/template/map.html`` HTML template.
+      * The ``bin/util.py`` Python library.
+      * The ``file/map-template.html`` HTML template.
```
MetRonnie: This was correct before. We're getting these _from_ those locations in the Cylc forecasting workflow.

sphinx/tutorial/rose/applications.rst
```diff
@@ -227,7 +235,8 @@ An application can be run using the :ref:`command-rose-app-run` command:
    #. **Move environment variables defined in the** ``flow.cylc`` **file.**
 
       In the ``[runtime][forecast][environment]`` section of the ``flow.cylc``
-      file in the weather-forecasting workflow we set a few environment variables:
+      file in the :ref:`Cylc tutorials <tutorial-cylc-runtime-forecasting-workflow>`
+      we set a few environment variables:
```
MetRonnie:
```suggestion
      we had set a few environment variables:
```

sphinx/tutorial/rose/applications.rst
samuel-denton: I thought that was the case, but I'm not sure that's useful, I figured it was more useful for the user to know which recourses they would end up with and their location.

Might be better to just have the file name and no path?

I know when I ran this, I spent a while looking for those locations and couldn't find them, so assumed it was a mistake.

sphinx/tutorial/rose/applications.rst
samuel-denton: I've reverted this text and added a small note instead:

```
      .. note::
         :class: hint

         The values given here are slightly different to those in the previous
         Cylc tutorial; use these values.
```

sphinx/tutorial/rose/suites.rst
samuel-denton: Reworded to:
          Note the env vars ``CYLC_TASK_CYCLE_POINT``, ``RESOLUTION`` and ``DOMAIN`` should no
          longer be set in :rose:file:`rose-app.conf`. The ``RESOLUTION`` and ``DOMAIN`` are provided
          by the suite configuration, and the ``CYLC_TASK_CYCLE_POINT`` should be provided by Cylc.

sphinx/tutorial/rose/suites.rst
```diff
@@ -477,6 +477,12 @@ Rose Applications In Rose Suite Configurations
          RAINFALL_FILE=$CYLC_WORKFLOW_WORK_DIR/$CYLC_TASK_CYCLE_POINT/get_rainfall/rainfall.csv
          MAP_FILE=${CYLC_TASK_LOG_ROOT}-map.html
 
+      .. tip::
+
+          Note the env vars ``CYLC_TASK_CYCLE_POINT``, ``RESOLUTION`` and ``DOMAIN`` should no
+          longer be set in :rose:file:`rose-app.conf`. The ``RESOLUTION`` and ``DOMAIN`` are provided
+          by the suite configuration, and the ``CYLC_TASK_CYCLE_POINT`` should be provided by Cylc.
```
oliver-sanders: This might cause confusion as I think we've asked them to add these variables into the `rose-app-test.conf` file?

samuel-denton: Ah yeah, that makes sense.  At this point I think they will be defined in the ``rose-suite.conf`` and overwritten in ``rose-app-test.conf`` so maybe it should say:

"Note the env vars ``CYLC_TASK_CYCLE_POINT``, ``RESOLUTION`` and ``DOMAIN`` should no longer be set in`rose-app.conf`. The ``RESOLUTION`` and ``DOMAIN`` are provided by `rose-suite.conf` and overwritten in `rose-app-test.conf`, and the ``CYLC_TASK_CYCLE_POINT`` should be provided by Cylc."

That's accurate but maybe a little convoluted, any suggestions?

sphinx/tutorial/rose/applications.rst
MetRonnie: I think the idea it was trying to communicate was, "In the forecasting workflow from the Cylc tutorials, we had these files in different places that may have been logical enough because they are different types of files, but seeing as they are only used for this app/task, they can be moved all together into the app directory"

oliver-sanders: The point of the exercise is:

> In this practical we will convert the forecast task from the weather-forecasting workflow in the [Cylc tutorials](https://cylc.github.io/cylc-doc/stable/html/tutorial/runtime/runtime-configuration.html#tutorial-cylc-runtime-forecasting-workflow) into a standalone Rose application.

So, yes, this was correct before. But I agree is was misleading as the instruction began:

> The application gets three resources from different places:

Note it says "application" (implicitly a Rose Application) when we are really talking about a Cylc task.

It is perhaps a little confusing to introduce this file hierarchy, but then provide the commands which construct it.

So, should it read something like:

```
The `forecast` task from the Cylc workflow used resources in three locations:
* ...
* ...
* ...

We will now create a Rose application which bundles all of these resources into a single directory.
```

?

samuel-denton: Updated back to old paths and added that text for clarity.

sphinx/tutorial/rose/applications.rst
```diff
@@ -175,13 +175,14 @@ An application can be run using the :ref:`command-rose-app-run` command:
...
 
       * The ``bin/forecast`` script.
-      * The ``bin/util.py`` Python library.
-      * The ``file/map-template.html`` HTML template.
+      * The ``lib/python/util.py`` Python library.
+      * The ``lib/template/map.html`` HTML template.
 
-      We will move all these resources into a single application directory.
+      We will now create a Rose application which bundles all of these resources into
+      a single directory.
```
MetRonnie: Sorry to be pernickety but we have already created the Rose app in the previous step

samuel-denton: How's this:

```suggestion
      We will now bundle these resources into a single directory in our Rose Application.
```

oliver-sanders: LGTM. Sorry, I hadn't skimmed above last time around.

Perhaps just, "we will now bundle these resource into our Rose Application".

samuel-denton:
```suggestion
      We will now bundle these resources into our Rose Application.
```

oliver-sanders (approved):
I think we good here :+1:

MetRonnie (approved):

PR #3048, opened 2026-06-01
Sync linting completed

### LINKED ISSUE #3034: sync linting from cylc-flow
Cylc Flow uses these Flake8 plugins:

https://github.com/cylc/cylc-flow/blob/f731469a46da66469af360b8ed2bf185111f59e8/setup.cfg#L100-L109

With a couple of rule exceptions:

https://github.com/cylc/cylc-flow/blob/master/tox.ini

Synchronise these plugins and exceptions to metomi-rose and fix any issues raised.

Note, you may find the `ruff check --select<code> --fix` command comes in handy here.

Crosslink:
* https://github.com/cylc/cylc-rose/issues/424
* https://github.com/cylc/cylc-uiserver/issues/809
* https://github.com/metomi/rose/issues/3034

Scott-Owen-James:
closes #3034

### DIFF
--- a/metomi/rose/apps/comparisons/exact.py
+++ b/metomi/rose/apps/comparisons/exact.py
@@ -28,12 +28,11 @@ class Exact:
         """Perform an exact comparison between the result and the KGO data"""
         if len(task.resultdata) != len(task.kgo1data):
             raise DataLengthError(task)
-        location = 0
         if not task.resultdata:
             task.set_pass(ExactComparisonSuccess(task))
             return task
-        for val1, val2 in zip(task.resultdata, task.kgo1data):
-            location += 1
+        for val1, val2, location in enumerate(zip(task.resultdata,
+                                                  task.kgo1data)):
             if val1 != val2:
                 task.set_failure(
                     ExactComparisonFailure(task, val1, val2, location)
--- a/metomi/rose/apps/comparisons/exact.py
+++ b/metomi/rose/apps/comparisons/exact.py
@@ -28,11 +28,11 @@ class Exact:
         """Perform an exact comparison between the result and the KGO data"""
         if len(task.resultdata) != len(task.kgo1data):
             raise DataLengthError(task)
-        location = 0
         if not task.resultdata:
             task.set_pass(ExactComparisonSuccess(task))
             return task
-        for val1, val2 in zip(task.resultdata, task.kgo1data):
+        for location, (val1, val2) in enumerate(zip(task.resultdata,
+                                                task.kgo1data)):
             location += 1
             if val1 != val2:
                 task.set_failure(
--- a/metomi/rose/apps/comparisons/exact.py
+++ b/metomi/rose/apps/comparisons/exact.py
@@ -31,8 +31,8 @@ class Exact:
         if not task.resultdata:
             task.set_pass(ExactComparisonSuccess(task))
             return task
-        for location, val1, val2 in enumerate(zip(task.resultdata,
-                                                  task.kgo1data)):
+        for location, (val1, val2) in enumerate(zip(task.resultdata,
+                                                task.kgo1data)):
             if val1 != val2:
                 task.set_failure(
                     ExactComparisonFailure(task, val1, val2, location)
--- a/metomi/rose/apps/comparisons/exact.py
+++ b/metomi/rose/apps/comparisons/exact.py
@@ -31,7 +31,7 @@ class Exact:
         if not task.resultdata:
             task.set_pass(ExactComparisonSuccess(task))
             return task
-        for val1, val2, location in enumerate(zip(task.resultdata,
+        for location, val1, val2 in enumerate(zip(task.resultdata,
                                                   task.kgo1data)):
             if val1 != val2:
                 task.set_failure(
--- a/metomi/rose/apps/comparisons/exact.py
+++ b/metomi/rose/apps/comparisons/exact.py
@@ -33,6 +33,7 @@ class Exact:
             return task
         for location, (val1, val2) in enumerate(zip(task.resultdata,
                                                 task.kgo1data)):
+            location += 1
             if val1 != val2:
                 task.set_failure(
                     ExactComparisonFailure(task, val1, val2, location)
--- a/metomi/rose/config_editor/data_helper.py
+++ b/metomi/rose/config_editor/data_helper.py
@@ -76,10 +76,8 @@ class ConfigDataHelper:
     def is_ns_content(self, ns):
         """Return whether a namespace has any existing content."""
         config_name = self.util.split_full_ns(self.data, ns)[0]
-        for section in self.get_sections_from_namespace(ns):
-            if section in self.data.config[config_name].sections.now:
-                return True
-        return self.is_ns_sub_data(ns)
+        return any(section in self.data.config[config_name].sections.now for
+                   section in self.get_sections_from_namespace(ns))
 
     def get_metadata_for_config_id(self, node_id, config_name):
         """Retrieve the corresponding metadata for a variable."""
--- a/metomi/rose/config_editor/data_helper.py
+++ b/metomi/rose/config_editor/data_helper.py
@@ -76,8 +76,11 @@ class ConfigDataHelper:
     def is_ns_content(self, ns):
         """Return whether a namespace has any existing content."""
         config_name = self.util.split_full_ns(self.data, ns)[0]
-        return any(section in self.data.config[config_name].sections.now for
-                   section in self.get_sections_from_namespace(ns))
+        if any(section in self.data.config[config_name].sections.now for
+           section in self.get_sections_from_namespace(ns)):
+            return True
+        else:
+            return self.is_ns_sub_data(ns)
 
     def get_metadata_for_config_id(self, node_id, config_name):
         """Retrieve the corresponding metadata for a variable."""
--- a/metomi/rose/config_editor/main.py
+++ b/metomi/rose/config_editor/main.py
@@ -641,68 +641,66 @@ class MainController:
             ("/TopMenuBar/Help/About", self.main_handle.about_dialog),
         ]
         is_toggled = dict(
-            [
-                (
-                    "/TopMenuBar/View/View fixed vars",
-                    metomi.rose.config_editor.SHOULD_SHOW_FIXED_VARS,
-                ),
-                (
-                    "/TopMenuBar/View/View ignored vars",
-                    metomi.rose.config_editor.SHOULD_SHOW_IGNORED_VARS,
-                ),
-                (
-                    "/TopMenuBar/View/View user-ignored vars",
-                    metomi.rose.config_editor.SHOULD_SHOW_USER_IGNORED_VARS,
-                ),
-                (
-                    "/TopMenuBar/View/View latent vars",
-                    metomi.rose.config_editor.SHOULD_SHOW_LATENT_VARS,
-                ),
-                (
-                    "/TopMenuBar/Metadata/Prefs/View without descriptions",
-                    metomi.rose.config_editor.SHOULD_SHOW_NO_DESCRIPTION,
-                ),
-                (
-                    "/TopMenuBar/Metadata/Prefs/View without help",
-                    metomi.rose.config_editor.SHOULD_SHOW_NO_HELP,
-                ),
-                (
-                    "/TopMenuBar/Metadata/Prefs/View without titles",
-                    metomi.rose.config_editor.SHOULD_SHOW_NO_TITLE,
-                ),
-                (
-                    "/TopMenuBar/View/View ignored pages",
-                    metomi.rose.config_editor.SHOULD_SHOW_IGNORED_PAGES,
-                ),
-                (
-                    "/TopMenuBar/View/View user-ignored pages",
-                    metomi.rose.config_editor.SHOULD_SHOW_USER_IGNORED_PAGES,
-                ),
-                (
-                    "/TopMenuBar/View/View latent pages",
-                    metomi.rose.config_editor.SHOULD_SHOW_LATENT_PAGES,
-                ),
-                (
-                    "/TopMenuBar/View/Flag opt config vars",
-                    metomi.rose.config_editor.SHOULD_SHOW_FLAG_OPT_CONF_VARS,
-                ),
-                (
-                    "/TopMenuBar/View/Flag optional vars",
-                    metomi.rose.config_editor.SHOULD_SHOW_FLAG_OPTIONAL_VARS,
-                ),
-                (
-                    "/TopMenuBar/View/Flag no-metadata vars",
-                    metomi.rose.config_editor.SHOULD_SHOW_FLAG_NO_META_VARS,
-                ),
-                (
-                    "/TopMenuBar/View/View status bar",
-                    metomi.rose.config_editor.SHOULD_SHOW_STATUS_BAR,
-                ),
-                (
-                    "/TopMenuBar/Metadata/Switch off metadata",
-                    self.metadata_off,
-                ),
-            ]
+            (
+                "/TopMenuBar/View/View fixed vars",
+                metomi.rose.config_editor.SHOULD_SHOW_FIXED_VARS,
+            ),
+            (
+                "/TopMenuBar/View/View ignored vars",
+                metomi.rose.config_editor.SHOULD_SHOW_IGNORED_VARS,
+            ),
+            (
+                "/TopMenuBar/View/View user-ignored vars",
+                metomi.rose.config_editor.SHOULD_SHOW_USER_IGNORED_VARS,
+            ),
+            (
+                "/TopMenuBar/View/View latent vars",
+                metomi.rose.config_editor.SHOULD_SHOW_LATENT_VARS,
+            ),
+            (
+                "/TopMenuBar/Metadata/Prefs/View without descriptions",
+                metomi.rose.config_editor.SHOULD_SHOW_NO_DESCRIPTION,
+            ),
+            (
+                "/TopMenuBar/Metadata/Prefs/View without help",
+                metomi.rose.config_editor.SHOULD_SHOW_NO_HELP,
+            ),
+            (
+                "/TopMenuBar/Metadata/Prefs/View without titles",
+                metomi.rose.config_editor.SHOULD_SHOW_NO_TITLE,
+            ),
+            (
+                "/TopMenuBar/View/View ignored pages",
+                metomi.rose.config_editor.SHOULD_SHOW_IGNORED_PAGES,
+            ),
+            (
+                "/TopMenuBar/View/View user-ignored pages",
+                metomi.rose.config_editor.SHOULD_SHOW_USER_IGNORED_PAGES,
+            ),
+            (
+                "/TopMenuBar/View/View latent pages",
+                metomi.rose.config_editor.SHOULD_SHOW_LATENT_PAGES,
+            ),
+            (
+                "/TopMenuBar/View/Flag opt config vars",
+                metomi.rose.config_editor.SHOULD_SHOW_FLAG_OPT_CONF_VARS,
+            ),
+            (
+                "/TopMenuBar/View/Flag optional vars",
+                metomi.rose.config_editor.SHOULD_SHOW_FLAG_OPTIONAL_VARS,
+            ),
+            (
+                "/TopMenuBar/View/Flag no-metadata vars",
+                metomi.rose.config_editor.SHOULD_SHOW_FLAG_NO_META_VARS,
+            ),
+            (
+                "/TopMenuBar/View/View status bar",
+                metomi.rose.config_editor.SHOULD_SHOW_STATUS_BAR,
+            ),
+            (
+                "/TopMenuBar/Metadata/Switch off metadata",
+                self.metadata_off,
+            ),
         )
         for address, action in menu_list:
             widget = self.menubar.uimanager.get_widget(address)
@@ -1043,11 +1041,11 @@ class MainController:
             )
             tab_window.connect(
                 "destroy-event",
-                lambda w, e: self.tab_windows.remove(w) and False,
+                lambda w, e: self.tab_windows.remove(w), False,
             )
             tab_window.connect(
                 "delete-event",
-                lambda w, e: self.tab_windows.remove(w) and False,
+                lambda w, e: self.tab_windows.remove(w), False,
             )
         else:
             tab_window = old_window
@@ -1413,7 +1411,7 @@ class MainController:
             metomi.rose.macro.pretty_format_config(config, ignore_error=True)
             try:
                 metomi.rose.config.dump(config, save_path)
-            except (OSError, IOError) as exc:
+            except OSError as exc:
                 metomi.rose.gtk.dialog.run_dialog(
                     metomi.rose.gtk.dialog.DIALOG_TYPE_ERROR,
                     metomi.rose.config_editor.ERROR_SAVE_PATH_FAIL.format(exc),
@@ -1496,7 +1494,7 @@ class MainController:
         try:
             os.mkdir(os.path.dirname(new_path))
             metomi.rose.config.dump(new_config, new_path)
-        except (OSError, IOError) as exc:
+        except OSError as exc:
             text = metomi.rose.config_editor.ERROR_CONFIG_CREATE.format(
                 new_path, type(exc), str(exc)
             )
@@ -1547,7 +1545,7 @@ class MainController:
         if dirpath is not None:
             try:
                 shutil.rmtree(dirpath)
-            except (shutil.Error, OSError, IOError) as exc:
+            except (shutil.Error, OSError) as exc:
                 text = metomi.rose.config_editor.ERROR_CONFIG_DELETE.format(
                     dirpath, type(exc), str(exc)
                 )
@@ -1770,9 +1768,9 @@ class MainController:
             )[0]
             self._get_pagelist()
             page_namespaces = [page.namespace for page in self.pagelist]
-            if config_name in configs:
-                if current_namespace in page_namespaces:
-                    self.view_page(current_namespace, current_id)
+            if (config_name in configs
+               and current_namespace in page_namespaces):
+                self.view_page(current_namespace, current_id)
 
     def load_custom_metadata(self):
         # open metadata dialog, use list() to pass by value
@@ -2142,7 +2140,7 @@ def spawn_window(
         full_namespaces = []
         for namespace in initial_namespaces:
             exp = re.compile(r"(.*%s?[^\/]+)" % (re.escape(namespace),))
-            for ns in sorted(sorted(ctrl.data.namespace_meta_lookup), key=len):
+            for ns in sorted(ctrl.data.namespace_meta_lookup, key=len):
                 match = exp.search(ns)
                 if match:
                     full_namespaces.append(match.groups()[0])
--- a/metomi/rose/config_editor/main.py
+++ b/metomi/rose/config_editor/main.py
@@ -1041,11 +1041,11 @@ class MainController:
             )
             tab_window.connect(
                 "destroy-event",
-                lambda w, e: self.tab_windows.remove(w), False,
+                lambda w, e: self.tab_windows.remove(w) and False,  # NOQA: 223
             )
             tab_window.connect(
                 "delete-event",
-                lambda w, e: self.tab_windows.remove(w), False,
+                lambda w, e: self.tab_windows.remove(w) and False,  # NOQA: 223
             )
         else:
             tab_window = old_window
@@ -2140,7 +2140,9 @@ def spawn_window(
         full_namespaces = []
         for namespace in initial_namespaces:
             exp = re.compile(r"(.*%s?[^\/]+)" % (re.escape(namespace),))
-            for ns in sorted(ctrl.data.namespace_meta_lookup, key=len):
+            for ns in sorted(sorted(  # NOQA: C414
+                                      ctrl.data.namespace_meta_lookup),
+                             key=len):
                 match = exp.search(ns)
                 if match:
                     full_namespaces.append(match.groups()[0])
--- a/metomi/rose/config_editor/menuwidget.py
+++ b/metomi/rose/config_editor/menuwidget.py
@@ -255,7 +255,7 @@ class MenuWidget(Gtk.Box):
             err_item.set_tooltip_text(self.my_variable.error[error])
             err_item.connect(
                 "activate",
-                lambda e: dialog_func(
+                dialog_func(
                     Gtk.STOCK_DIALOG_WARNING,
                     self.my_variable.error[error],
                     title,
@@ -275,7 +275,7 @@ class MenuWidget(Gtk.Box):
             warn_item.set_tooltip_text(self.my_variable.warning[warning])
             warn_item.connect(
                 "activate",
-                lambda e: dialog_func(
+                dialog_func(
                     Gtk.STOCK_DIALOG_INFO,
                     self.my_variable.warning[warning],
                     title,
--- a/metomi/rose/config_editor/menuwidget.py
+++ b/metomi/rose/config_editor/menuwidget.py
@@ -255,7 +255,7 @@ class MenuWidget(Gtk.Box):
             err_item.set_tooltip_text(self.my_variable.error[error])
             err_item.connect(
                 "activate",
-                dialog_func(
+                lambda e: dialog_func(
                     Gtk.STOCK_DIALOG_WARNING,
                     self.my_variable.error[error],
                     title,
@@ -275,7 +275,7 @@ class MenuWidget(Gtk.Box):
             warn_item.set_tooltip_text(self.my_variable.warning[warning])
             warn_item.connect(
                 "activate",
-                dialog_func(
+                lambda e: dialog_func(
                     Gtk.STOCK_DIALOG_INFO,
                     self.my_variable.warning[warning],
                     title,
--- a/metomi/rose/config_editor/page.py
+++ b/metomi/rose/config_editor/page.py
@@ -847,23 +847,22 @@ class ConfigPage(Gtk.Box):
             return
         if self.show_modes[metomi.rose.config_editor.SHOW_MODE_LATENT]:
             for widget in self.get_main_variable_widgets():
-                if hasattr(widget.get_parent(), "variable"):
-                    if widget.get_parent().variable.name == "":
-                        widget.get_parent().grab_focus()
-                        return
+                if (hasattr(widget.get_parent(), "variable")
+                   and widget.get_parent().variable.name == ""):
+                    widget.get_parent().grab_focus()
+                    return
         names = [v.name for v in (self.panel_data + self.ghost_data)]
         if focus_variable is None or focus_variable.name not in names:
             return
         if self.panel_data:
             for widget in self.get_main_variable_widgets():
-                if hasattr(widget.get_parent(), "variable"):
-                    var = widget.get_parent().variable
-                    if var.name == focus_variable.name:
-                        if var.metadata.get(
-                            "id"
-                        ) == focus_variable.metadata.get("id"):
-                            widget.get_parent().grab_focus()
-                            return
+                var = widget.get_parent().variable
+                if (hasattr(widget.get_parent(), "variable")
+                   and var.name == focus_variable.name
+                   and var.metadata.get("id") ==
+                   focus_variable.metadata.get("id")):
+                    widget.get_parent().grab_focus()
+                    return
 
     def refresh(self, only_this_var_id=None):
         """Reload the page or selectively refresh widgets for one variable."""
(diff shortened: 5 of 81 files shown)


oliver-sanders:
Hmmm:

* check-contributor/test: failing, but not because of this PR.
* test/lint: also failing for what appear to be unrelated issues
* test/test: appears to be failing for genuine reasons

```
+  File "/home/runner/micromamba/envs/rose-test/lib/python3.12/site-packages/metomi/rose/macros/compulsory.py", line 240, in _generate_aliases_for_sections
+    basic_sections = set(
+                     ^^^^
+TypeError: set expected at most 1 argument, got 2
```

metomi/rose/metadata_gen.py
```diff
@@ -103,12 +104,12 @@ def type_gen(value):
...
     """
-    types = []
     length = 0
+    types = []
     if not value:
-        return None, str(length)
-    for val in metomi.rose.variable.array_split(value):
-        length += 1
+        return None, 0
+    for val, lenTemp in enumerate(metomi.rose.variable.array_split(value)):
```
oliver-sanders: @Scott-Owen-James, I think these two are the wrong way around:

```suggestion
    for lenTemp, val in enumerate(metomi.rose.variable.array_split(value)):
```

Spotted from this error: https://github.com/metomi/rose/actions/runs/26812939552/job/79047364318?pr=3048#step:13:116

Scott-Owen-James: Corrected this and other existing instances of the same problem.

metomi/rose/config_editor/pagewidget/table.py
```diff
@@ -108,7 +108,7 @@ def add_variable_widget(self, variable):
     def attach_variable_widgets(self, variable_is_ghost_list, start_index=0):
         """Create and attach variable widgets for these inputs."""
 
-        for variable, is_ghost, counter in enumerate(variable_is_ghost_list):
+        for counter, variable, is_ghost in enumerate(variable_is_ghost_list):
```
oliver-sanders: I'm not sure this syntax is going to work:

```python
>>> x = [['a', 'b'], ['c', 'd']]

>>> for i, j, k in enumerate(x):
...     print(f'{i=} {j=} {k=}')
... 
ValueError: not enough values to unpack (expected 3, got 2)

>>> for i, (j, k) in enumerate(x):
...     print(f'{i=} {j=} {k=}')
... 
i=0 j='a' k='b'
i=1 j='c' k='d'
```

Scott-Owen-James: Corrected this and all other instances.

Scott-Owen-James: Seen and fixed.

oliver-sanders:
(re-running the failing checks...)

pyproject.toml
```diff
@@ -85,7 +85,15 @@ tests = [
     # version is pinned to the previous rose release)
 ]
 lint = [
-    "flake8>=6.0.0",
+    "flake8>=3.0.0",
+    "en-line>=0.3.0",
```
oliver-sanders: This line is causing the CI to fail:

https://github.com/metomi/rose/actions/runs/26880815410/job/80332875431?pr=3048#step:4:36

I'm not sure what this is?

pyproject.toml
```diff
@@ -85,7 +85,15 @@ tests = [
...
+    "flake8-bugbear>=21.0.0",
+    "flake8-builtins>=1.5.0",
+    "flake8-comprehensions>=3.5.0",
+    "flake8-debugger>=4.0.0",
+    "flake8-implicit-str-concat>=0.4",
+    "flake8-mutable>=1.2.0",
+    "flake8-simplify>=0.14.0",
     "flake8-type-checking",
```
oliver-sanders, line 97: This block doesn't quite match what we have in cylc-flow:

https://github.com/cylc/cylc-flow/blob/07cacb2537e7b13f254ed2674ecd192b1ab10900/setup.cfg#L100-L108

pyproject.toml
```diff
@@ -85,7 +85,15 @@ tests = [
...
-    "flake8>=6.0.0",
+    "flake8>=3.0.0",
+    "flake8-broken-line>=0.3.0",
+    "flake8-bugbear>=21.0.0",
+    "flake8-builtins>=1.5.0",
+    "flake8-comprehensions>=3.5.0",
+    "flake8-debugger>=4.0.0",
+    "flake8-implicit-str-concat>=0.4",
+    "flake8-mutable>=1.2.0",
+    "flake8-simplify>=0.14.0",
```
oliver-sanders, line 96: This line doesn't actually match what we have in cylc-flow, however, [looking at the history](https://github.com/cylc/cylc-flow/commit/09fbb6cac5fd59f54c618924bc0f9704ee2780c6), I think the version selector is not needed, so this is ok.

oliver-sanders:
Tests have now been fixed on master (with the exception of the docs test which will be fixed by https://github.com/metomi/rose/pull/3077).

Note, closing and re-opening the PR causes GitHub to re-run the CI on a fresh merge commit, pulling in recent changes.

metomi/rose/config_editor/main.py
```diff
@@ -1043,11 +1041,11 @@ def _handle_detach_request(self, page, old_window=None):
             )
             tab_window.connect(
                 "destroy-event",
-                lambda w, e: self.tab_windows.remove(w) and False,
+                lambda w, e: self.tab_windows.remove(w), False,
```
oliver-sanders: I'm not sure about this one, the lint error was:

```
SIM223 Use 'False' instead of '... and False'
```

So it was asking us to change this to `lambda w, e: False` rater than `lambda w, e: self.tab_windows.remove(w), False`.

This is a rather unusual piece of code. I think that it is trying to run ` self.tab_windows.remove(w)` and return `False` (irrespective of what `tab_windows.remove` returns. This is a rather unusual requirement and was solved in a rather sneaky way.

It might be that `, False` does the same thing? I haven't gone digging into the documentation for `tab_window.connect` to find out?

Unless we know this does the same thing, suggest putting it back the way it was, we can tell `flake8` to ignore this violation using a `noqa` comment (run `git grep noqa` for examples of how to do this).

metomi/rose/config_editor/main.py
```diff
@@ -1043,11 +1041,11 @@ def _handle_detach_request(self, page, old_window=None):
...
             )
             tab_window.connect(
                 "delete-event",
-                lambda w, e: self.tab_windows.remove(w) and False,
+                lambda w, e: self.tab_windows.remove(w), False,
```
oliver-sanders: Ditto.

metomi/rose/config_editor/main.py
```diff
@@ -2142,7 +2140,7 @@ def spawn_window(
         full_namespaces = []
         for namespace in initial_namespaces:
             exp = re.compile(r"(.*%s?[^\/]+)" % (re.escape(namespace),))
-            for ns in sorted(sorted(ctrl.data.namespace_meta_lookup), key=len):
+            for ns in sorted(ctrl.data.namespace_meta_lookup, key=len):
```
oliver-sanders: Ooh, this is another unusual one.

I've gotta feeling that the intention may have been to sort by length first, and alpha order second?

Here's an example to play around with:

```python
>>> with open('/usr/share/dict/words', 'r') as dictionary:
...     words = dictionary.readlines()
... 
>>> slice = words[1000:1050]
>>> sorted(sorted(slice), key=len) == sorted(slice, key=len)
False
```

oliver-sanders: (have raised an issue with the flake8 extension that implements C414 as this is a false positive)

metomi/rose/config_editor/menuwidget.py
oliver-sanders: This doesn't do the same thing as before.

With the lambda, the `dialog_func` would only get called at a later date (when this "connect" logic calls it), but without the lambda, it gets called straight away.

I can't see a flake8 error on this line?

metomi/rose/config_editor/menuwidget.py
```diff
@@ -275,7 +275,7 @@ def _popup_option_menu(self, option_ui, actions, event):
             warn_item.set_tooltip_text(self.my_variable.warning[warning])
             warn_item.connect(
                 "activate",
-                lambda e: dialog_func(
+                dialog_func(
```
oliver-sanders: Ditto.

metomi/rose/gtk/util.py
```diff
@@ -251,8 +251,8 @@ def __init__(
         stock_id=None,
         size=Gtk.IconSize.SMALL_TOOLBAR,
         tip_text=None,
-        menu_items=[],
-        menu_funcs=[],
+        menu_items=None,
+        menu_funcs=None,
```
oliver-sanders, line 255: The lint rule is quite right here, we shouldn't use mutable arguments.

However, we can't *just* change it to `None`, because the code is expecting a list, not a `None`.

The usual way to handle these situations is to check the value of the variable and apply the default value at the top of the function like so:

```python
def foo(a, b=None, c=None):
    if b is None:
        b = []
    if c is None:
        c = []
```

metomi/rose/gtk/util.py
```diff
@@ -288,7 +288,7 @@ def __init__(
 class ToolBar(Gtk.Toolbar):
     """An easier-to-use Gtk.Toolbar."""
 
-    def __init__(self, widgets=[], sep_on_name=[]):
+    def __init__(self, widgets=None, sep_on_name=None):
```
oliver-sanders, line 295: Ditto.

metomi/rose/gtk/util.py
```diff
@@ -318,7 +318,7 @@ def __init__(self, widgets=[], sep_on_name=[]):
             }
             self.insert(icon_tool_item, 0)
 
-    def set_widget_function(self, name, function, args=[]):
+    def set_widget_function(self, name, function, args=None):
```
oliver-sanders, line 329: Ditto.

metomi/rose/metadata_gen.py
```diff
@@ -103,12 +104,12 @@ def type_gen(value):
...
 
     """
-    types = []
     length = 0
+    types = []
     if not value:
-        return None, str(length)
-    for val in metomi.rose.variable.array_split(value):
-        length += 1
+        return None, 0
```
oliver-sanders: Not quite the same thing:

```python
>>> str(0)
'0'
>>> 0
0
```

metomi/rose/run.py
```diff
@@ -194,33 +195,26 @@ def run(self, opts, args):
...
-                    pass
             # Change back to original working directory
-            try:
+            with contextlib.suppress(OSError):
                 os.chdir(cwd)
-            except OSError:
-                pass
             # Reset os.environ
-            os.environ = dict(environ)
+            os.environ.clear()
```
oliver-sanders, line 218: Hmmm, I don't think this resets the environment back to what it was, I think it wipes it completely.

```python
>>> os.environ.clear()
>>> os.environ
environ({})
```

I don't seem to be getting a lint error on this line so am not sure it needs to change.

metomi/rose/variable.py
```diff
@@ -465,7 +465,7 @@ def _scan_trigger_string(string):
             and not any(is_in_quotes.values())
             and i + 1 < len(string)
         ):
-            for delim, token in delim_tokens.items():
+            for delim, _ in delim_tokens.items():
```
oliver-sanders: If we only need the keys we don't need to iterate over items:

```suggestion
            for delim in delim_tokens:
```

FYI:
* `.keys` gives us the keys.
* `.values` gives us the values.
* But `.items` gives us both.

sphinx/ext/auto_cli_doc.py
```diff
@@ -189,8 +186,6 @@ def rst_body(text):
 
 def write(ns, cmds, _write):
     for cmd, content in cmds.items():
-        if 'USAGE' not in content:
-            breakpoint()
```
oliver-sanders, line 193: MB 🙃

tox.ini
```diff
@@ -7,15 +7,40 @@ exclude=
     ; purposely corrupt file
     t/rose-metadata-check/lib/custom_macro_corrupt.py,
     node_modules
+    ;Temporary!
+    metomi/rose/macros/rule.py
```
oliver-sanders, line 11: Temporary?

Remind me why we've done this.

metomi/rose/config_editor/menuwidget.py
Scott-Owen-James: Nor can I, unfortunately I did it months ago and now can't remember what prompted the change, reverting it regardless.

metomi/rose/run.py
Scott-Owen-James: > Assigning to `os.environ` doesn't clear the environment. Subprocesses are going to see outdated variables, in disagreement with the current process. Use `os.environ.clear()` or the `env=` argument to Popen.(B003)

Set an ignore?

tox.ini
Scott-Owen-James: If I would remember correctly there was a handful of files and rules that were not ready/too complicated to fix so it was decided to have them temporarily be ignored.

metomi/rose/run.py
oliver-sanders: I don't seem to be getting an error in the first place, so am not sure an ignore is needed.

Note, we only need to respond to flake8 violations here as seen on the CLI, you might have additional checks turned on in your text editor.

metomi/rosie/svn_post_commit.py
```diff
@@ -421,15 +423,12 @@ def _update_info_db(self, dao, changeset_attribs, branch_attribs):
             "revision": changeset_attribs["revision"],
         }
         # Latest table
-        try:
+        with contextlib.suppress(al.exc.IntegrityError):
```
oliver-sanders, line 426: (preserving the comment from the `except` block)

```suggestion
        with contextlib.suppress(al.exc.IntegrityError):
            # NOTE: if idx and branch were just added: there is
            # no previous record.
```

oliver-sanders:
@Scott-Owen-James, the remaining lint failure is to do with that `os.environ` thing.

I think this is a legitimate problem (though I'm not sure it actually presents as a bug given the way Rose uses it). It has been recently spotted and fixed here: https://github.com/metomi/rose/pull/2893/changes/bdeddf4b7e9ab652ff0c3c14aa287e98ed5fc818

The easiest approach might be to employ that two line change (which I can confirm is correct).

metomi/rose/run.py
```diff
@@ -215,7 +215,8 @@ def run(self, opts, args):
             with contextlib.suppress(OSError):
                 os.chdir(cwd)
             # Reset os.environ
-            os.environ = dict(environ)
+            os.environ.clear()
+            os.environ.update(environ)
```
oliver-sanders, line 219: @Scott-Owen-James, @samuel-denton 

Here's a quick example to help explain why the linter was complaining here and demonstrate the fix:

```python
import os
import sys
from subprocess import Popen


old_environ = dict(os.environ)
os.environ['ANSWER'] = '42'


def test(label):
  print(f'# {label}')

  print(f'{os.environ.get("ANSWER", "")=}')

  sys.stdout.write(f'$ echo $ANSWER = ')
  sys.stdout.flush()
  Popen('echo $ANSWER', shell=True).wait()
  print('\n')


def crude_clear():
  os.environ = old_environ


def proper_clear():
  os.environ.clear()
  os.environ.update(old_environ)


test('before')
# crude_clear()
proper_clear()
test('after')
```

With the previous method (`crude_clean`):
```
# before
os.environ.get("ANSWER", "")='42'
$ echo $ANSWER = 42


# after
os.environ.get("ANSWER", "")=''
$ echo $ANSWER = 42
```

With the new method:

```
# before
os.environ.get("ANSWER", "")='42'
$ echo $ANSWER = 42


# after
os.environ.get("ANSWER", "")=''
$ echo $ANSWER = 
```

metomi/rose/config_editor/main.py
```diff
@@ -2142,7 +2140,9 @@ def spawn_window(
         full_namespaces = []
         for namespace in initial_namespaces:
             exp = re.compile(r"(.*%s?[^\/]+)" % (re.escape(namespace),))
-            for ns in sorted(sorted(ctrl.data.namespace_meta_lookup), key=len):
+            for ns in sorted(sorted(  # NOQA: C414
+                                      ctrl.data.namespace_meta_lookup),
+                             key=len):
```
samuel-denton, line 2145: Has this been split over 3 lines so the NOQA fits in the allowed line length? If so could this work:
```
            for ns in sorted(sorted(ctrl.data.namespace_meta_lookup), key=len): # NOQA: C414,E501
```
Or just:
```
            for ns in sorted(sorted(ctrl.data.namespace_meta_lookup), key=len): # NOQA
```
Bit pedantic but if feels wrong to alter code to make the NOQA comment fit.

oliver-sanders: > `# NOQA: C414,E501`

No excuse to `NOQA` a line length rule when the line can easily fit within length IMO.

> `# NOQA`

Better to be explicit (otherwise we're disabling all checks for this line)

Scott-Owen-James: You are correct, that is the reason for that, unfortunately there's several other with similar issues too.

metomi/rose/gtk/splash.py
```diff
@@ -264,11 +265,8 @@ def run(self):
             GObject.idle_add(self._update_splash_screen, update_input)
 
     def stop(self):
-        try:
+        with contextlib.suppress(RuntimeError):
             Gtk.main_quit()
-        except RuntimeError:
-            # This can result from gtk having already quit.
-            pass
```
samuel-denton, line 271: There are 2 examples of this in this file (and a few in other files); whats the reason for supressing a warning here rather than using the previous try catch?

oliver-sanders: This PR is syncing the set of lint rules we use in cylc-flow to metomi-rose.

The lint rule which requests this change is one of them.

Try reverting and linting to see...

samuel-denton: Did a bit of learning and understand now, these examples are all errors we pass on exception, which linters don't like, and this is a more explicit way of saying we know this error can be raised here but we want to ignore it.  Identical but slightly more readable.

sphinx/ext/auto_cli_doc.py
```diff
@@ -129,12 +130,8 @@ def parse(text):
     lines = text.splitlines()
     ret = {}
 
-    try:
+    with contextlib.suppress(AttributeError):
         ret['USAGE'] = RE_USAGE.search(lines[0]).group()
-    except AttributeError:
-        pass
-    except IndexError:
-        breakpoint()
```
samuel-denton, line 137: This one does appear to be a change in behaviour but might be intended?  Not used breakpoint() a lot but I think it's normally a temporary debug line so might have been left accidentally previously?

Scott-Owen-James: Yes, there were a couple left in there by accident, general policy is to remove them, the indexError might have been legitimate for testing or something though, can't remember, will have to double check.

samuel-denton (approved):
All LGTM.  I'm happy that every change is either equivalent to prior or intended.
Cleaning up tracebacks with ```raise ... from None``` is nice, something I need to remember to use.

oliver-sanders:
> Cleaning up tracebacks with raise ... from None is nice, something I need to remember to use.

Good news on that front, our linting won't let us forget!

tox.ini
oliver-sanders: There were a few cases where violations needed to be ignored (false positives really), but the rest could be handled:

https://github.com/Scott-Owen-James/rose/pull/1

oliver-sanders:
Wash up: https://github.com/Scott-Owen-James/rose/pull/1

oliver-sanders:
@samuel-denton, please could you eyeball the last couple of commits which are new since your review.

Plz ignore the check contributor list failure - will be resolved by a sync PR later.

I'm going to ping copilot just to double check this diff. Copilot, if you're listening, this is purely a lint change with a single bugfix, we are expecting this to work just as it did before. We're interested in any cases where behaviour may differ as a result of the change, but are not interested in comments on the code itself beyond this.

samuel-denton (approved):
Yep last 3 commits look good to me

samuel-denton (approved):
Latest commit looks good to me, I think it addresses all co-pilots comments, mostly a few extra brackets that aren't needed, moving a variable assignment back to after the hasattr check and adding back a fallback for is_ns_content().
The main change (hardest to follow) is correcting a previous change that had slightly altered a set of conditionals in config_editor/pagewidget/table.py.

To be honest, these are all things I should have spotted in my review but missed.  Does highlight that co-pilot can be helpful when reviewing a large number of small changes like this where it's very hard to spot small mistakes without writing/performing tests for every change.

metomi/rose/apps/comparisons/exact.py
```diff
@@ -28,11 +28,11 @@ def run(self, task):
...
         if len(task.resultdata) != len(task.kgo1data):
             raise DataLengthError(task)
-        location = 0
         if not task.resultdata:
             task.set_pass(ExactComparisonSuccess(task))
             return task
-        for val1, val2 in zip(task.resultdata, task.kgo1data):
+        for location, (val1, val2) in enumerate(zip(task.resultdata,
+                                                task.kgo1data)):
             location += 1
```
MetRonnie, line 36: By keeping the `location += 1`, this introduces a bug

metomi/rose/config_editor/panelwidget/summary_data.py
```diff
@@ -993,7 +993,7 @@ def get_model_data(self):
         sub_sect_names = list(self.sections.keys())
         sub_var_names = []
         self.var_id_map = {}
-        for section, variables in list(self.variables.items()):
+        for _, variables in list(self.variables.items()):
```
MetRonnie, line 996:
```suggestion
        for variables in self.variables.values():
```

metomi/rose/config_editor/plugin/um/widget/stash_add.py
```diff
@@ -172,7 +172,7 @@ def generate_tree_view(self, is_startup=False):
         if is_startup:
             group_model = Gtk.TreeStore(str)
             group_model.append(None, [""])
-            for i, name in enumerate(self.column_names):
+            for _, name in enumerate(self.column_names):
```
MetRonnie, line 175:
```suggestion
            for name in self.column_names:
```

MetRonnie (changes requested):

metomi/rose/config_editor/valuewidget/array/entry.py
```diff
@@ -341,9 +341,7 @@ def populate_table(self, focus_widget=None):
             self.add_button.show()
         if self.max_length.isdigit() and len(self.entries) <= int(
             self.max_length
-        ):
-            self.del_button.hide()
-        elif len(self.entries) == 0:
+        ) or len(self.entries) == 0:
```
MetRonnie, line 344: For clarity, can we have parentheses around the first check?
```python
if (
    self.max_length.isdigit()
    and len(self.entries) <= int(self.max_length)
) or len(self.entries) == 0:
```
This is so that it's clear what is going on, `(A and B) or C` vs `A and (B or C)`

MetRonnie (changes requested):

metomi/rose/apps/comparisons/exact.py
Scott-Owen-James: I vaguely remember some sort of issue with this that requires that line to be there. In any case removing that line causes test failures, I'm not sure for the specifics though..

metomi/rose/apps/comparisons/exact.py
```diff
@@ -28,12 +28,11 @@ def run(self, task):
...
         if len(task.resultdata) != len(task.kgo1data):
             raise DataLengthError(task)
-        location = 0
         if not task.resultdata:
             task.set_pass(ExactComparisonSuccess(task))
             return task
-        for val1, val2 in zip(task.resultdata, task.kgo1data):
-            location += 1
+        for location, (val1, val2) in enumerate(zip(task.resultdata,
+                                                task.kgo1data)):
```
MetRonnie: Ah, also need `start=1`
```suggestion
        for location, (val1, val2) in enumerate(
            zip(task.resultdata, task.kgo1data), start=1
        ):
```

metomi/rose/config_editor/valuewidget/array/mixed.py
```diff
@@ -234,9 +234,7 @@ def _decide_show_buttons(self):
         else:
             self.add_button.show()
             self.del_button.show()
-        if len(self.rows) == 1:
-            self.del_button.hide()
-        elif len(self.rows) == 2 and self.has_titles:
+        if len(self.rows) == 1 or len(self.rows) == 2 and self.has_titles:
```
MetRonnie:
```suggestion
        if len(self.rows) == 1 or (len(self.rows) == 2 and self.has_titles):
```

metomi/rose/config_editor/data_helper.py
```diff
@@ -76,10 +76,11 @@ def is_ns_sub_data(self, ns):
     def is_ns_content(self, ns):
         """Return whether a namespace has any existing content."""
         config_name = self.util.split_full_ns(self.data, ns)[0]
-        for section in self.get_sections_from_namespace(ns):
-            if section in self.data.config[config_name].sections.now:
-                return True
-        return self.is_ns_sub_data(ns)
+        if any(section in self.data.config[config_name].sections.now for
+           section in self.get_sections_from_namespace(ns)):
```
MetRonnie: This code style is a bit difficult to read. Generally we prefer code to look something like the "[black](https://github.com/psf/black)" code style
```suggestion
        if any(
            section in self.data.config[config_name].sections.now
            for section in self.get_sections_from_namespace(ns)
        ):
```

metomi/rose/config_editor/page.py
```diff
@@ -1325,26 +1324,23 @@ def _get_page_info_widgets(self):
...
-                error_label = Gtk.Label()
-                info = metomi.rose.config_editor.ERROR_ORPHAN_SECTION.format(
-                    self.section.name
-                )
-                error_label.set_text(info)
-                error_label.show()
-                button_list.append(error_button)
-                label_list.append(error_label)
+        ) and self.section is not None and self.section.name.startswith(
+                "namelist:"):
```
MetRonnie: Re-adding the code comment
```suggestion
        ) and (
            # This adds an 'orphaned' warning, only if the section is enabled.
            self.section is not None
            and self.section.name.startswith("namelist:")
        ):
```

metomi/rose/config_editor/main.py
MetRonnie: Difficult to tell which `sorted` the `key` applies to. Also would be good to have a comment explaining this unusual code:
```suggestion
            # Sort by alphabetical order first, then by length
            for ns in sorted(
                sorted(ctrl.data.namespace_meta_lookup),  # NOQA: C414
                key=len,
            ):
```

metomi/rose/config_editor/valuewidget/choice.py
```diff
@@ -249,10 +249,8 @@ def _get_is_implicit(self, name):
         values = self._get_value_values()
         if self.all_group in values:
             return True
-        for group in self.groups:
-            if group in values and set(group).issubset(name) and group != name:
-                return True
-        return False
+        return any(group in values and set(group).issubset(name)
+                   and (group != name) for group in self.groups)
```
MetRonnie:
```suggestion
        return any(
            group in values and set(group).issubset(name) and (group != name)
            for group in self.groups
        )
```

metomi/rose/apps/comparisons/exact.py
MetRonnie: Ah, so it doesn't introduce a bug because `location` gets reset to what `enumerate` yields on each iteration. However it looks like a bug, and can be gotten rid of by using the `start` argument: https://github.com/metomi/rose/pull/3048#discussion_r4027737525

samuel-denton: Interesting, so enumerate remembers what index it got to and sets location to that, rather than incrementing the local location variable?  So the +1 just means location stays 1 ahead of the loops index.  

That is pretty unintuitive to me, ~~I would be tempted to add a comment to it.~~  It looked wrong to me as well when you pointed it out.

EDIT: Just spotted your suggestion to change this to just have a start value of 1, much cleaner.

metomi/rose/apps/comparisons/exact.py
samuel-denton: That is much more readable to me than the location += 1

Scott-Owen-James: Huh, didn't know you could do that...

metomi/rose/config_editor/main.py
Scott-Owen-James: It's the first time sort is called that has to be ignored, needs to stay on first line.

metomi/rose/gtk/splash.py
```diff
@@ -292,12 +290,9 @@ def main(argv=sys.argv):
         splash_screen, stop_event, sys.stdin
     )
     update_thread.start()
-    try:
+    with contextlib.suppress(KeyboardInterrupt):
         Gtk.main()
-    except KeyboardInterrupt:
-        pass
-    finally:
-        update_thread.join()
```
MetRonnie, line 300: This is actually needed. The `finally` block _always_ runs, regardless of what exception is raised inside the `try` block. Whereas now it will only run if `KeyboardInterrupt` is raised (or no exception).

metomi/rose/suite_engine_procs/cylc.py
```diff
@@ -81,13 +82,13 @@ def get_suite_jobs_auths(
...
         if cycle_name_tuples is not None:
-            for cycle, name in cycle_name_tuples:
+            for cycle, _ in cycle_name_tuples:
                 new_platforms = get_platforms_from_task_jobs(suite_name, cycle)
                 task_platforms[cycle] = new_platforms
 
         # For each platform get a list of hosts.
         hosts = []
-        for cycle, tasks in task_platforms.items():
+        for _, tasks in task_platforms.items():
```
MetRonnie: Dictionaries have 3 methods for iterating: `.keys()`, `.values()` and `.items()` which yields a tuple of the key and value. If we are only using either the key or value, then we can use the appropriate method instead of `.items()`
```suggestion
        for tasks in task_platforms.values():
```

metomi/rose/metadata_gen.py
```diff
@@ -103,12 +104,12 @@ def type_gen(value):
...
-    types = []
     length = 0
+    types = []
     if not value:
-        return None, str(length)
-    for val in metomi.rose.variable.array_split(value):
-        length += 1
+        return None, '0'
+    for lenTemp, val in enumerate(metomi.rose.variable.array_split(value)):
+        length = lenTemp + 1
```
MetRonnie, line 112: Did the linter flag up the old way? I think the old way is clearer in this case

metomi/rose/macros/value.py
```diff
@@ -287,12 +283,11 @@ def check_range(
                 tiny_config.set([sect, key], val)
                 tiny_meta_config = metomi.rose.config.ConfigNode()
                 evaluator = metomi.rose.macros.rule.RuleEvaluator()
-                try:
+                with contextlib.suppress(metomi.rose.
+                                         macros.rule.RuleValueError):
```
MetRonnie:
```suggestion
                with contextlib.suppress(
                    metomi.rose.macros.rule.RuleValueError
                ):
```

metomi/rose/gtk/splash.py
Scott-Owen-James: Little confused here, update_thread.join() should still always run after the `with` block concludes and should produce the same result, can you enlighten me?

metomi/rose/metadata_gen.py
Scott-Owen-James: If I changed it there was a linter issue, I can try to make it look cleaner, there were some parts of the code for which we could resolve linting issues easily but not cleanly.

metomi/rose/gtk/splash.py
MetRonnie: If e.g. a `ValueError` is raised inside of `Gtk.main()`, the program will crash and `update_thread.join()` will not run.

Whereas before, the join _would_ run, because it's inside the `finally` clause, before the program crashes.

You can try this out yourself with some simple examples

```python
try:
    raise ValueError(0)
except KeyboardInterrupt:
    pass
finally:
    print("Finally")
```
vs
```python
with suppress(KeyboardInterrupt):
    raise ValueError(0)
print("Finally")
```

Does that make sense?

metomi/rose/metadata_gen.py
MetRonnie: Can revert this and stick a `# noqa: SIM113` comment on it

Scott-Owen-James: Done.

metomi/rose/namelist_dump.py
```diff
@@ -112,7 +112,7 @@ def namelist_dump(args=None, output_file=None, case_mode=None):
         config.set([section, "source"], " ".join(group_sections))
 
     # Add namelist: sections
-    for name, groups in groups_by_name.items():
+    for _, groups in groups_by_name.items():
```
MetRonnie:
```suggestion
    for groups in groups_by_name.values():
```

metomi/rose/task_env.py
```diff
@@ -112,7 +112,7 @@ def get_prepend_paths(
             prepend_paths_map[name].extend(more_prepend_paths)
         elif more_prepend_paths:
             prepend_paths_map[name] = more_prepend_paths
-    for key, prepend_paths in prepend_paths_map.items():
+    for _, prepend_paths in prepend_paths_map.items():
```
MetRonnie: (Can you check for other instances of this in this PR?)
```suggestion
    for prepend_paths in prepend_paths_map.values():
```

Scott-Owen-James: Checked through it, as far as I can tell there aren't any more.

metomi/rose/variable.py
```diff
@@ -278,9 +278,9 @@ def _is_quote_state_change(string, index, quote_lookup, quote_state):
     prev_letters_escaped = i % 2 == 0
     if index < len(string) - 1:
         next_letter_is_same = string[index + 1] == letter
-    if letter in quote_state and not quote_state[quote_lookup[letter]]:
-        if prev_letters_escaped and not next_letter_is_same:
-            return True
+    if ((letter in quote_state and not quote_state[quote_lookup[letter]])
+       and (prev_letters_escaped and not next_letter_is_same)):
+        return True
```
MetRonnie, line 283: This function looks suspect. Just above we have
```python
i = 0
while i > 0:
```
The `while` loop will never execute!

Unless I am mistaken, this function is not actually used anywhere in rose or cylc-rose, so we should remove it entirely.

metomi/rosie/db.py
```diff
@@ -300,7 +301,7 @@ def _get_compound_sql_expr(self, items):
         """Construct a complex logical expression containing "(", and, etc."""
         levels = [[[]]]
         level = 0
-        for i, item in enumerate(items):
+        for _, item in enumerate(items):
```
MetRonnie, line 304:
```suggestion
        for item in items:
```

metomi/rosie/vc.py
```diff
@@ -294,7 +294,8 @@ def create(
                 )
                 self.popen("svn", "export", "-q", "--force", from_id_url, dir_)
             else:
-                open(os.path.join(dir_, "rose-suite.conf"), "w").close()
+                with open(os.path.join(dir_, "rose-suite.conf"), "w") as file:
+                    file.close()
```
MetRonnie, line 298: Which lint rule was this falling foul of? I think it was fine before.

metomi/rosie/ws_client_auth.py
```diff
@@ -128,7 +129,7 @@ def get_socket(cls) -> socket.socket:
         try:
             gpg_socket.connect(socket_address)
         except socket.error as exc:
-            raise GPGAgentStoreConnectionError(f"socket error: {exc}")
+            raise GPGAgentStoreConnectionError(f"socket error: {exc}") from exc
```
MetRonnie, line 132: If we are including the original exception message in the new exception message, we don't need to include the original as context using `raise from`.
```suggestion
            raise GPGAgentStoreConnectionError(f"socket error: {exc}") from None
```

metomi/rosie/db.py
Scott-Owen-James: Missed some, that's what I get for searching with ctrl+f

metomi/rosie/vc.py
Scott-Owen-James: Was in violation of SIM115, Oliver decided this was a rose anti-pattern when we encountered others of this rule that were harder to modify so it was added to the ignore pile in tox.ini midway through. This was implemented before that decision, now redundant so will revert.

metomi/rose/variable.py
Scott-Owen-James: Function deleted.

PR #3058, opened 2026-06-24
2.7.x sync

oliver-sanders:


oliver-sanders:
`pip` is backtracking and replacing `rose` with an earlier version, likely cause would be out-of-sync cylc-admin metadata.

oliver-sanders:
Cause: https://github.com/metomi/rose/pull/2632#issuecomment-4789832029

oliver-sanders:
Test failing due to compat mode removal....


https://github.com/metomi/rose/pull/3075

MetRonnie:
Kicking tests

oliver-sanders:
The docs have now sprung a fresh failure (is that 4 in a row?!).

But the regular CI is now going through fine :sweat:

metomi/rose/__init__.py
```diff
@@ -144,4 +144,4 @@
 # Paths in the Rose distribution.
 FILEPATH_README = "README.md"
 
-__version__ = "2.7.1.dev"
+__version__ = "2.7.2.dev"
```
MetRonnie: This should probably be 2.8.0.dev but was never bumped on the master branch?

MetRonnie (approved):
Happy with sorting out the docs failures later.

metomi/rose/__init__.py
oliver-sanders:
```suggestion
__version__ = "2.8.0.dev"
```

oliver-sanders:
:facepalm: 

And the version bump broke things :cry:, should really have attempted that separately to unblock the sync PRs, but may as well get it fixed now:  https://github.com/cylc/cylc-rose/pull/436

MetRonnie (approved):

PR #3064, opened 2026-07-22
Tutorial: Update value of `MAP_TEMPLATE` in a practical step of `rose-app.conf`

mo-wilfwilson:
The lines a little above say:

>    Copy the remaining environment variables defined in the `forecast`
>      task within the `flow.cylc` file into the `rose-app.conf`
>      file of the `forecast` application, replacing any values already
>      specified if necessary. Remove the lines from the `flow.cylc` file
>      when you are done.

And `MAP_TEMPLATE` has a new value in `flow.cylc` for me, so shouldn't the answer look like I've indicated in the diff?

### DIFF
--- a/sphinx/tutorial/rose/suites.rst
+++ b/sphinx/tutorial/rose/suites.rst
@@ -470,7 +470,7 @@ Rose Applications In Rose Suite Configurations
          INTERVAL=60
          N_FORECASTS=5
          WEIGHTING=1
-         MAP_TEMPLATE=map-template.html
+         MAP_TEMPLATE=$CYLC_WORKFLOW_RUN_DIR/lib/template/map.html
          SPLINE_LEVEL=0
          WIND_FILE_TEMPLATE=$CYLC_WORKFLOW_WORK_DIR/{cycle}/consolidate_observations/wind_{xy}.csv
          WIND_CYCLES=0, -3, -6


mo-wilfwilson:
Okay, probably not, because then the workflow doesn't work:
```
Plotting disabled
  File "/home/users/wilf.wilson/cylc-run/rose-suite-tutorial/run4/app/forecast/bin/forecast", line 253, in <module>
    main(*args)
  File "/home/users/wilf.wilson/cylc-run/rose-suite-tutorial/run4/app/forecast/bin/forecast", line 241, in main
    util.generate_html_map(map_file, map_template, forecasts, domain,
  File "/home/users/wilf.wilson/cylc-run/rose-suite-tutorial/run4/app/forecast/bin/util.py", line 301, in generate_html_map
    with open(template_file, 'r') as template:  # noqa: SIM117
         ^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/users/wilf.wilson/cylc-run/rose-suite-tutorial/run4/lib/template/map.html'
[FAIL] forecast $INTERVAL $N_FORECASTS # return-code=1
2026-07-22T15:01:43+01:00 CRITICAL - failed/ERR
```
So maybe the code for `flow.cylc` that the tutorial creates is wrong? i.e. shouldn't contain the line `MAP_TEMPLATE=$CYLC_WORKFLOW_RUN_DIR/lib/template/map.html`?

samuel-denton (changes requested):
Good spot!  You are correct that there is a problem currently.  The MAP_TEMPLATE should indeed be "map-template.html" as the tutorial states, but within the provided flow.cylc, it has the MAP_TEMPLATE set to "$CYLC_WORKFLOW_RUN_DIR/lib/template/map.html".

However, correcting this is not quite as simple as altering the source of the tutorial files.  That file is located here:
https://github.com/metomi/rose/blob/master/metomi/rose/etc/tutorial/rose-suite-tutorial/flow.cylc
However, it is simply a reference to a different tutorial file (Most of the tutorials use the same set of source files), located here:
https://github.com/metomi/rose/blob/master/metomi/rose/etc/tutorial/cylc-forecasting-workflow/flow.cylc
And the tutorial that is part of requires the MAP_TEMPLATE remain as it is.

The reason is that this tutorial is meant to continue the previous Cylc tutorial, converting a Cylc app into a Rose app, so in theory a user could continue editing their existing Cylc tutorial workflow (although in practice most people will get fresh files using ```rose tutorial rose-suite-tutorial ~/cylc-src/rose-suite-tutorial```).  Unlike Cylc, rose automatically installs files within applicationXYZ/files to the root, so converting to a rose application, we also update how we reference that file.

So I think the solution here is to update the tutorial text telling users to copy over all variables.
The tutorial in question is found here:
https://github.com/mo-wilfwilson/rose/edit/patch-2/sphinx/tutorial/rose/suites.rst
Something like:
```rst
      Copy the remaining environment variables defined in the ``forecast``
      task within the ``flow.cylc`` file into the :rose:file:`rose-app.conf`
      file of the ``forecast`` application, replacing any values already
      specified if necessary. The exception is MAP_TEMPLATE, which should
      remain as "map-template.html".  Remove the lines from the ``flow.cylc``
      file when you are done.
```

mo-wilfwilson:
Oh wait, actually there's not a problem except for my ability to skip over paragraphs without reading them properly. Because the preceding paragraph already says:
```
      We have moved the map template file (``map-template.html``) into the
      ``forecast`` application so we can delete the ``MAP_TEMPLATE``
      environment variable from the ``[runtime]forecast`` section of the
      ``flow.cylc`` file.
```
So the `MAP_TEMPLATE` variable is already taken care of, right?

I've updated this PR to include an extra line of prose for explicitness. You're welcome to either accept the change, or just close this PR seeing as the problem was my fault.

samuel-denton (approved):
Nice clarification.  
At some point we intend to re-write the tutorials to have less little copy paste steps and little edits that don't really teach you how to use Cylc/Rose so much as testing how carefully you follow a long list of instructions!

PR #3066, opened 2026-07-22
Tutorial: Change `title` to `description` in `metadata.rst`

mo-wilfwilson:
In the example above, a `description` is set rather than a `title`. Since, according to the page, `title` and `description` are both valid keys, maybe it's a bit confused to use `description` in the code but `title` in the text.

I suggest being consistent, one way or the other.

### DIFF
--- a/sphinx/tutorial/rose/metadata.rst
+++ b/sphinx/tutorial/rose/metadata.rst
@@ -69,7 +69,7 @@ The Metadata Format
    values=Mercury, Venus, Earth, Mars, Jupiter,
          =Saturn, Uranus, Neptune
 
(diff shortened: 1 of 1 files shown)


samuel-denton (approved):
LGTM - Just needs the contributor list etc filling out and tests to pass.

mo-wilfwilson:
Thanks for the review @samuel-denton. I'm not sure that the test failures were my fault; I'll see what they say when they run again.

I've added the necessary `CONTRIBUTING` changes in #3068, so hopefully this PR can be merged once that one has been.

samuel-denton:
All checks pass, LGTM

PR #3074, opened 2026-08-06
actions: fix rose installation

oliver-sanders:
Install Python packages in a single operation to avoid the risk of up/down-grading components or entering erroneous pip-backtracking loops.

Should fix this problem: https://github.com/metomi/rose/pull/3058#issuecomment-4789807210

And get the tests passing on master again.


oliver-sanders:
That's done it:

Screenshot from 2026-08-06 13-18-07

Merged upstream and switched this back to `v1`.

MetRonnie (approved):

PR #3077, opened 2026-08-20
docs: fix warning

oliver-sanders:
This fixes a newly added Sphinx warning which has been causing the docs to fail.

This code gets auto-documented here: https://metomi.github.io/rose/doc/html/api/configuration/api.html#python

The presentation of those two tables will change as expected, but everything else seems to work fine.


ChrisPaulBennett (approved):
Looks good

PR #3085, opened 2026-09-16
Prefix local git paths with `file://` to support filtered clones

### LINKED ISSUE #3084: loc handlers: git: fail when doing filtered fetch from remote starting with '/'
If you file install from a repo on a filesystem with a path beginning with '/', with filtering enabled, the current Rose file-install code will fail:

```bash
#!/bin/bash

set -eu

SOURCE_REPO=$(mktemp -d)
cd $SOURCE_REPO
git init -q
touch foo
touch bar
git add .
git commit -q -m "initial import"
git config uploadpack.allowAnySHA1InWant true
git config uploadpack.allowFilter true
cd $(mktemp -d)
cat >rose-app.conf <<__APP__
[command]
default=true

[file:foo]
source=git:$SOURCE_REPO::foo::main
__APP__
rose app-run
```

fails with:

warning: promisor remote name cannot begin with '/': .................
fatal: missing blob object '...............'
error: remote did not send all necessary objects

This is a Git issue / behaviour with filtered clones: https://code.googlesource.com/git/+/98c496fcd09b5894966f3e499217eb2bdf8b964d/promisor-remote.c#56 and https://gitperf.com/chapter-11.html.

There is a workaround and a correction - the workaround is to use url.insteadOf settings and use some kind of alias there. The correction is, e.g., inside the loc handler git code, add `git remote add origin MY_REMOTE_THAT_STARTS_WITH_SLASH` then run the git fetch referencing `origin`.

jfrost-mo:
This makes git treat it as a remote URL, allowing the filtering to work. Additionally an e2e test is added.

An alternative approach would be to drop the filtering for local source repositories and instead use the `--shared` option to skip copying the history, which would still create the full working tree, but would skip the objects. For many (most?) repositories there will be more historic objects than files in the working tree, so that should provide a speedup. Repositories created by --shared are at risk of being broken if the source repository is deleted, but for rose file we immediately discard the repository anyway, so the would not be an issue.

Fixes #3084


samuel-denton (approved):
Neat fix, LGTM

jfrost-mo:
Hmm, tests are failing in CI, but were fine locally.

jfrost-mo:
Ah, the issue seems to be that I have `init.defaultBranch = main` specified in my .gitconfig, while the runner does not, so is defaulting to `master` instead. Give this is going to change at Git 3.0, we should encode it in the `git init` call.

samuel-denton:
> Ah, the issue seems to be that I have `init.defaultBranch = main` specified in my .gitconfig, while the runner does not, so is defaulting to `master` instead. Give this is going to change at Git 3.0, we should encode it in the `git init` call.

Yep I had a few errors with the test but fixed them by altering some branch names locally.  I had to play a little to get it to install the correct branch and make use of your fix too.  Not 100% on how it should be setup for CI

samuel-denton:
The URL prefix seems to be impacting some other tests too.  I think:
t/rose-app-run/28-git.t
Needs updating to expect your prefix.

jfrost-mo:
Got the tests passing now; as you said I needed to update the captured log messages in the existing git tests. This raises the question of whether I should add my new test to that existing file rather than creating a new one.

samuel-denton:
> Got the tests passing now; as you said I needed to update the captured log messages in the existing git tests. This raises the question of whether I should add my new test to that existing file rather than creating a new one.

I think either way is fine, it would fit in 28-git.t but it's different enough that splitting it out is justified.

jfrost-mo:
In that case I'll leave it as is.

jfrost-mo:
Also, should I retarget this PR against the `2.7.x` branch?

samuel-denton:
> Also, should I retarget this PR against the `2.7.x` branch?

TBH this is still something I get stuck on, best get @MetRonnie's opinion.
We have this page: 
https://cylc.github.io/cylc-admin/status/status.html
which tells you which branches should be compatible and so which branch to target depending on which release it's going to be part of.
There is this page as well that shows priorities:
https://github.com/orgs/cylc/projects

As this is a bugfix, I think you are correct that it should target 2.7.x but I seem to get it wrong every time :P

EDIT: And this page explaining it:
https://cylc.github.io/cylc-admin/howto/releases-versions-branches

MetRonnie:
Yes, as this is a bugfix it should be rebased onto 2.7.x. The only exception would be if the bugfix refactored a lot of important code, in which case our low risk appetite for ".x" branches would dictate basing it on master (bit of a judgment call really).

MetRonnie:
The approach here looks good to me, @benfitzpatrick can you confirm you are happy with it?

benfitzpatrick:
Yep clever way of doing it :)

MetRonnie (approved):

MetRonnie:
Just kicking the tests as they don't appear to have run since the change of base branch

MetRonnie:
Docs linkcheck timed out, not important

